PR #2, opened 2020-01-23
Pooled risk assessment changes

roxdanila:


reemaSomish:
Hi @roxdanila , I think review is pending in this one, once done we can close out issues #89 , #90

PR #3, opened 2020-01-27
Distributor contract

roxdanila:


contracts/Distributor.sol
```diff
@@ -0,0 +1,148 @@
...
+
+  struct TokenData {
+    uint expirationTimestamp;
+    address lastOwner;
+    bytes4 coverCurrency;
+    uint[] coverDetails;
+    uint coverId;
+  }
+
+  uint public constant CLAIM_VALIDITY_MAX_DAYS_OVER_COVER_PERIOD = 30 days;
```
roxdanila: it should be read from here: https://github.com/danoctavian/smart-contracts/blob/8d2eaf6d322ace4bdf31879114a4c55f09041fe9/contracts/TokenData.sol#L26

contracts/Distributor.sol
```diff
@@ -0,0 +1,148 @@
...
+  uint public constant CLAIM_DEPOSIT_PERCENTAGE = 5;
```
roxdanila: this is tokensRetained / 2: https://github.com/NexusMutual/smart-contracts/blob/master/contracts/QuotationData.sol#L67

contracts/Distributor.sol
```diff
@@ -0,0 +1,148 @@
...
+    uint requiredValue = priceLoadPercentage.mul(coverDetails[1]).div(100).add(coverDetails[1]);
+    require(msg.value == requiredValue, "Incorrect value sent");
+
+    Pool1.Pool1 p1 = Pool1.Pool1(nxMaster.getLatestAddress("P1"));
+    p1.makeCoverBegin.value(coverDetails[1])(coveredContractAddress, coverCurrency, coverDetails, coverPeriod, _v, _r, _s);
+
+    mintToken(coverCurrency, coverDetails, coverPeriod);
+  }
+
+  function buyCoverUsingCA(
```
roxdanila: I'd suggest combining this with the function above

contracts/Distributor.sol
```diff
@@ -0,0 +1,148 @@
...
+    require(_isApprovedOrOwner(msg.sender, tokenId), "Not approved or owner");
+    require(allTokenData[tokenId].expirationTimestamp > block.timestamp, "Token is expired");
+    require(allTokenData[tokenId].coverCurrency == "ETH", "currency not ETH");
+    uint coverAmount = allTokenData[tokenId].coverDetails[1];
+    require(msg.value == CLAIM_DEPOSIT_PERCENTAGE.mul(coverAmount).div(100), "Deposit value is incorrect");
+
+    _submitClaim(tokenId);
+  }
+
+  function submitClaimUsingCA(
```
roxdanila: same as above, we could combine them into one submitClaim function for easier integration with frontend

contracts/Distributor.sol
roxdanila: given it's hard coded in the constructor in QuodationData, for now we don't have a choice, so we can keep it as is

contracts/Distributor.sol
```diff
@@ -0,0 +1,293 @@
...
+    Pool1.Pool1 p1 = Pool1.Pool1(nxMaster.getLatestAddress("P1"));
+    p1.makeCoverBegin.value(coverDetails[1])(coveredContractAddress, coverCurrency, coverDetails, coverPeriod, _v, _r, _s);
+
+    // add fee to the withdrawable pool
+    withdrawableETH = withdrawableETH.add(requiredValue.sub(coverDetails[1]));
+
+    mintToken(coverCurrency, coverDetails, coverPeriod);
+  }
+
+  function buyCoverUsingCA(
```
roxdanila: Can we merge these 2 functions? In terms of interface, I don't see why we'd need 2 functions. Any concerns?

contracts/Distributor.sol
```diff
@@ -0,0 +1,293 @@
...
+
+    address payable pool1Address = nxMaster.getLatestAddress("P1");
+    Pool1.Pool1 p1 = Pool1.Pool1(pool1Address);
+    erc20.approve(pool1Address, coverDetails[1]);
+    p1.makeCoverUsingCA(coveredContractAddress, coverCurrency, coverDetails, coverPeriod, _v, _r, _s);
+
+    // add fee to the withdrawable pool
+    withdrawableDAI = withdrawableDAI.add(requiredValue.sub(coverDetails[1]));
+
+    mintToken(coverCurrency, coverDetails, coverPeriod);
```
roxdanila: let's keep this generic to any ERC20, shouldn't be specific to DAI

contracts/Distributor.sol
```diff
@@ -0,0 +1,230 @@
...
+
+  constructor(address _masterAddress, uint _priceLoadPercentage) public {
+    nxMaster = INXMMaster.INXMMaster(_masterAddress);
+    priceLoadPercentage = _priceLoadPercentage;
+  }
+
+  function buyCover(
+        address coveredContractAddress,
+        bytes4 coverCurrency,
+        uint[] memory coverDetails,
```
roxdanila, line 62: make the elements of coverDetails explicit in the interface

contracts/Distributor.sol
```diff
@@ -0,0 +1,230 @@
...
+        bytes32 _r,
+        bytes32 _s
+  )
+     public
+     payable
+  {
+    uint requiredValue = priceLoadPercentage.mul(coverDetails[1]).div(100).add(coverDetails[1]);
+
+    if (coverCurrency == "ETH") {
+      require(msg.value == requiredValue, "Incorrect value sent");
```
roxdanila, line 74: shouldn't this be before the if-else block?

contracts/Distributor.sol
```diff
@@ -0,0 +1,230 @@
...
+
+      // add fee to the withdrawable pool
+      withdrawableETH = withdrawableETH.add(requiredValue.sub(coverDetails[1]));
+    } else {
+      PoolData.PoolData pd = PoolData.PoolData(nxMaster.getLatestAddress("PD"));
+      IERC20.IERC20 erc20 = IERC20.IERC20(pd.getCurrencyAssetAddress(coverCurrency));
+      require(erc20.transferFrom(msg.sender, address(this), requiredValue), "Transfer failed");
+
+      address payable pool1Address = nxMaster.getLatestAddress("P1");
+      Pool1.Pool1 p1 = Pool1.Pool1(pool1Address);
```
roxdanila, line 87: this also looks like it can be taken up, before the if-else block

contracts/Distributor.sol
```diff
@@ -0,0 +1,230 @@
...
+      erc20.approve(pool1Address, coverDetails[1]);
+      p1.makeCoverUsingCA(coveredContractAddress, coverCurrency, coverDetails, coverPeriod, _v, _r, _s);
+
+      // add fee to the withdrawable pool
+      withdrawableTokens[coverCurrency] = withdrawableTokens[coverCurrency].add(requiredValue.sub(coverDetails[1]));
```
roxdanila, line 92: for readability I'd put this in a variable `requiredValue.sub(coverDetails[1]`

PR #7, opened 2020-03-10
Pooled staking dev

shark0der:


contracts/PooledStaking.sol
```diff
@@ -0,0 +1,215 @@
...
+    revert("NOT IMPLEMENTED");
+  }
+
+  function deallocate(address contractAddress, uint amount) onlyInternal public {
+    contractAddress;
+    amount;
+    revert("NOT IMPLEMENTED");
+  }
+
+  function updateParameter(ParamType param, uint value) onlyGoverned external {
```
danoctavian: should this be called `updateUintParameter` or `updateUint256Parameter` in case another type is added

contracts/PooledStaking.sol
```diff
@@ -0,0 +1,215 @@
...
+
+  function stake(uint amount) onlyMembers external {
+
+    require(amount > MIN_DEPOSIT_AMOUNT, "Amount is less than minimum allowed");
+
+    Staker storage staker = stakers[msg.sender];
+    uint oldStake = staker.staked;
+    staker.staked = staker.staked.add(amount);
+
+    for (uint i = 0; i < staker.contracts.length; i++) {
```
danoctavian: i think we should document with a comment why this loop is bounded, based on the parameters MIN_STAKE_PERCENTAGE and MAX_LEVERAGE as discussed.

test/PooledStaking/updateParameter.js
```diff
@@ -0,0 +1,69 @@
...
+] = accounts;
+
+describe('updateParameter', function () {
+
+  beforeEach(async function () {
+    const master = await MasterMock.new();
+    const staking = await PooledStaking.new();
+
+    await master.enrollMember(member, Role.Member);
+    await master.enrollMember(abMember, Role.AdvisoryBord);
```
danoctavian: `Role.AdvisoryBoard`  ?

test/PooledStaking/updateParameter.js
```diff
@@ -0,0 +1,69 @@
...
+
+const { Role, ParamType } = require('../utils/constants');
+
+const MasterMock = contract.fromArtifact('MasterMock');
+const PooledStaking = contract.fromArtifact('PooledStaking');
+
+const [
+  nonMember,
+  member,
+  abMember,
```
danoctavian: how about `advisoryBoardMember` ? i noticed a lot of tests in solidity world use these abbreviations, but with an IDE typing it shouldn't be an issue

test/PooledStaking/updateParameter.js
shark0der: Accepted!

contracts/PooledStaking.sol
danoctavian: ok this is part of a view now

contracts/PooledStaking.sol
```diff
@@ -0,0 +1,512 @@
...
+    );
+
+    require(
+      _contracts.length == _insertAfter.length,
+      "Contracts and _insertAfter arrays should have the same length"
+    );
+
+    Staker storage staker = stakers[msg.sender];
+
+    for (uint i = 0; i < _contracts.length; i++) {
```
danoctavian, line 257: need to check how much gas these loops use as you increase _contracts and compare with gas block limit

ideally we have a test that monitors gas usage for an acceptable upper limit

contracts/PooledStaking.sol
```diff
@@ -0,0 +1,512 @@
...
+
+      bool canBurn = burn.burnedAt != 0; // end not reached
+      bool canDeallocate = deallocation.deallocateAt != 0 && deallocation.deallocateAt <= now;
+
+      if (!canBurn && !canDeallocate) {
+        // everything is processed
+        break;
+      }
+
+      // TODO: check if there's enough gas for the next iteration
```
danoctavian, line 362: thinking about allowing  an optional iteration limit here, because if the gas check is at any point off it could potentially block this function

contracts/PooledStaking.sol
```diff
@@ -0,0 +1,512 @@
...
+      uint oldAllocation = staker.allocations[contractAddress];
+
+      // formula: staker_burn = staker_allocation / total_contract_stake * contract_burn
+      // reordered for precision loss prevention
+      uint stakerBurn = oldAllocation.mul(burnAmount).div(_contract.staked);
+      uint newStake = staker.staked.sub(stakerBurn);
+      staker.allocations[contractAddress] = oldAllocation.sub(stakerBurn);
+
+      // reduce other contracts' stakes if needed
+      for (uint j = 0; j < staker.contracts.length; j++) {
```
danoctavian, line 410: since this is O(m*n) we need tests around gas usage here to see exactly what can we support and potentially make a test fail if a certain amount of gas is used that is not acceptable

contracts/PooledStaking.sol
```diff
@@ -0,0 +1,512 @@
...
+    uint staked; // amount of nxm staked for this contract
+    // TODO: find a way to remove zero-amount stakers. Linked list?
+    address[] stakers; // used for iteration
+  }
+
+  struct Burn {
+    uint amount;
+    uint burnedAt;
+    address contractAddress;
+    uint next; // id of the next deallocation request in the linked list
```
danoctavian, line 57: to clarify the reasons for using he linked list here are:

* the list is sorted by deallocation time
* O(1) insertion into said list using index hint

contracts/PooledStaking.sol
```diff
@@ -0,0 +1,512 @@
...
+
+    Burn storage burn = burns[firstBurn];
+    Contract storage _contract = contracts[burn.contractAddress];
+    address contractAddress = burn.contractAddress;
+
+    uint burnAmount = burn.amount <= _contract.staked ? burn.amount : _contract.staked;
+    uint burned = 0;
+    uint newContractStake = 0;
+
+    for (uint i = 0; i < _contract.stakers.length; i++) {
```
danoctavian, line 397: is there any known upper bound to the _contract.stakers.length as of now?

we established for staker.contracts there is one that results from 

MIN_STAKE_PERCENTAGE and MAX_LEVERAGE

contracts/PooledStaking.sol
danoctavian, line 362: the gas check here would now on _contract.stakers.length * staker.contracts.length so we'd need some bounds there to know if there is enough gas

contracts/PooledStaking.sol
```diff
@@ -0,0 +1,522 @@
...
+        break;
+      }
+
+      if (
+        canBurn &&
+        (!canDeallocate || burn.burnedAt < deallocation.deallocateAt) &&
+        (!canReward || burn.burnedAt < reward.rewardedAt)
+      ) {
+        // O(n*m)
+        _processFirstBurn();
```
danoctavian, line 368: i'd guess we need to go 1 level deeper with gas estimations since at this level only n can be evaluated contracts.length

contracts/PooledStaking.sol
roxdanila: worst case scenario, that's all members - that's around 700 atm, but I expect it to grow. now, this WC scenario is quite unlikely, as it's not economically viable at this point (i.e. small rewards distributed to too many people -> negligible rewards)

contracts/PooledStaking.sol
```diff
@@ -0,0 +1,593 @@
...
+      uint oldAllocation = staker.allocations[contractAddress];
+
+      // formula: staker_burn = staker_allocation / total_contract_stake * contract_burn
+      // reordered for precision loss prevention
+      uint stakerBurn = oldAllocation.mul(burnAmount).div(_contract.staked);
+      uint newStake = staker.staked.sub(stakerBurn);
+      staker.allocations[contractAddress] = oldAllocation.sub(stakerBurn);
+
+      // reduce other contracts' stakes if needed
+      for (uint j = 0; j < staker.contracts.length; j++) {
```
danoctavian, line 447: i think we should consider enforcing a low upper bound here from a feature perspective here similar to the one resulting from the percentage limitation.

contracts/PooledStaking.sol
danoctavian: was thinking in case we need to cap this for  the pool to perform properly we could agree on a feature level to a cap where only a "select club" can enter the pool based on some rules

danoctavian: or first come first served and a limit on number of stakers per a particular contract

contracts/PooledStaking.sol
```diff
@@ -425,47 +428,63 @@ contract PooledStaking is MasterAware, TokenAware {
...
+      uint contractCount = staker.contracts.length;
+
+      // if needed, reduce stakes for other contracts
+      for (uint j = processedToContractIndex; j < contractCount; j++) {
 
         address _staker_contract = staker.contracts[j];
         uint prevAllocation = staker.allocations[_staker_contract];
 
+        // can't have allocated more than staked
+        // branch won't be executed for the burned contact since we updated the allocation earlier
```
roxdanila, line 461: typo: burned. contract*

contracts/PooledStaking.sol
```diff
@@ -250,23 +271,28 @@ contract PooledStaking is MasterAware {
...
-        .add(newAllocation);
+      contracts[contractAddress].staked = contracts[contractAddress].staked.add(increase);
+
+      emit Allocated(contractAddress, msg.sender, increase);
     }
 
     require(
       totalAllocation <= staker.staked.mul(MAX_LEVERAGE),
       "Total allocation exceeds maximum allowed"
     );
```
roxdanila: move this further up

contracts/PooledStaking.sol
```diff
@@ -284,27 +310,26 @@ contract PooledStaking is MasterAware {
...
       address contractAddress = _contracts[i];
       uint allocated = staker.allocations[contractAddress];
       uint pendingDeallocation = staker.pendingDeallocations[contractAddress];
-      uint requested = _amounts[i];
+      uint requestedAmount = _amounts[i];
       uint max = pendingDeallocation > allocated ? 0 : allocated.sub(pendingDeallocation);
 
       require(max > 0, "Nothing to deallocate on this contract");
 
       // To prevent spam, Small stakes and deallocations are not allowed
```
roxdanila, line 325: no need to capitalize "Small"

PR #10, opened 2020-05-04
Feature/pooled staking integration

danoctavian:


contracts/ClaimsReward.sol
```diff
@@ -207,8 +210,10 @@ contract ClaimsReward is Iupgradable {
      */
     function claimAllPendingReward(uint records) public isMemberAndcheckPause {
         _claimRewardToBeDistributed(records);
-        _claimStakeCommission(records);
-        tf.unlockStakerUnlockableTokens(msg.sender); 
+
+        pooledStaking.withdrawReward(pooledStaking.stakerReward(msg.sender));
+        pooledStaking.unstake(pooledStaking.getMaxUnstakable(msg.sender));
```
roxdanila, line 215: remove this, so we can keep this function only for claiming rewards (as the name says)

contracts/ClaimsReward.sol
```diff
@@ -207,8 +210,10 @@ contract ClaimsReward is Iupgradable {
      */
     function claimAllPendingReward(uint records) public isMemberAndcheckPause {
         _claimRewardToBeDistributed(records);
-        _claimStakeCommission(records);
-        tf.unlockStakerUnlockableTokens(msg.sender); 
+
+        pooledStaking.withdrawReward(pooledStaking.stakerReward(msg.sender));
```
roxdanila, line 214: just as a note, `stakerReward` doesn't exist in the PooledStaking contract

contracts/ClaimsReward.sol
danoctavian: ok so basically we change the semantics of this. the button in the UI won't give you all things

contracts/interfaces/IPooledStaking.sol
```diff
@@ -0,0 +1,15 @@
+
+interface IPooledStaking {
+    function pushReward(address contractAddress, uint amount) external;
+    function pushBurn(address contractAddress, uint amount) external;
+
+    function contractStakedAmount(address contractAddress) external view returns (uint);
+    function stakerReward(address staker) external view returns (uint);
```
danoctavian, line 7: these 2 getters don't exist yet @shark0der  need smth like this or any other name you see a good fit

contracts/TokenFunctions.sol
```diff
@@ -182,17 +111,7 @@ contract TokenFunctions is Iupgradable {
      * @param _stakerAddress address of the Staker.
      */ 
     function getStakerAllLockedTokens(address _stakerAddress) external view returns (uint amount) {
```
roxdanila, line 113: I couldn't find this function being used anywhere but the UI. If that's the case it'd be worth deprecating

contracts/TokenFunctions.sol
```diff
@@ -204,19 +123,9 @@ contract TokenFunctions is Iupgradable {
     )
         external
         view
-        returns (uint amount)
+        returns (uint)
```
roxdanila, line 126: looks like it's only been used in line 227 of ClaimRewards.sol, but you replaced it. if that's the case, it can be deprecated

contracts/ClaimsReward.sol
roxdanila: indeed. the flows are pretty different from a user / UX point of view:
-> withdraw all rewards will be served by this function
-> withdraw tokens you staked for claim assessment, so you can use them in voting again or for something else
-> withdraw stake (unstake) tokens from pooled staking

contracts/ClaimsReward.sol
```diff
@@ -222,12 +227,10 @@ contract ClaimsReward is Iupgradable {
...
         uint caReward = getRewardToBeDistributedByUser(_add);
-        uint commissionEarned = td.getStakerTotalEarnedStakeCommission(_add);
-        uint commissionReedmed = td.getStakerTotalReedmedStakeCommission(_add);
-        uint unlockableStakedTokens = tf.getStakerAllUnlockableStakedTokens(_add);
+        uint pooledStakingReward = pooledStaking.stakerReward(_add);
+        uint stakedTokens = pooledStaking.stakerStaked(_add);
         uint governanceReward = gv.getPendingReward(_add);
-        total = caReward.add(unlockableStakedTokens).add(commissionEarned.
-        sub(commissionReedmed)).add(governanceReward);
+        total = caReward.add(stakedTokens).add(pooledStakingReward).add(governanceReward);
```
roxdanila, line 233: I think it's worth adding on this branch the updates from. I think this is the only place where it impact you, there is an extra line missing (`uint unlockableStakedTokens = tf.getStakerAllUnlockableStakedTokens(_add);`)
https://github.com/NexusMutual/smart-contract-upgrades/tree/fix-issue-181-unlockBug

roxdanila: You can assume they will be on master when we deploy/merge this

PR #23, opened 2020-08-21
Disable oraclize

shark0der:
This PR removes the oraclize contract dependency and generates a time-based id that is returned instead of the oraclize one.
Needs to be merged only after merging the goverance dead code removal PR.


shark0der:
All oraclize actions will be available to be called manually, just like before.
In particular, the following oraclize queries are impacted as described:

1. Close emergency pause - given the fact that this will be a grand event, the emergency pause closing will have lots of eyes on it and the tx will be sent manually.
2. Close claim - a service for closing the claims would be needed, it is simple enough to build one.
3. Close cover - we already have a service to call the ones that oraclize can't close or misses, we're fine.
4. MCR oraclize, MCR fail oraclize - we don't need this
5. Save IA details - we haven't used this in a while

PR #26, opened 2020-09-02
[DO NOT MERGE] Reward grouping preview

shark0der:
Do not accept/merge this PR!
This PR exists with the sole purpose of having a clean diff until `feature/migrate-truffle-to-oz` is merged.


contracts/modules/staking/PooledStaking.sol
```diff
@@ -489,28 +500,104 @@ contract PooledStaking is MasterAware, IPooledStaking {
...
-    emit RewardRequested(contractAddress, amount);
+  /**
+   * @dev Add reward for contract. Automatically triggers distribution if enough time has passed.
+   */
+  function addPendingReward(address contractAddress, uint amount) external onlyInternal whenNotPausedAndInitialized {
+
+    // will push rewards if needed
+    address[] memory contractAddresses = new address[](1);
+    contractAddresses[0] = contractAddress;
+    _pushRewards(contractAddresses, false);
```
danoctavian, line 583: a lucky cover buyer can potentially pay a lot of for gas here if rewards are pushed in line right?

contracts/modules/staking/PooledStaking.sol
```diff
@@ -489,28 +500,104 @@ contract PooledStaking is MasterAware, IPooledStaking {
...
+
+    if (pushedRewards != 0) {
 
-    rewards[++lastRewardId] = Reward(amount, now, contractAddress);
+      lastRewardId = lastRewardIdCounter;
 
-    if (firstReward == 0) {
-      firstReward = lastRewardId;
+      if (firstReward == 0) {
+        firstReward = lastRewardIdCounter;
```
danoctavian, line 562: there could be many rewards pushed as part of this loop, shouldn't this be the first value that was used as a reward id in that loop, instead of the last?

contracts/modules/staking/PooledStaking.sol
```diff
@@ -489,28 +500,104 @@ contract PooledStaking is MasterAware, IPooledStaking {
     address contractAddress, uint amount
   ) public onlyInternal whenNotPausedAndInitialized noPendingBurns {
 
+    address[] memory contractAddresses = new address[](1);
+    contractAddresses[0] = contractAddress;
+    _pushRewards(contractAddresses, true);
```
danoctavian, line 505: what's the reason for pushing here?

contracts/modules/staking/PooledStaking.sol
```diff
@@ -126,6 +128,15 @@ contract PooledStaking is MasterAware, IPooledStaking {
...
+  // rewards to be distributed at the end of the current round
+  // contract address => ContractRewards
+  mapping(address => ContractReward) public accumulatedRewards;
+
+  uint public REWARD_ROUND_DURATION;
+  uint public REWARD_ROUNDS_START;
+
   /* Modifiers */
 
   modifier noPendingActions {
```
danoctavian, line 142: in regards to noPending modifiers:

noPendingUnstakeRequests and noPendingRewards seem to be unused now, can be removed

also noPendingActions and noPendingBurns are used only once, could inline the require there, either way it's fine

contracts/modules/staking/PooledStaking.sol
danoctavian, line 583: not sure if this is necessary here (gas reduction purposes only), isn't just the normal push sufficient?

PR #32, opened 2020-10-12
Feature/proof of loss

horodnicdragos:


contracts/mocks/ClaimProofs/CPQuotationDataMock.sol
```diff
@@ -0,0 +1,74 @@
...
+
+  Cover[] internal allCovers;
+
+  constructor() public {
+    allCovers.push(Cover(address(0), "0x00", 0, 0, 0, address(0), 0));
+
+		// Mock covers
+		// CID 1
+    allCovers.push(Cover(address(0), "0x00", 0, 30, now.add(30 days), address(0), 0));
+		coverStatus[1] = uint8(CoverStatus.Active);
```
danoctavian, line 47: strange identation here

contracts/modules/claims/ClaimProofs.sol
```diff
@@ -0,0 +1,26 @@
...
+    ms = INXMMaster(_masterAddress);
+  }
+
+  event ProofAdded(uint indexed coverId, address indexed owner, string ipfsHash);
+
+  function addProof(uint _coverId, string calldata _ipfsHash) external {
+    QuotationData qd = QuotationData(ms.getLatestAddress("QD"));
+    uint8 coverStatus;
+    (, coverStatus,,,) = qd.getCoverDetailsByCoverID2(_coverId);
+    require(coverStatus != uint8(QuotationData.CoverStatus.ClaimSubmitted), "Claim already submitted");
```
danoctavian, line 22: to confirm: there is a race that a third party that can submit a proof on your behalf, and override yours before you submit claim?

contracts/modules/claims/ClaimProofs.sol
```diff
@@ -0,0 +1,26 @@
...
+    emit ProofAdded(_coverId, msg.sender, _ipfsHash);
```
danoctavian, line 23: for this event i assume the UI will pick up the event with the most recent timestamp as the current proof ipfs hash

shark0der (changes requested):

test/unit/ClaimProofs/addProof.js
```diff
@@ -0,0 +1,51 @@
...
+  it('should emit an event with coverId, ipfsHash and sender when called', async function () {
+
+    const { claimProofs } = this;
+
+    const coverId = new BN(1);
+    const ipfsHash = 'mockedIpfsHash';
+    const res = await claimProofs.addProof(coverId, ipfsHash);
+    const proofAddedLog = res.logs.find(x => x.event === 'ProofAdded');
+    assert(proofAddedLog !== undefined, 'Expected a ProofAdded event');
+    assert(proofAddedLog.args[0].eq(coverId), `Expected ProofAdded event first argument to be ${coverId}`);
```
danoctavian, line 22: use expectEvent for this kind of assertions like here:

https://github.com/NexusMutual/shield-mining/blob/master/test/unit/depositRewards.js#L35

test/integration/utils/buyCover.js
```diff
@@ -0,0 +1,27 @@
+const { getQuoteValues } = require('./getQuote');
+
+function coverToCoverDetailsArray (cover) {
+  return [cover.amount, cover.price, cover.priceNXM, cover.expireTime, cover.generationTime];
+}
+
+async function buyCover ({ cover, coverHolder, qt, p1 }) {
```
danoctavian, line 7: why are parameters passed as a object instead of normal parameters?

contracts/modules/claims/ClaimProofs.sol
horodnicdragos: Yes, by using block numbers to be precise.

test/integration/utils/buyCover.js
horodnicdragos: I usually use this pattern to leave room for new params. If any of those would be optional you wouldn't have to move all the optional params at the end and it also makes it clearer what each param is. In this case however if you think it's unlikely for it to change we can add them one by one.

shark0der (approved):

danoctavian (approved):
LGTM

PR #35, opened 2020-10-14
Feature: claim payout address

shark0der:


danoctavian:
Looks good can we get at least one test where payout address is changed? integration, fork, anything is fine for me given we need to ship it quickly.

Also worth copying the contracts over into truffle and running the tests at least once

roxdanila (approved):

PR #52, opened 2021-02-02
Fix/testnet governance

horodnicdragos:


contracts/mocks/DisposableProposalCategory.sol
```diff
@@ -5,6 +5,8 @@ import "../modules/governance/MemberRoles.sol";
 
 contract DisposableProposalCategory is ProposalCategory {
 
+  uint[] allowedToCreateProposal;
+
```
shark0der, line 9: Do not create storage variables in Disposable contracts (Governance one is an exception due to contract size limit, we create the same variables in disposable as we have in the original).

PR #58, opened 2021-03-03
Feature/on chain mcr

danoctavian:


contracts/modules/capital/MCR.sol
```diff
@@ -199,53 +140,43 @@ contract MCR is Iupgradable {
...
-
-    if (len > 1) {
-      uint vtp = pool.getPoolValueInEth();
-      (lowerThreshold, upperThreshold) = getThresholdValues(vtp, vF, getAllSumAssurance(), pd.minCap());
-    }
+  function updateMCR(uint poolValueInEth) external onlyInternal {
 
-    if (mcrP > dynamicMincapThresholdx100) {
-      variableMincap = (variableMincap.mul(dynamicMincapIncrementx100.add(10000)).add(minCapFactor.mul(pd.minCap().mul(dynamicMincapIncrementx100)))).div(10000);
+    if (lastUpdateTime + minUpdateTime > now) {
```
danoctavian, line 145: Initially the buyNXM function used around 95k gas.

if it returns early the function uses 103091
if it does the update it uses 159097

contracts/modules/capital/MCR.sol
```diff
@@ -199,53 +140,43 @@ contract MCR is Iupgradable {
...
-  internal
-  {
-    uint lowerThreshold = 0;
-    uint upperThreshold = 0;
-
-    if (len > 1) {
-      uint vtp = pool.getPoolValueInEth();
-      (lowerThreshold, upperThreshold) = getThresholdValues(vtp, vF, getAllSumAssurance(), pd.minCap());
-    }
+  function updateMCR(uint poolValueInEth) external onlyInternal {
```
danoctavian, line 143: untested, want to validate the approach first, will iron out bugs later

this means:
MCR stays static until updated. sell and buy update it at MCR intervals of at least 1 hour (minUpdateTime). between 

Other alternative explored:

make `getMCR` compute the 'view' value in Pool.sol 
getMCR computes the "real" value of the MCR at all times and commits it once in a while to storage (if not written to storage lastTimeUpdated doesn't get changed).

the computation of that value is expensive and if it is incurred on every buy/sell NXM adds 40-50k gas  with this current implementation (some optimizations may be possible).

The added benefit of this is not significant, except that the MCR slowly changes and looks more "smooth" instead of getting bumped up/down sometimes

contracts/modules/capital/MCR.sol
```diff
@@ -19,38 +19,34 @@ import "@openzeppelin/contracts/math/SafeMath.sol";
...
+  uint public mcr;
+  uint public mcrFloor;
+  uint public lastUpdateTime = 0;
+
+  uint public mcrFloorIncrementThreshold = 13000;
+  uint public maxMCRFloorIncrement = 100;
+  uint public maxMCRIncrement = 500;
+  uint public gearingFactor = 48000;
+  uint public minUpdateTime = 3600;
+
```
danoctavian, line 43: the above shall be made configurable through gov

contracts/modules/capital/MCR.sol
```diff
@@ -168,14 +112,11 @@ contract MCR is Iupgradable {
   function getUintParameters(bytes8 code) external view returns (bytes8 codeVal, uint val) {
```
shark0der, line 112: I'd remove this function entirely.

contracts/modules/capital/MCR.sol
roxdanila: what's the max bump?

roxdanila:
@danoctavian unsure if this was mentioned, but I think we need to smooth MCR regardless of what branch is on (i.e. floor vs gearing)

contracts/modules/capital/MCR.sol
```diff
@@ -112,11 +72,28 @@ contract MCR is Iupgradable {
...
+
+    address currentMCR = ms.getLatestAddress("MC");
+
+    if (address(previousMCR) == address(0) || currentMCR != address(this)) {
+      // already initialized or not ready for initialization
+      return;
+    }
+
+    // fetch MCR parameters from previous contract
+    mcrFloor = uint112(previousMCR.variableMincap());
```
danoctavian, line 88: proposal:

set mcrFloor to be previousMCR.getLastMCREther()

since currently in prod there is a discrepancy between the 2. if this code is launched as is, the mcr will gradually decrease to the mcrFloor (Which currently doesn't happen as part of the MCR posting)

test/integration/PooledStaking/rewardsAndBurns.js
```diff
@@ -85,7 +85,7 @@ describe('burns', function () {
       expireTime: '7972408607',
       generationTime: '7972408607001',
       currency,
-      period: 61,
+      period: 63,
```
danoctavian, line 88: unresolved:  no root cause for why this is an issue. only found to fail for the number 61.

PR #59, opened 2021-03-05
add coverId param to getPayoutOutcome

danoctavian:


test/integration/ClaimPayoutAddress/payouts.js
```diff
@@ -85,50 +85,7 @@ describe('send claim payout to the payout address', function () {
 
     assert(actualPayout.eq(expectedPayout), 'should have transfered the cover amount');
   });
-
-  it('[A1, status: 0, 7, 14] CA accept, closed with closeClaim()', async function () {
```
danoctavian, line 89: mistake from prev branch added redundant test for debugging, removing it here

contracts/modules/cover/Cover.sol
```diff
@@ -53,8 +53,8 @@ contract Cover is MasterAware {
   );
```
shark0der, line 53: Index `uint coverId` as well

shark0der: Should we index `coverAsset` as well? So we can filter quickly like "fetch all DAI covers"...?

danoctavian: the limitation here is only 3 indexed fields i already have 3 

    address indexed buyer,
    address indexed contractAddress,
        CoverType indexed coverType,
        
        
        can swawp coverType out for coverAsset would make sense, but can't do more than that

PR #60, opened 2021-03-05
Extend claim submission time

shark0der:


contracts/modules/cover/Quotation.sol
```diff
@@ -69,44 +69,42 @@ contract Quotation is Iupgradable {
...
+    require(coverStatus != uint(QuotationData.CoverStatus.CoverExpired), "Quotation: cover already expired");
 
-    expire = qd.getValidityOfCover(_cid) < uint64(now);
+    if (coverStatus != uint(QuotationData.CoverStatus.ClaimAccepted)) {
+      (,, address contractAddress, bytes4 currency, uint amount,) = qd.getCoverDetailsByCoverID1(coverId);
+      qd.subFromTotalSumAssured(currency, amount);
+      qd.subFromTotalSumAssuredSC(contractAddress, currency, amount);
+    }
 
+    qd.changeCoverStatusNo(coverId, uint8(QuotationData.CoverStatus.CoverExpired));
```
shark0der, line 90: This overwrites the `ClaimAccepted` status, make sure we don't allow new claims to be submitted.

contracts/modules/cover/Quotation.sol
```diff
@@ -69,44 +69,42 @@ contract Quotation is Iupgradable {
...
-    _removeSAFromCSA(_cid, _amount);
+  function withdrawCoverNote(uint[] calldata coverIds, uint[] calldata reasonIndexes) external {
+
+    uint gracePeriod = tc.claimSubmissionGracePeriod();
+    bytes32[] memory reasons = new bytes32[](reasonIndexes.length);
+
+    for (uint i = 0; i < coverIds.length; i++) {
+
+      uint expirationDate = qd.getValidityOfCover(coverIds[i]);
+      require(expirationDate.add(gracePeriod) < now, "Quotation: can't withdraw during grace period");
```
shark0der, line 101: This doesn't check if there's a claim in progress.

contracts/modules/cover/Quotation.sol
```diff
@@ -69,44 +69,42 @@ contract Quotation is Iupgradable {
...
+
+      // note: cover owner is implicitly checked using the reason hash
+      reasons[i] = keccak256(abi.encodePacked("CN", msg.sender, coverIds[i]));
```
shark0der, line 104: Needs a `coverOwner` parameter instead of `msg.sender`.

contracts/modules/token/TokenController.sol
```diff
@@ -240,15 +253,23 @@ contract TokenController is IERC1132, Iupgradable {
   /**
    * @dev Updates Uint Parameters of a code
    * @param code whose details we want to update
-   * @param val value to set
+   * @param value value to set
    */
-  function updateUintParameters(bytes8 code, uint val) public {
+  function updateUintParameters(bytes8 code, uint value) external onlyGovernance {
+
     require(ms.checkIsAuthToGoverned(msg.sender));
```
shark0der, line 260: Remove this

contracts/modules/token/TokenController.sol
```diff
@@ -240,15 +253,23 @@ contract TokenController is IERC1132, Iupgradable {
...
+
     if (code == "MNCLT") {
-      minCALockTime = val.mul(1 days);
-    } else {
-      revert("Invalid param code");
+      minCALockTime = value.mul(1 days);
```
shark0der, line 263: Remove `.mul(1 days)`

contracts/modules/token/TokenController.sol
```diff
@@ -508,6 +529,47 @@ contract TokenController is IERC1132, Iupgradable {
...
+    address _of,
+    bytes32[] calldata _reasons,
+    uint[] calldata _indexes
+  ) external onlyInternal {
+
+    uint reasonCount = lockReason[_of].length;
+    require(reasonCount > 0, "TokenController: no locked cover notes found");
+
+    uint lastReasonIndex = reasonCount.sub(1);
+    uint total = 0;
```
shark0der, line 542: `totalAmount`

contracts/modules/token/TokenController.sol
```diff
@@ -508,6 +529,47 @@ contract TokenController is IERC1132, Iupgradable {
...
+
+    // The iteration is done from the last to first to prevent reason indexes from
+    // changing due to the way we delete the items (copy last to current and pop last).
+    // The provided indexes array must be ordered, otherwise reason index checks will fail.
+
+    for (uint i = _reasons.length; i >= 0; i--) {
```
shark0der, line 548: This underflows

contracts/modules/token/TokenController.sol
```diff
@@ -508,6 +529,47 @@ contract TokenController is IERC1132, Iupgradable {
     emit Unlocked(_of, _reason, _amount);
   }
 
+  function withdrawCoverNote(
```
shark0der, line 532: Add getter for all cover notes

contracts/modules/token/TokenController.sol
```diff
@@ -508,6 +529,47 @@ contract TokenController is IERC1132, Iupgradable {
...
+
+      bytes32 _reason = _reasons[i];
+      uint _index = _indexes[i];
+      require(lockReason[_of][_index] == _reason, "TokenController: bad reason index");
+
+      uint amount = locked[_of][_reason].amount;
+      require(amount != 0, "TokenController: amount is zero");
+
+      total = total.add(amount);
+      locked[_of][_reason].amount = 0;
```
shark0der, line 558: `delete locked[_of][_reason]`

contracts/modules/claims/Claims.sol
```diff
@@ -173,38 +173,29 @@ contract Claims is Iupgradable {
...
-      qd.changeCoverStatusNo(coverId, uint8(QuotationData.CoverStatus.Requested));
-    }
+
+    require(!ms.isPause(), "Claims: System is paused");
+
+    address coverOwner = qd.getCoverMemberAddress(coverId);
+    require(coverOwner == msg.sender, 'Claims: Not cover owner');
+
+    uint coverStatus = qd.getCoverStatusNo(coverId);
+    require(coverStatus != uint8(QuotationData.CoverStatus.ClaimSubmitted), "Claim already submitted");
```
shark0der, line 183: Needs:
1. Claim count per cover
2. Status != ClaimAccepted
3. Need a way to know accepted status for expired covers

contracts/modules/claims/Claims.sol
```diff
@@ -173,38 +173,29 @@ contract Claims is Iupgradable {
...
+
+    uint expirationDate = qd.getValidityOfCover(coverId);
+    uint gracePeriod = tc.claimSubmissionGracePeriod();
+    require(expirationDate.add(gracePeriod) < now, 'Claims: Grace period has expired');
```
shark0der, line 187: `>` not `<`

contracts/modules/cover/Quotation.sol
shark0der:
```
    (, bool isDeposited) = td.depositedCN(coverId);
    require(!isDeposited, "Cover note is deposited and can not be released");
```

contracts/modules/token/TokenController.sol
```diff
@@ -19,18 +19,56 @@ import "@openzeppelin/contracts/math/SafeMath.sol";
...
 
   NXMToken public token;
   IPooledStaking public pooledStaking;
-  uint public minCALockTime = uint(30).mul(1 days);
+
+  uint public minCALockTime;
+  uint public claimSubmissionGracePeriod;
+
+  // coverId => CoverInfo
+  mapping(uint => CoverInfo) public coverInfo;
```
danoctavian, line 52: this cover info being in tokencontroller doesn't make sense in terms of logical location, there's no other proxy contract where it would make more sense though, only some other data contract where you hold cover data

contracts/modules/token/TokenController.sol
```diff
@@ -508,6 +608,50 @@ contract TokenController is IERC1132, Iupgradable {
...
+    uint totalAmount = 0;
+
+    // The iteration is done from the last to first to prevent reason indexes from
+    // changing due to the way we delete the items (copy last to current and pop last).
+    // The provided indexes array must be ordered, otherwise reason index checks will fail.
+
+    for (uint i = _reasons.length; i > 0; i--) {
+
+      bytes32 _reason = _reasons[i - 1];
+      uint _resonIndex = _indexes[i - 1];
```
danoctavian, line 630: typo -> _reasonIndex

contracts/modules/token/TokenController.sol
```diff
@@ -508,6 +608,50 @@ contract TokenController is IERC1132, Iupgradable {
...
+    require(reasonCount > 0, "TokenController: no locked cover notes found");
+
+    uint lastReasonIndex = reasonCount.sub(1);
+    uint totalAmount = 0;
+
+    // The iteration is done from the last to first to prevent reason indexes from
+    // changing due to the way we delete the items (copy last to current and pop last).
+    // The provided indexes array must be ordered, otherwise reason index checks will fail.
+
+    for (uint i = _reasons.length; i > 0; i--) {
```
danoctavian, line 627: find the use of i here as the "real index + 1" a bit confusing since the item you want is at i - 1.

guessing the reason for this is the underflow you would get after it hits 0

contracts/modules/token/TokenController.sol
```diff
@@ -508,6 +608,50 @@ contract TokenController is IERC1132, Iupgradable {
...
+
+      if (lastReasonIndex != _resonIndex) {
+        lockReason[_of][_resonIndex] = lockReason[_of][lastReasonIndex];
+      }
+
+      lockReason[_of].pop();
+      emit Unlocked(_of, _reason, amount);
+
+      if (i > 1) {
+        // if i > 1 then the input has more entries but there are none left in the lockReason array
```
danoctavian, line 647: would make this more clear

"reduce the array size (remove last element). if no items left in _lockReasons, _reasons contains too many items" 

something like that

contracts/modules/cover/Quotation.sol
```diff
@@ -69,44 +69,45 @@ contract Quotation is Iupgradable {
...
 
-  /**
-   * @dev Updates the Sum Assured Amount of all the quotation.
-   * @param _cid Cover id
-   * @param _amount that will get subtracted Current Sum Assured
-   * amount that comes under a quotation.
-   */
-  function removeSAFromCSA(uint _cid, uint _amount) public onlyInternal {
-    _removeSAFromCSA(_cid, _amount);
+  function withdrawCoverNote(address coverOwner, uint[] calldata coverIds, uint[] calldata reasonIndexes) external {
```
danoctavian, line 96: curious how much gas does this use. since the user would execute this to remove his own deposit, if the gas cost doesn't offset  why would he do it? (in the case of small premiums this may be the case)

contracts/modules/claims/Claims.sol
```diff
@@ -173,38 +173,26 @@ contract Claims is Iupgradable {
...
-    for (uint i = firstClaimIndexToSubmitAfterEP; i < lengthOfClaimSubmittedAtEP; i++) {
-      (coverId, dateUpd, submit) = cd.getClaimOfEmergencyPauseByIndex(i);
-      require(submit == false);
-      qadd = qd.getCoverMemberAddress(coverId);
-      _addClaim(coverId, dateUpd, qadd);
-      cd.setClaimSubmittedAtEPTrue(i, true);
-    }
-    cd.setFirstClaimIndexToSubmitAfterEP(lengthOfClaimSubmittedAtEP);
+  function submitClaimAfterEPOff() external pure {
+    revert("Deprecated");
```
danoctavian, line 195: put: "Claims: Deprecated"

contracts/modules/token/TokenController.sol
danoctavian: note on off-band discussion: 
we know that the CoverInfo mapping only gets populated for covers that have at least 1 claim, for others the fields are zeroed out.

Therefore in case we need to migrate this data structure to a different contract, we would only need to migrate the covers that have at least 1 claim which are as of today few in number so it would be feasible in terms of costs

contracts/modules/token/TokenFunctions.sol
```diff
@@ -112,26 +102,33 @@ contract TokenFunctions is Iupgradable {
...
     bytes32 reason = keccak256(abi.encodePacked("CN", _of, coverId));
-    tc.burnLockedTokens(_of, reason, amount);
-    success = true;
+    uint lockedAmount = tc.tokensLocked(_of, reason);
+
+    (uint amount,) = td.depositedCN(coverId);
+    amount = amount.div(2);
+
+    // limit burn amount to actual amount locked
+    uint burnAmount = lockedAmount < amount ? lockedAmount : amount;
```
danoctavian, line 114: is this necessary? suppose it's the first claim, this doesn't apply. suppose it's the 2nd one, then amount.div(2)  would still yield smth that is <= lockedAmount

only way this logic executes is if you're attempting a 3rd burn, do you forsee that happening somehow

contracts/modules/token/TokenController.sol
```diff
@@ -293,6 +373,26 @@ contract TokenController is IERC1132, Iupgradable {
...
+  public
+  view
+  returns (uint256 amount, uint256 validity)
+  {
+
+    bool claimed = locked[_of][_reason].claimed;
+    amount = locked[_of][_reason].amount;
+    validity = locked[_of][_reason].validity;
+
+    if (claimed) {
```
danoctavian, line 391: why is this necessary? wouldn't they be released if it was claimed so it's 0

PR #74, opened 2021-03-25
Cleanup

shark0der:
- Fix compiler warnings
- Fix some solhint issues
- Remove redundant requires
- Remove unused mocks and interfaces


contracts/modules/token/TokenController.sol
```diff
@@ -302,7 +302,7 @@ contract TokenController is LockHandler, Iupgradable {
     if (withdrawableTokens > 0) {
       locked[_of]["CLA"].claimed = true;
       emit Unlocked(_of, "CLA", withdrawableTokens);
-      require(token.transfer(_of, withdrawableTokens), "TokenController: Transfer failed");
+      token.transfer(_of, withdrawableTokens);
```
danoctavian, line 305: why delete the error message?

shark0der: `token.transfer` will never return false. It will either revert without a revert message or return true, so this revert will never throw.

danoctavian (approved):
LGTM

PR #76, opened 2021-03-30
Stacked risk (covered token)

shark0der:


contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+        // check covered protocol
+        require(coveredToken == incident.coveredToken, "Incidents: Covered token != incident token");
+
+        // note: technically this check should have happened at cover purchase time
+        underlyingToken = underlyingTokens[coveredToken];
+        address coverCurrencyAddress = claimsReward().getCurrencyAssetAddress(currency);
+        require(coverCurrencyAddress == underlyingToken, "Incidents: Cover asset != underlying asset");
+      }
+
+      // decrese total sum assured
```
danoctavian, line 139: *decrease

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    {
+      Pool p1 = pool();
+      address payable coverOwnerPayable = address(uint160(coverOwner));
+      address payable payoutAddress = memberRoles().getClaimPayoutAddress(coverOwnerPayable);
+
+      // pull depeged tokens and send the payoutAmount
+      IERC20(coveredToken).safeTransferFrom(msg.sender, address(this), coveredTokenAmount);
+      bool success = p1.sendClaimPayout(underlyingToken, payoutAddress, payoutAmount);
```
danoctavian, line 178: something to keep in mind: this will trigger updateMCR as well in this call

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+      require(success, "Incidents: Payout failed");
+
+      // burn
+      uint burnAmount = p1.getTokenPrice(underlyingToken).mul(payoutAmount);
```
danoctavian, line 182: is there a possibility/incentive of manipulating token price here to reduce your burn?

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+
+  mapping(uint => address payable) contracts;
+
+  Incident[] public incidents;
+
+  // covered token => peg token
+  mapping(address => address) public underlyingTokens;
+
+  // claim id => payout amount
+  mapping(uint => uint) public claimPayouts;
```
danoctavian, line 51: what is this used for?

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+      // decrese total sum assured
+      qd.subFromTotalSumAssured(currency, sumAssured);
+      qd.subFromTotalSumAssuredSC(coveredToken, currency, sumAssured);
+    }
+
+    {
+      // min/max checks. sumAssured assumes 18 decimals
+      uint coverAmount = sumAssured.mul(1e18);
+      uint maxAmount = coverAmount.div(incident.priceBefore);
+      uint minAmount = maxAmount.mul(20).div(100);
```
danoctavian, line 148: can use a named constant here for that 20

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+
+      // decrese total sum assured
+      qd.subFromTotalSumAssured(currency, sumAssured);
+      qd.subFromTotalSumAssuredSC(coveredToken, currency, sumAssured);
+    }
+
+    {
+      // min/max checks. sumAssured assumes 18 decimals
+      uint coverAmount = sumAssured.mul(1e18);
+      uint maxAmount = coverAmount.div(incident.priceBefore);
```
danoctavian, line 147: if you have sumAssured in wad (units) you need to multiply by 1e36
because you then divide by price which is in wei so that maxAmount stays in wei

contracts/modules/cover/Gateway.sol
```diff
@@ -28,7 +28,7 @@ import "../token/TokenFunctions.sol";
 import "./QuotationData.sol";
 import "../claims/ClaimsReward.sol";
 
-contract Cover is MasterAware {
+contract Gateway is MasterAware {
```
danoctavian, line 31: tests naming, and actual deployment aside

there's one place where the Gateway is referred. to by code here:

https://github.com/NexusMutual/distributor-contract/blob/3f4f9f9823cdee2466b4eda2d73ab0b4c77fbf9d/contracts/DistributorFactory.sol#L30


we'd need to update that at the very least to make it compatible with the new "CO" -> "GW"

contracts/modules/claims/Incidents.sol
danoctavian: ah this calls straight into p1 my bad.

this *will need to trigger updateMCR if we are to keep it consistent to the other payout flow.

gas considerations here

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  function addIncident(
+    address coveredToken,
+    uint incidentDate,
+    uint priceBefore,
+    uint priceAfter
+  ) external onlyGovernance {
+
+    require(priceBefore > priceAfter, "Incidents: No depeg");
```
danoctavian, line 68: this condition can be stronger since we have 10% minimum price de-peg in the cover wording. however if we put it here we'd need to do a contract upgrade if we change that.

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+
+  // contract identifiers
+  enum ID {CD, CR, QD, TC, MR, P1, PS}
+
+  mapping(uint => address payable) contracts;
+
+  Incident[] public incidents;
+
+  // covered token => peg token
+  mapping(address => address) public underlyingTokens;
```
danoctavian, line 48: what's the plan for migrating these once we get more than 1 supported underlying token? since this is a proxy contract memory layout will stay the same, we should prepare this multiple underlying tokens

contracts/modules/claims/Incidents.sol
danoctavian, line 182: there should be division by price here. 

Example: 1 NXM = 70 DAI



payoutAmount = 7000 DAI

NXM amount = 7000 / 70 = 100

https://github.com/NexusMutual/smart-contracts/blob/e31f61f609a3ae6470ade4870a17f76cb1649636/contracts/modules/claims/ClaimsReward.sol#L131

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+      address payable payoutAddress = memberRoles().getClaimPayoutAddress(coverOwnerPayable);
+
+      // pull depeged tokens and send the payoutAmount
+      IERC20(coveredToken).safeTransferFrom(msg.sender, address(this), coveredTokenAmount);
+      bool success = p1.sendClaimPayout(underlyingToken, payoutAddress, payoutAmount);
+      require(success, "Incidents: Payout failed");
+
+      // burn
+      uint burnAmount = p1.getTokenPrice(underlyingToken).mul(payoutAmount);
+      pooledStaking().pushBurn(coveredToken, burnAmount);
```
danoctavian, line 183: for the final burnAmount  was the consensus to use a fixed 20% burn rate?

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+import "../claims/ClaimsReward.sol";
+import "../cover/QuotationData.sol";
+import "../governance/MemberRoles.sol";
+import "../token/TokenController.sol";
+
+contract Incidents is MasterAware {
+  using SafeERC20 for IERC20;
+  using SafeMath for uint;
+
+  struct Incident {
```
danoctavian, line 33: will we have a notion of an expired Incident? 

if there is no such thing the latest you would be able to submit for it is 365 days + grace period, so any incident will be valid unti the last cover pre-dating it expires.

to me having some tigher bound there would make sense, although i guess it's debatable. suppose the system is completely dead and people still have the tokens, once can still come at any time for his cover validity to redeem them, since the system is not coming back

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,234 @@
...
+    address coveredToken,
+    address underlyingToken
+  ) external onlyGovernance {
+    require(underlyingTokens[coveredToken] == address(0), "Incidents: Underlying token already set");
+    underlyingTokens[coveredToken] = underlyingToken;
+  }
+
+  function addIncident(
+    address coveredToken,
+    uint incidentDate,
```
danoctavian, line 63: i'd call these timestamp instead of date since it's not really represented as date

shark0der: There's no date type in solidity to confuse this with so I'd leave it as it is

contracts/modules/claims/Incidents.sol
shark0der: This is subjective and there was a discussion whether to include this check in the code or not. Let's discuss this with the team.

contracts/modules/claims/Incidents.sol
shark0der: We have partial payouts so we're storing the payout amount here. This is needed for distributors.

contracts/modules/cover/Quotation.sol
```diff
@@ -207,51 +185,43 @@ contract Quotation is Iupgradable {
...
+    bytes32 hash = getOrderHash(coverDetails, coverPeriod, currency, contractAddress);
     return isValidSignature(hash, _v, _r, _s);
   }
 
   /**
    * @dev Gets order hash for given cover details.
    * @param coverDetails details realted to cover.
    * @param coverPeriod validity of cover.
-   * @param smaratCA smarat contract address.
+   * @param contractAddress smart contract address.
```
danoctavian, line 211: ????  what's wrong with Smarat? did you leave this review?

https://www.tripadvisor.com/ShowUserReviews-g304551-d456058-r6527063-Hotel_Samrat-New_Delhi_National_Capital_Territory_of_Delhi.html

contracts/modules/cover/Quotation.sol
```diff
@@ -275,78 +245,56 @@ contract Quotation is Iupgradable {
...
+    address contractAddress,
+    bytes4 coverCurrency,
     uint[] memory coverDetails,
     uint16 coverPeriod
-  )
-  internal
-  {
+  ) internal {
+
+    address underlyingToken = incidents.underlyingToken(contractAddress);
```
danoctavian, line 261: since you called it productId there, and now it's an id not simply a smart contract address, it makes sense to do that here too

contracts/modules/claims/Incidents.sol
```diff
@@ -0,0 +1,310 @@
...
+      require(coverExpirationDate >= incident.date, "Incidents: Cover end date is after the incident");
+
+      // check grace period
+      uint gracePeriod = tc.claimSubmissionGracePeriod();
+      require(coverExpirationDate.add(gracePeriod) >= block.timestamp, "Incidents: Grace period has expired");
+    }
+
+    {
+      // assumes 18 decimals (eth & dai)
+      uint decimalPrecision = 1e18;
```
danoctavian, line 164: why not const at the top ?

shark0der: because it will call `IERC20(asset).decimals()` at some point instead of being hardcoded so it will end up here

PR #84, opened 2021-05-11
Claim tokens from gateway

horodnicdragos:


contracts/modules/claims/Incidents.sol
```diff
@@ -131,40 +131,60 @@ contract Incidents is MasterAware {
...
       // check cover validity
       require(coverStartDate <= incident.date, "Incidents: Cover start date is before the incident");
       require(coverExpirationDate >= incident.date, "Incidents: Cover end date is after the incident");
 
       // check grace period
-      uint gracePeriod = tc.claimSubmissionGracePeriod();
+      uint gracePeriod;
+      {
+        // definind tc in isolated block to avoid stack too deep
+        TokenController tc = tokenController();
```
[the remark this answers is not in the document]
horodnicdragos: Yes

contracts/modules/cover/Gateway.sol
```diff
@@ -173,10 +182,19 @@ contract Gateway is MasterAware {
   {
     (, uint coverId) = claimsData.getClaimCoverId(claimId);
     (, uint internalClaimStatus) = claimsData.getClaimStatusNumber(claimId);
+    (,address productId) = quotationData.getscAddressOfCover(coverId);
+    address coveredTokenAddress = incidents.coveredToken(productId);
```
shark0der, line 186: The `coveredTokenAddress` and `productId` reads need to happen conditionally. We are not always using their value.

contracts/modules/cover/Gateway.sol
```diff
@@ -161,6 +164,12 @@ contract Gateway is MasterAware {
     return claimId;
   }
 
+  function claimTokens(uint coverId, uint incidentId, uint coveredTokenAmount)
```
shark0der, line 167: We need to proxy the tokens through the Gateway contract. Even though it will take slightly more gas - the gateway contract idea is to be the only interface an integrator would use. If we're not proxying the token transfer via this contract they'd have to approve the Incidents contract instead of the gateway which is confusing.

contracts/modules/cover/Gateway.sol
```diff
@@ -74,6 +76,7 @@ contract Gateway is MasterAware {
     quotationData = QuotationData(master.getLatestAddress("QD"));
     claimsData = ClaimsData(master.getLatestAddress("CD"));
     claims = Claims(master.getLatestAddress("CL"));
+    incidents = Incidents(master.getLatestAddress("IC"));
```
shark0der, line 79: Remove the DAI if condition below, it's already initialized.

horodnicdragos: Removing it seems to break the test since DAI isn't already initialised and thus resulting in a revert: `Cover: unknown asset`. On that note, I'd rename Cover revert reason prefix to Gateway and revert those changes.

contracts/mocks/Disposables/DisposableGateway.sol
```diff
@@ -0,0 +1,32 @@
...
+    quotation = Quotation(master.getLatestAddress("QT"));
+    nxmToken = NXMToken(master.tokenAddress());
+    tokenController = TokenController(master.getLatestAddress("TC"));
+    quotationData = QuotationData(master.getLatestAddress("QD"));
+    claimsData = ClaimsData(master.getLatestAddress("CD"));
+    claims = Claims(master.getLatestAddress("CL"));
+    incidents = Incidents(master.getLatestAddress("IC"));
+    pool = Pool(master.getLatestAddress("P1"));
+    memberRoles = MemberRoles(master.getLatestAddress("MR"));
+    DAI = daiAddress;
```
shark0der, line 30: The only required initializations are the `masterAddress` and `daiAddress`. Everything else will be pulled on upgrade automatically.

contracts/mocks/Disposables/DisposableGateway.sol
```diff
@@ -0,0 +1,32 @@
...
+import "../../modules/token/TokenData.sol";
+import "../../modules/token/TokenFunctions.sol";
+import "../../modules/cover/QuotationData.sol";
+import "../../modules/cover/Gateway.sol";
+import "../../modules/claims/Incidents.sol";
+
+contract DisposableGateway is Gateway {
+
+  constructor() public Gateway() {
+  }
```
shark0der, line 17: Remove the constructor, it is not required

contracts/mocks/EvilContract.sol
```diff
@@ -0,0 +1,77 @@
...
+
+    uint minLength = fallbackValues.length < _fallbackValues.length
+      ? fallbackValues.length
+      : _fallbackValues.length;
+
+    uint maxStorageIndex = minLength > 0 ? minLength - 1 : 0;
+
+    for (uint i = 0; i < maxLength; i++) {
+      if (i > maxStorageIndex) {
+        delete _fallbackTargets[i];
```
shark0der, line 31: FYI: This does not reset the length, just clears the values. The execution will have the target address the zero address and will revert. Leave a todo and we'll fix it next time we're using this contract.

contracts/modules/claims/Incidents.sol
```diff
@@ -131,40 +130,57 @@ contract Incidents is MasterAware {
...
-  ) external returns (uint claimId, uint payoutAmount) {
+  ) external returns (uint claimId, uint payoutAmount, address payoutToken) {
+    (claimId, payoutAmount, payoutToken) = _redeemPayout(coverId, incidentId, coveredTokenAmount, msg.sender);
+  }
 
+  function _redeemPayout(
+    uint coverId,
+    uint incidentId,
+    uint coveredTokenAmount,
+    address owner
```
shark0der, line 154: Please rename `owner` to something like `callingMember` / `caller` / `msgSender` since this is an untrusted input and we don't know if this is the owner yet.

contracts/modules/claims/Incidents.sol
```diff
@@ -184,26 +200,27 @@ contract Incidents is MasterAware {
...
       // mark cover as having a successful claim
       tc.markCoverClaimOpen(coverId);
       tc.markCoverClaimClosed(coverId, true);
 
       // create the claim
       ClaimsData cd = claimsData();
       claimId = cd.actualClaimLength();
-      cd.addClaim(claimId, coverId, coverOwner, now);
-      cd.callClaimEvent(coverId, coverOwner, claimId, now);
+      cd.addClaim(claimId, coverId, owner, now);
```
shark0der, line 211: Can we rename `owner` to `memberAddress`? `owner` is a term used for contract ownership and it's very confusing to see it here.

contracts/modules/cover/Gateway.sol
```diff
@@ -15,16 +15,16 @@
 
 pragma solidity ^0.5.0;
 
+import "hardhat/console.sol";
```
shark0der, line 18: Remove console import

shark0der (changes requested):
see comments

PR #104, opened 2021-06-08
Feature: interfaces for all contracts

danoctavian:


shark0der:
1. Sort imports alphabetically, except for oz imports which stay at the top.
2. Inherit the interface with the same name first, then other inherited interfaces/contracts. Ex: `contract MCR is IMCR, MasterAware`.
3. Replace license text with SPDX identifiers in all contracts, so newer solc won't complain when using them.

PR #117, opened 2021-07-06
Feature/system pause

danoctavian:


contracts/modules/governance/NXMaster.sol
```diff
@@ -20,13 +20,13 @@ import "./external/OwnedUpgradeabilityProxy.sol";
...
 
-  struct EmergencyPause {
+  struct UnusedStruct0 {
     bool pause;
     uint time;
     bytes4 by;
   }
 
-  EmergencyPause[] public emergencyPaused;
+  UnusedStruct0[] public _unused0;
```
shark0der, line 29: You can remove the struct definition and replace all this with just `uint public _unused0;`
The array occupies an entire slot by default.

contracts/modules/governance/NXMaster.sol
```diff
@@ -38,7 +38,10 @@ contract NXMaster is INXMMaster, Governed {
   bool internal reentrancyLock;
   bool public masterInitialized;
   address public owner;
-  uint public pauseTime;
+  uint public _unused1;
+
+  address public emergencyAdmin;
+  bool public paused;
```
shark0der, line 44: I think we can replace `_unused1` with `emergencyAdmin` + `paused` :grin: 
Currently `pauseTime` value is `2419200` which is binary `0b1001001110101000000000`. The first item in a storage slot is stored lower-order aligned (as per solidity docs). I think we can have them in any order and `paused` will be `false` (eight zero bits) and `emergencyAdmin` will be set to a "random" address (whatever we get after conversion) which we can initialize afterwards.

I'm not strongly for this, just an idea. We can fork test it or we can drop it if you consider this as being too dangerous.

PR #120, opened 2021-08-06
Feature assessment v2

horodnicdragos:


contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,754 @@
...
+      msg.value == getSubmissionFee(),
+      "Assessment: Submission fee different that the expected value"
+     );
+    // [todo] Cover premium and total amount need to be obtained from the cover
+    // itself. The premium needs to be converted to NXM using a TWAP at claim time.
+    uint104 coverAmount = 1000 ether;
+    uint16 coverPeriod = 365;
+    Asset asset = Asset.ETH; // take this form cover asset
+    uint80 nxmPriceSnapshot = uint80(1 ether);
+    require(coverPeriod <= 2**11-1, "Assessment: Cover period cannot exceed 2047 days");
```
roxdanila, line 457: this should be enforced at cover level, not here. 
max cover period will be 1 year.

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,754 @@
...
+   *
+   *  @param coverId          Cover identifier
+   *  @param requestedAmount  The amount expected to be received at payout
+   *  @param withProof        When true, a ProofSubmitted event is emitted with ipfsProofHash.
+   *                          When false, no ProofSubmitted event is emitted to save gas if the
+   *                          cover wording doesn't enforce a proof of loss.
+   *  @param ipfsProofHash    The IPFS hash required for proof of loss. It is ignored if withProof
+   *                          is false
+   */
+  function submitClaimForAssessment(
```
roxdanila, line 441: rename to `submitClaim`

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,754 @@
...
+    require(stake.amount != 0, "Assessment: No tokens staked");
+    uint withdrawableAtTimestamp = votesOf[msg.sender][voteCount - 1].timestamp +
+      VOTING_PERIOD_DAYS_MAX + PAYOUT_COOLDOWN_DAYS;
+    require(
+      _blockTimestamp() > withdrawableAtTimestamp,
+      "Assessment: Stake is not withdrawable at the moment"
+     );
+
+    nxm.transferFrom(address(this), msg.sender, stake.amount);
+    stake.amount = 0;
```
roxdanila, line 499: You passed `amount` as a parameter, so you probably want to substract it here.

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,754 @@
...
+    ));
+  }
+
+  function depositStake (uint104 amount) external onlyMember {
+    Stake storage stake = stakeOf[msg.sender];
+    stake.amount += amount;
+    nxm.transferFrom(msg.sender, address(this), amount);
+  }
+
+  function withdrawReward (address user, uint104 untilIndex) external {
```
roxdanila, line 455: does this allow users to withdraw the reward as soon as they voted?

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,708 @@
...
+   *  @param accepted     If the assessor voted to accept the event it's true otherwise it's false
+   *  @param timestamp    Date and time when the vote was cast
+   *  @param tokenWeight  How many tokens were staked when the vote was cast
+   *  @param eventType    Can be a claim or an incident (See EventType enum)
+   */
+  struct Vote {
+    uint104 eventId;
+    bool accepted;
+    uint32 timestamp;
+    uint104 tokenWeight;
```
roxdanila, line 58: this isn't a "weight", but rather the number of tokens that were voted with; it confused me a bit.
I'd call it `stake`

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,708 @@
...
+   *
+   *  @param eventId      Can be either a claimId or an IncidentId
+   *  @param accepted     If the assessor voted to accept the event it's true otherwise it's false
+   *  @param timestamp    Date and time when the vote was cast
+   *  @param tokenWeight  How many tokens were staked when the vote was cast
+   *  @param eventType    Can be a claim or an incident (See EventType enum)
+   */
+  struct Vote {
+    uint104 eventId;
+    bool accepted;
```
roxdanila, line 56: call it "verdict"

horodnicdragos: Accepted is semantically easier to understand and closer to natural language. Booleans in general shouldn't be nouns but adjectives, adverbs etc. They are 99% of the time used in conditionals:

if (vote.accepted) ...
vs
if (vote.verdict) ... / if (vote.verdict == true)

contracts/modules/v2/Assessment.sol
horodnicdragos: https://en.wikipedia.org/wiki/Weighted_voting

Votes are not created equal because a vote from someone with a stake of 100 NXM **weighs** more than a vote from someone with 10 NXM staked. The term weight was also used in the proposal doc and explained in the annotation above. Even if tokenWeight is nothing more than a snapshot of the user's stake, in the context of votes it makes more sense to call it weight as it better describes the purpose of the value (determining how important a vote is rather than what the stake of a voter was) and it also dispels any potential confusion when it's used in formulas that also involve stake in their terms.

contracts/libraries/Assessment/AssessmentClaimsLib.sol
```diff
@@ -0,0 +1,108 @@
...
+        coverPeriod,
+        payoutAsset,
+        nxmPriceSnapshot,
+        CONFIG.CLAIM_ASSESSMENT_DEPOSIT_PERC,
+        false
+      )
+    );
+
+    uint payoutImpact = AssessmentUtilsLib._getPayoutImpactOfClaim(claim.details);
+    claim.poll.end = AssessmentUtilsLib._calculatePollEndDate(CONFIG, claim.poll, payoutImpact);
```
roxdanila, line 73: here claim.poll.end can be set to MIN_DAYS by default when you create the Poll above. this will get rid of the `payoutImpact` variable above

contracts/libraries/Assessment/AssessmentClaimsLib.sol
```diff
@@ -0,0 +1,108 @@
...
+
+    claims.push(claim);
+  }
+
+  function redeemClaimPayout (
+    IAssessment.Configuration calldata CONFIG,
+    IPool pool,
+    IMemberRoles memberRoles,
+    uint104 id,
```
roxdanila, line 82: rename all `id` variables to include the item type (i.e. claimId)

contracts/libraries/Assessment/AssessmentClaimsLib.sol
```diff
@@ -0,0 +1,108 @@
...
+    address payable coverOwner,
```
roxdanila, line 83: this shouldn't be passed as a param, as it need to be validated

contracts/libraries/Assessment/AssessmentClaimsLib.sol
```diff
@@ -0,0 +1,108 @@
...
+
+import "../../interfaces/IMemberRoles.sol";
+import "../../interfaces/IPool.sol";
+import "../../interfaces/IAssessment.sol";
+import "./AssessmentUtilsLib.sol";
+
+library AssessmentClaimsLib {
+
+  // Percentages are defined between 0-10000 i.e. double decimal precision
+  uint16 internal constant PERC_BASIS_POINTS = 10000;
```
roxdanila, line 13: make this `uint` instead of `uint16`

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,397 @@
...
+/**
+ *  Provides a way for cover owners to submit claims and redeem the payouts and facilitates
+ *  assessment processes where members decide the outcome of the events that lead to potential
+ *  payouts.
+ */
+contract Assessment is IAssessment, MasterAwareV2 {
+
+  /* ========== STATE VARIABLES ========== */
+
+  INXMToken internal nxm;
```
roxdanila, line 26: make it immutable

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,397 @@
...
+
+  Configuration public CONFIG;
```
roxdanila, line 28: lowercase `CONFIG`?

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,397 @@
...
+  // members through governance, burnFraud uses this root to burn the fraudulent assessors' stakes
+  // and correct the outcome of the poll.
+  bytes32[] internal fraudMerkleRoots;
+  mapping(uint8 => mapping(uint104 => Poll)) internal pollFraudOfEvent;
+
+  Claim[] public override claims;
+
+  Incident[] public override incidents;
+  mapping(uint104 => address) internal incidentProponent;
+  mapping(uint104 => AffectedToken) internal tokenAffectedByIncident;
```
roxdanila, line 53: why is it needed here? the product should know its token

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,397 @@
...
+  // fraud attempt by one or multiple addresses. Once the root is submitted by adivsory board
+  // members through governance, burnFraud uses this root to burn the fraudulent assessors' stakes
+  // and correct the outcome of the poll.
+  bytes32[] internal fraudMerkleRoots;
+  mapping(uint8 => mapping(uint104 => Poll)) internal pollFraudOfEvent;
+
+  Claim[] public override claims;
+
+  Incident[] public override incidents;
+  mapping(uint104 => address) internal incidentProponent;
```
roxdanila, line 52: shouldn't this be in the incident struct?

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,397 @@
...
+    CONFIG.INCIDENT_IMPACT_ESTIMATE_PERC = 30; // 30%
+    CONFIG.MIN_VOTING_PERIOD_DAYS = 3; // days
+    CONFIG.MAX_VOTING_PERIOD_DAYS = 30; // days
+    CONFIG.PAYOUT_COOLDOWN_DAYS = 1; //days
+    CONFIG.CLAIM_ASSESSMENT_DEPOSIT_PERC = 500; // 5% i.e. 0.05 ETH submission flat fee
+    CONFIG.INCIDENT_ASSESSMENT_DEPOSIT_PERC = 0;
+    addressOfAsset[uint(Asset.ETH)] = eth;
+    addressOfAsset[uint(Asset.DAI)] = dai;
+
+    nxm = INXMToken(master.tokenAddress());
```
roxdanila, line 71: token should become immutable hence should stay here in constructor

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,397 @@
...
+  function changeDependentContractAddress() external override {
+    INXMMaster master = INXMMaster(master);
+    internalContracts[uint(ID.TC)] = master.getLatestAddress("TC");
+    internalContracts[uint(ID.MR)] = master.getLatestAddress("MR");
+    internalContracts[uint(ID.P1)] = master.getLatestAddress("P1");
+  }
+
+}
+
+contract AssessmentViewer is MasterAwareV2 {
```
roxdanila, line 276: move this to a new file

contracts/libraries/Assessment/AssessmentVoteLib.sol
```diff
@@ -0,0 +1,176 @@
...
+    bool accepted,
+    mapping(address => IAssessment.Stake) storage stakeOf,
+    mapping(address => IAssessment.Vote[]) storage votesOf,
+    mapping(bytes32 => bool) storage hasAlreadyVotedOn,
+    IAssessment.Claim[] storage claims,
+    IAssessment.Incident[] storage incidents
+  ) external {
+
+    {
+      bytes32 voteHash = keccak256(abi.encodePacked(id, msg.sender, eventType));
```
roxdanila, line 95: check if mapping(member => mapping(id => boolOrEventTypeToBoolMapping)) uses 2 SLOADS or just one

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,397 @@
...
+
+  // Mapping used to determine if a user has already voted, using a vote hash as a key
+  mapping(bytes32 => bool) public override hasAlreadyVotedOn;
+
+  // An array of merkle tree roots used to indicate fraudulent assessors. Each root represents a
+  // fraud attempt by one or multiple addresses. Once the root is submitted by adivsory board
+  // members through governance, burnFraud uses this root to burn the fraudulent assessors' stakes
+  // and correct the outcome of the poll.
+  bytes32[] internal fraudMerkleRoots;
+  mapping(uint8 => mapping(uint104 => Poll)) internal pollFraudOfEvent;
```
roxdanila, line 47: non uint256 keys in mappings?

contracts/test/Assessment/AssessmentUtilsLibTest.sol
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  function _getCooldownEndDate (
+    IAssessment.Configuration calldata CONFIG,
+    uint32 pollEnd
+  ) external pure returns (uint32) {
+    return AssessmentUtilsLib._getCooldownEndDate(CONFIG, pollEnd);
+  }
+
+  function _calculatePollEndDate (
```
roxdanila, line 50: Do not use a separate function to unpack the variables. Define them in-place (if really needed), ex: `uint96 accepted = poll.accepted;`

if the calls are removed from `submitClaim` and `submitIncident`, we can move this function to `AssessmentVoteLib`

contracts/interfaces/IAssessment.sol
```diff
@@ -0,0 +1,175 @@
...
+    uint16 impactEstimatePerc;
+    // A copy of INCIDENT_ASSESSMENT_DEPOSIT_PERC if it changes while voting is still open.
+    uint16 assessmentDepositPerc;
+    // True when the assessment deposit has already been redeemed and false otherwise.
+    bool depositRedeemed;
+  }
+
+  struct AffectedToken {
+    uint96 priceBefore;
+    address contractAddress;
```
roxdanila, line 132: needs `productId`, not `contractAddress`

contracts/interfaces/IAssessment.sol
```diff
@@ -0,0 +1,175 @@
...
+    uint32 coverId;
+   // Cover period represented as days, used to calculate rewards
+    uint16 coverPeriod;
+   // The asset which is expected at payout. E.g ETH, DAI (See Asset enum)
+    uint8 payoutAsset;
+   // The price (TWAP) of 1 NXM in the covered asset, at claim-time
+    uint80 nxmPriceSnapshot;
+   // A snapshot of CLAIM_ASSESSMENT_DEPOSIT_PERC if it is changed before the payout
+    uint16 assessmentDepositPerc;
+   // True when the payout is redeemed. Prevents further payouts on the claim.
```
roxdanila, line 100: check if deposit is returned as well

contracts/interfaces/IAssessment.sol
```diff
@@ -0,0 +1,175 @@
...
+    // The minimum number of days the users can vote on polls
+    uint8 MIN_VOTING_PERIOD_DAYS;
+    // The maximum number of days the users can vote on polls
+    uint8 MAX_VOTING_PERIOD_DAYS;
+    // Number of days the users must wait after a poll closes, to redeem payouts.
+    uint8 PAYOUT_COOLDOWN_DAYS;
+    // Percentage used to calculate assessment rewards (0-10000 i.e. double decimal precision)
+    uint16 REWARD_PERC;
+    // Percentage used to calculate potential impact of an incident
+    uint16 INCIDENT_IMPACT_ESTIMATE_PERC;
```
roxdanila, line 35: I'd drop "estimate" in the name, it's a bit misleading. 

incident_burn_perc?
incident_payout_perc?
max_payout_perc?

contracts/libraries/Assessment/AssessmentClaimsLib.sol
```diff
@@ -0,0 +1,108 @@
...
+    // [todo] Cover premium and total amount need to be obtained from the cover
+    // itself. The premium needs to be converted to NXM using a TWAP at claim time.
+    {
+      uint96 coverAmount = 1000 ether;
+      require(requestedAmount <= coverAmount, "Cannot claim more than the covered amount");
+    }
+    uint16 coverPeriod = 365;
+    uint8 payoutAsset = uint8(IAssessment.Asset.ETH); // take this form cover asset
+    //uint80 nxmPriceSnapshot = 147573952589676412928; // 1 NXM ~ 147 DAI
+    uint80 nxmPriceSnapshot = uint80(38200000000000000); // 1 NXM ~ 0.0382 ETH
```
danoctavian, line 51: is this a dummy value?

contracts/libraries/Assessment/AssessmentClaimsLib.sol
```diff
@@ -0,0 +1,108 @@
...
+    IAssessment.Claim[] storage claims,
+    mapping(uint => address) storage addressOfAsset
+  ) external {
+    IAssessment.Claim memory claim = claims[id];
+    require(
+      AssessmentUtilsLib._getPollStatus(claim.poll) == IAssessment.PollStatus.ACCEPTED,
+      "The claim must be accepted"
+    );
+    require(
+      block.timestamp >= AssessmentUtilsLib._getCooldownEndDate(CONFIG, claim.poll.end),
```
roxdanila, line 93: could inline this

contracts/libraries/Assessment/AssessmentUtilsLib.sol
```diff
@@ -0,0 +1,90 @@
...
+    return x >= 0 ? x : -x;
+  }
+
+  function min(uint a, uint b) internal pure returns (uint) {
+    return a <= b ? a : b;
+  }
+
+  function pollFraudExists(IAssessment.Poll memory poll) internal pure returns (bool) {
+    return poll.start > 0;
+  }
```
roxdanila, line 22: inline this function

contracts/libraries/Assessment/AssessmentUtilsLib.sol
```diff
@@ -0,0 +1,90 @@
...
+      abs(int(2 * accepted * PRECISION / (accepted + denied)) - int(PRECISION))
+    );
+    uint tokenDrivenStrength = min((accepted + denied) * PRECISION / payoutImpact, 10 * PRECISION) / 10;
+
+    return uint32(start + CONFIG.MIN_VOTING_PERIOD_DAYS * 1 days +
+      (1 * PRECISION - min(consensusDrivenStrength,  tokenDrivenStrength)) *
+      (CONFIG.MAX_VOTING_PERIOD_DAYS * 1 days - CONFIG.MIN_VOTING_PERIOD_DAYS * 1 days) / PRECISION);
+  }
+
+  function _calculatePollEndDate (
```
roxdanila, line 82: Do not use a separate function to unpack the variables. Define them in-place (if really needed), ex: uint96 accepted = poll.accepted;

if the calls are removed from submitClaim and submitIncident, we can move this function to AssessmentVoteLib

contracts/libraries/Assessment/AssessmentClaimsLib.sol
danoctavian, line 93: agree this helper makes it harder to read, can comment and explain here

contracts/libraries/Assessment/AssessmentUtilsLib.sol
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  function _getVoteLockupEndDate (
+    IAssessment.Configuration calldata CONFIG,
+    IAssessment.Vote memory vote
+   ) internal pure returns (uint) {
+    return vote.timestamp + CONFIG.MAX_VOTING_PERIOD_DAYS + CONFIG.PAYOUT_COOLDOWN_DAYS;
+  }
+
+  function _getCooldownEndDate (
```
roxdanila, line 54: could be inlined

contracts/abstract/MasterAwareV2.sol
```diff
@@ -0,0 +1,62 @@
...
+import "../interfaces/INXMMaster.sol";
+
+abstract contract MasterAwareV2 {
+
+  mapping(uint => address payable) internal internalContracts;
+
+  INXMMaster public master;
+
+  // [todo] Are there any missing contracts here?
+  enum ID {GW, GV, MR, CL, CR, MC, P1, QT, TF, PC, PS, TC, IC, AS}
```
roxdanila, line 14: should this be centralized at system level, so we don't duplicate it across various components?

contracts/interfaces/IAssessment.sol
```diff
@@ -0,0 +1,175 @@
...
+    MIN_VOTING_PERIOD_DAYS,
+    MAX_VOTING_PERIOD_DAYS,
+    PAYOUT_COOLDOWN_DAYS,
+    REWARD_PERC,
+    INCIDENT_IMPACT_ESTIMATE_PERC,
+    CLAIM_ASSESSMENT_DEPOSIT_PERC,
+    INCIDENT_ASSESSMENT_DEPOSIT_PERC
+  }
+
+  struct Configuration {
```
roxdanila, line 25: nice!

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,129 @@
...
+      )
+    );
+
+    uint payoutImpact = AssessmentUtilsLib._getPayoutImpactOfIncident(incident.details);
+    incident.poll.end = AssessmentUtilsLib._calculatePollEndDate(CONFIG, incident.poll, payoutImpact);
+
+
+    if (CONFIG.INCIDENT_ASSESSMENT_DEPOSIT_PERC > 0) {
+      uint deposit = payoutImpact * CONFIG.INCIDENT_ASSESSMENT_DEPOSIT_PERC / PERC_BASIS_POINTS;
+      nxm.transferFrom(msg.sender, address(this), deposit);
```
danoctavian, line 82: the function is called getIncidentToSubmit but it has thiss side-effect here of sending NXM (not sure what is the reason for this one)

that's unexpected from a naming perspective

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,397 @@
...
+    ) = AssessmentIncidentsLib.getIncidentToSubmit(
+      CONFIG,
+      nxm,
+      memberRoles(),
+      productId,
+      priceBefore,
+      date
+    );
+
+    AssessmentIncidentsLib.saveIncident (
```
danoctavian, line 149: this function has about 4 lines of code is it possible to inline here? how close are we to max limit of size

contracts/libraries/Assessment/AssessmentVoteLib.sol
```diff
@@ -0,0 +1,176 @@
...
+
+    uint payoutImpact;
+    IAssessment.Poll memory poll;
+    uint32 blockTimestamp = uint32(block.timestamp);
+    if (IAssessment.EventType(eventType) == IAssessment.EventType.CLAIM) {
+      IAssessment.Claim memory claim = claims[id];
+      poll = claims[id].poll;
+      payoutImpact = AssessmentUtilsLib._getPayoutImpactOfClaim(claim.details);
+      require(blockTimestamp < poll.end, "Voting is closed");
+    } else {
```
danoctavian, line 111: if it's not a claim then what is it? what goes down this path

contracts/libraries/Assessment/AssessmentClaimsLib.sol
horodnicdragos: yes

contracts/libraries/Assessment/AssessmentVoteLib.sol
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    stakeOf[user].rewardsWithdrawnUntilIndex = withdrawUntilIndex;
+    nxm.mint(user, rewardToWithdraw);
+  }
+
+  function castVote (
+    IAssessment.Configuration calldata CONFIG,
+    uint8 eventType,
+    uint104 id,
```
danoctavian, line 85: call this claimId

danoctavian, line 85: call this claimId

contracts/interfaces/IAssessment.sol
horodnicdragos: It's a generic estimate used for all incidents (probably 30%). Wouldn't dropping the "estimate" lead you o think that it's value depends on each specific incident rather than applying the same percentage to all?

contracts/interfaces/IAssessment.sol
horodnicdragos: It happens at the same time with payout. If either of them fail, the whole tx fails.

contracts/libraries/Assessment/AssessmentVoteLib.sol
```diff
@@ -0,0 +1,176 @@
...
+    {
+      bytes32 voteHash = keccak256(abi.encodePacked(id, msg.sender, eventType));
+      require(!hasAlreadyVotedOn[voteHash], "Already voted");
+      hasAlreadyVotedOn[voteHash] = true;
+    }
+
+    IAssessment.Stake memory stake = stakeOf[msg.sender];
+    require(stake.amount > 0, "A stake is required to cast votes");
+
+    uint payoutImpact;
```
danoctavian, line 103: what is a payoutImpact ? is it just the amount to be paid out? name is a bit confusing

contracts/interfaces/IAssessment.sol
horodnicdragos: Needs the contract address of the ERC token, not the product id which is a uint24. This is needed to make the actual transfer when redeeming yield token cover payouts.

contracts/libraries/Assessment/AssessmentVoteLib.sol
horodnicdragos: It doesn't, but the bytecode increases significantly and for some reason it also uses more gas (probably because the hash is made on 4 nonpacked keys?).

contracts/libraries/Assessment/AssessmentUtilsLib.sol
```diff
@@ -0,0 +1,90 @@
...
+    IAssessment.Configuration calldata CONFIG,
+    uint96 accepted,
+    uint96 denied,
+    uint32 start,
+    uint payoutImpact
+  ) internal pure returns (uint32) {
+    if (accepted == 0 && denied == 0) {
+      return uint32(start + CONFIG.MIN_VOTING_PERIOD_DAYS * 1 days);
+    }
+
```
danoctavian, line 71: can use some documentation for this function to understand the logic behind it

in the case of tokenDrivenStrength what is the reason to have  '10 * PRECISION' and then divide the whole min(,) by 10?

also not immediately clear why the 2 strengths are the same unit of "measurement" assuming payoutImpact is in NXM

contracts/modules/v2/Assessment.sol
horodnicdragos: Would the struct properties look better in lowercase as well? It could be convenient to have it as it is but I'm happy to change it if you think it's more readable otherwise.

contracts/libraries/Assessment/AssessmentClaimsLib.sol
horodnicdragos: The function name already implies that id belongs to a claim. I would only do this where there there are at least two ids which need to be distinguished.

contracts/libraries/Assessment/AssessmentClaimsLib.sol
```diff
@@ -0,0 +1,108 @@
...
+      "The claim is in cooldown period"
+    );
+    require(!claim.details.payoutRedeemed, "Payout was already redeemed");
```
danoctavian, line 96: is claim.details.payoutRedeemed set to true anywhere? couldn't find it

contracts/libraries/Assessment/AssessmentClaimsLib.sol
```diff
@@ -0,0 +1,108 @@
...
+    claims[id].details.payoutRedeemed = true;
+    // [todo] Destroy and create a new cover nft
+    address payable payoutAddress = memberRoles.getClaimPayoutAddress(coverOwner);
+    address coverAsset = addressOfAsset[uint(IAssessment.Asset.ETH)]; // [todo]
+    bool succeeded = pool.sendClaimPayout(coverAsset, payoutAddress, claim.details.amount);
+    require(succeeded, "Claim payout failed");
+    uint assessmentDepositToRefund = 1 ether * uint(claim.details.assessmentDepositPerc) /
+      uint(PERC_BASIS_POINTS);
+    (bool refunded, /* bytes data */) = payoutAddress.call{value: assessmentDepositToRefund}("");
```
danoctavian, line 105: is this deposit always an ETH deposit? can it be smth else

contracts/abstract/MasterAwareV2.sol
horodnicdragos: Theoretically, having this here should be enough. Every contract inheriting this will use the ID enum at it's own deploy time. If more contracts are added it doesn't matter for contracts which are already deployed since it wouldn't use the new contract anyway. This is just used by getInternalContractAddress and the views in the deriving contract that return the callable contract interface. TLDR, it's just syntactic sugar and bytecode optimization as it doesn't inline the same thing in multiple places.

contracts/libraries/Assessment/AssessmentVoteLib.sol
```diff
@@ -0,0 +1,176 @@
...
+    if (poll.end < blockTimestamp) {
+      // When poll end date falls in the past, replace it with the current block timestamp
+      poll.end = blockTimestamp;
+    }
+
+    // [todo] Add condition when vote shifts poll end in the past and write end with the
+    // current blcok timestamp. Could also consider logic where the consensus is shifted at the
+    // very end of the voting period.
+
+    if (IAssessment.EventType(eventType) == IAssessment.EventType.CLAIM) {
```
danoctavian, line 143: i'm thinking whether incident and claim voting should be separate functions.

while this may save some lines, it makes it a bit harder to reason about each in isolation, also if their logic diverges at some point it can get more complex

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,397 @@
...
+      pool(),
+      memberRoles(),
+      id,
+      coverOwner,
+      claims,
+      addressOfAsset
+    );
+  }
+
+  function redeemIncidentPayout (uint104 incidentId, uint32 coverId, uint payoutAmount) external {
```
danoctavian, line 193: when do you use redeemClaimPayout vs redeemIncidentPayout as a user?

contracts/libraries/Assessment/AssessmentVoteLib.sol
horodnicdragos: The purpose of this rewrite is to use the same assessment mechanism for multiple type of events, namely claims and incidents at the moment, allowing the possibility to add more when needed.

contracts/modules/v2/Assessment.sol
horodnicdragos: Claim is for protocol cover and custodian cover, where you have to make individual claims. Incident is for yield token cover where people don't need to make a claim but just have an accepted incident from which they can redeem their payout.

contracts/libraries/Assessment/AssessmentClaimsLib.sol
horodnicdragos: It's deposited in Assessment.sol. It's ETH for claims. It is returned at payout. There isn't a function to do anything with the deposits from denied claims yet.

contracts/interfaces/IAssessment.sol
roxdanila: > Wouldn't dropping the "estimate" lead you o think that it's value depends on each specific incident rather than applying the same percentage to all?

not really. estimate makes me think it's something variable, not a "flat" number

contracts/libraries/Assessment/AssessmentVoteLib.sol
horodnicdragos: You got it right. When talking about claims it is the requested amount, when talking about incidents it is an estimate of total possible payouts determined by a fixed percentage (probably 30%) and the active cover amount. The two are unified under the term payout impact which best describes in either of the two cases the impact the payouts would have on the pool, hence the importance of the claim/incident. Large payout impacts means the polls are likely to get extended to the maximum period as they require higher participation.

contracts/libraries/Assessment/AssessmentVoteLib.sol
horodnicdragos: It can also be an incident id. A more verbose way to describe that adds no meaningful semantic to the parameter would be eventId.

contracts/modules/v2/Assessment.sol
roxdanila: I'd put `config` in lowercase and the properties in uppercase. 
i.e. `config.PAYOUT_COOLDOWN_DAYS`

shark0der: For the struct properties - I'm fine with both (camelCase or UPPER_CASE). The `config` variable itself IMO looks weird uppercased.

contracts/interfaces/IAssessment.sol
shark0der: The point here is that you'll need to check if the token address is the same as the product token otherwise you could pass WBTC instead of DAI for example (by mistake or malice).

contracts/libraries/Assessment/AssessmentVoteLib.sol
horodnicdragos: It could be made more explicit by expanding the else statement to:
else if (IAssessment.EventType(eventType) == IAssessment.EventType.INCIDENT) {...} else { revert }

contracts/interfaces/IAssessment.sol
shark0der: An "estimate" means we use some formula to estimate it. We're using a percentage to determine it and it's definitely not an estimate but a decision to burn that amount.

contracts/abstract/MasterAwareV2.sol
shark0der: The bytecode optimisation is minimal because the memory is not packed. In both cases (enum vs bytes2) there will be a 32-bytes slot used. Bytecode wise, this will benefit of a negligible optimisation in size because we'll have 1 byte instead of 2 bytes in the places where we're using the abbreviation in code (so if we have 5 usages - we save 5 bytes of bytecode).

The only real benefit is a check at compile-time that we don't mispell the abbreviation. This would be caught by integration tests anyway.

On the other hand we would have to upgrade the master contract and add the `getInternalContractAddress(ID id)` function for this to work.

contracts/interfaces/IAssessment.sol
horodnicdragos: incident_flat_impact_perc?

contracts/modules/v2/Assessment.sol
horodnicdragos: Three things:

- In the possibility where the product can be edited, the incident considers the token that was set at the time when it was proposed.
- Every time someone redeems a payout they don't have to make a call to the contract where the product details are stored, thus saving gas. There is no additional SSTORE for the proposer since we need priceBefore which is stored in the same slot anyway.
- It is also kept separately from incident as this information is only used at payout and in the UI. This means that we don't have an additional SLOAD everywhere else where we use the incident struct.

contracts/modules/v2/Assessment.sol
horodnicdragos: This would only be needed to return the deposit in the eventuality where all members can submit an incident for assessment, not just the AB. Given that it would only be required in one function that returns the deposit for an accepted incident, it means that we don't have an additional SLOAD everywhere else where we use the incident struct.

contracts/abstract/MasterAwareV2.sol
horodnicdragos: I wasn't  talking about the enum at all. The actual bytecode optimisation comes from here:
```  function pool() internal view returns (IPool) {
    return IPool(getInternalContractAddress(ID.P1));
  }
```
Writing `pool()` 2 times instead of `IPool(getInternalContractAddress(ID.P1))` uses a jump to the view instead of inlining the same bytecode twice.

contracts/interfaces/IAssessment.sol
horodnicdragos: That depends on the product. It will be copied from the product when the incident is submitted. This struct actually prevents "mistake or malice" by relying on external data at all times instead of once. In other words it is a snapshot that makes things both safer and cheaper.

contracts/modules/v2/Assessment.sol
horodnicdragos: While this doesn't make any difference gas wise, uint8 is for the event type and uint104 is for the event id. I don't understand the question.

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
horodnicdragos: True, I've updated this but for some reason github doesn't reflect the changes.

contracts/libraries/Assessment/AssessmentClaimsLib.sol
```diff
@@ -0,0 +1,169 @@
...
+pragma solidity ^0.8.0;
+
+import "../../interfaces/IMemberRoles.sol";
+import "../../interfaces/IPool.sol";
+import "../../interfaces/ICover.sol";
+import "../../interfaces/IAssessment.sol";
+import "../../interfaces/IMasterAwareV2.sol";
+import "../../libraries/Assessment/AssessmentVoteLib.sol";
+
+library AssessmentClaimsLib {
```
roxdanila, line 12: why not just ClaimsLib?

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+
+import "../../interfaces/INXMToken.sol";
+import "../../interfaces/IMemberRoles.sol";
+import "../../interfaces/IPool.sol";
+import "../../interfaces/ICover.sol";
+import "../../interfaces/IAssessment.sol";
+import "../../interfaces/IMasterAwareV2.sol";
+import "../../libraries/Assessment/AssessmentVoteLib.sol";
+
+library AssessmentIncidentsLib {
```
roxdanila, line 13: why not just IncidentsLib?

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+  // Percentages are defined between 0-10000 i.e. double decimal precision
+  uint internal constant PERC_BASIS_POINTS = 10000;
+
+  function _getExpectedIncidentPayoutNXM (IAssessment.IncidentDetails memory details)
+  internal pure returns (uint) {
+    return details.activeCoverAmount * details.impactEstimatePerc / PERC_BASIS_POINTS;
```
roxdanila, line 19: rename impactEstimatePerc as discussed

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+    IAssessment.Configuration calldata config,
+    INXMToken nxm,
+    IAssessment.Incident calldata incident
+  ) external {
+    if (config.incidentAssessmentDepositPercentage > 0) {
+      uint expectedPayoutNXM = _getExpectedIncidentPayoutNXM(incident.details);
+      uint deposit = expectedPayoutNXM * config.incidentAssessmentDepositPercentage / PERC_BASIS_POINTS;
+      nxm.transferFrom(msg.sender, address(this), deposit);
+    }
+  }
```
roxdanila, line 67: let's get rid of this

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,276 @@
...
+   *  @param withProof        When true, a ProofSubmitted event is emitted with ipfsProofHash.
+   *                          When false, no ProofSubmitted event is emitted to save gas if the
+   *                          cover wording doesn't enforce a proof of loss.
+   *  @param ipfsProofHash    The IPFS hash required for proof of loss. It is ignored if withProof
+   *                          is false
+   */
+  function submitClaim(
+    uint24 coverId,
+    uint96 requestedAmount,
+    bool withProof,
```
roxdanila, line 116: hasProof

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,276 @@
...
+      ipfsProofHash
+    );
+
+  }
+
+  function submitIncident(
+    uint24 productId,
+    uint96 priceBefore,
+    uint32 date
+  ) external override {
```
roxdanila, line 136: this should only be callable by an AB member

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,276 @@
...
+      withProof,
+      ipfsProofHash
+    );
+
+  }
+
+  function submitIncident(
+    uint24 productId,
+    uint96 priceBefore,
+    uint32 date
```
roxdanila, line 135: isn't this a timestamp?

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,276 @@
...
+  ) external override {
+    (
+      AffectedToken memory affectedToken,
```
roxdanila, line 138: is this the depegged token? i.e. the covered token?

contracts/modules/v2/Assessment.sol
roxdanila: I see this is checked in the library, but it's inconsistent with the submitClaim function above. pick one approach and stick to it

roxdanila: also, why not have a similar submitIncident function in the library, as you did for the claims? let's keep it consistent

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+
+  function saveIncident (
```
roxdanila, line 69: why have a separate function here?

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+    address tokenAddress = 0x0000000000000000000000000000000000000000;
+
+    IAssessment.Incident memory incident = IAssessment.Incident(
+      IAssessment.Poll(0,0,uint32(block.timestamp), 0),
+      IAssessment.IncidentDetails(
+        productId,
+        date,
+        payoutAsset,
+        activeCoverAmount, // NXM
+        config.incidentAssessmentDepositPercentage,
```
roxdanila, line 44: get rid of deposit for incidents everywhere

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+
+  function saveIncident (
+    IAssessment.Incident calldata incident,
+    IAssessment.Incident[] storage incidents,
+    IAssessment.AffectedToken calldata affectedToken,
+    mapping(uint104 => IAssessment.AffectedToken) storage tokenAffectedByIncident
+  ) external {
+    uint104 nextId = uint104(incidents.length);
+    tokenAffectedByIncident[nextId] = affectedToken;
+    incidents.push(incident);
```
roxdanila, line 77: emit an event

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+
+  function redeemIncidentPayout (
+    mapping(uint => address payable) storage internalContracts,
+    IAssessment.Incident calldata incident,
+    uint32 coverId,
+    uint depeggedTokens
+  ) external {
+
+    require(
+      AssessmentVoteLib._getPollStatus(incident.poll) == IAssessment.PollStatus.ACCEPTED,
```
roxdanila, line 88: I'd inline this here to get rid of this dependency

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+    uint104 nextId = uint104(incidents.length);
+    tokenAffectedByIncident[nextId] = affectedToken;
+    incidents.push(incident);
+  }
+
+  function redeemIncidentPayout (
+    mapping(uint => address payable) storage internalContracts,
+    IAssessment.Incident calldata incident,
+    uint32 coverId,
+    uint depeggedTokens
```
roxdanila, line 84: depeggedTokens -> coveredTokenAmount

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,276 @@
...
+    AssessmentClaimsLib.redeemClaimPayout(
+      config,
+      internalContracts,
+      claims,
+      claimants,
+      id
+    );
+  }
+
+  function redeemIncidentPayout (uint104 incidentId, uint32 coverId, uint depeggedTokens)
```
roxdanila, line 189: restricted to members only

contracts/modules/v2/Assessment.sol
```diff
@@ -0,0 +1,276 @@
...
+  external override {
+    AssessmentIncidentsLib.redeemIncidentPayout(
+      internalContracts,
+      incidents[incidentId],
```
roxdanila, line 193: does everything work as expected when the incidentId doesn't exist? should we explictly revert?

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+      (
+        uint24 productId,
+        /*uint8 payoutAsset*/,
+        /*uint8 deniedClaims*/,
+        uint96 amount,
+        uint32 start,
+        uint32 period
+      ) = coverContract.covers(coverId);
+
+      require (coverOwner == msg.sender, "Payout can only be redeemed by cover owner");
```
roxdanila, line 108: must check if the owner is a member too

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+      require(productId == incident.details.productId, "Incident id mismatch");
```
roxdanila, line 109: Product Id mismatch

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+      require(start <= incident.details.date, "Cover start date is after the incident");
+      require(start + period >= incident.details.date, "Cover end date is before the incident");
+      uint gracePeriod = 0; // [todo] Get from product
+      require(start + period + gracePeriod >= block.timestamp, "Grace period has expired");
+      // Should BURN_RATIO & DEDUCTIBLE_RATIO be stored in product details?
+      payoutAmount = depeggedTokens; // [todo] Calculate payout amount
+      require(payoutAmount <= amount, "Payout exceeds covered amount");
+      coverContract.performCoverBurn(coverId, coverOwner, payoutAmount);
```
roxdanila, line 117: performCoverBurn -> pushBurn

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
```diff
@@ -0,0 +1,135 @@
...
+      payoutAddress = memberRolesContract.getClaimPayoutAddress(coverOwner);
+    }
+
+    {
+      IPool poolContract = IPool(internalContracts[uint(IMasterAwareV2.ID.P1)]);
+      address asset = poolContract.assets(incident.details.payoutAsset); // [todo]
+      bool succeeded = poolContract.sendClaimPayout(asset, payoutAddress, payoutAmount);
+      require(succeeded, "Incident payout failed");
+    }
+  }
```
roxdanila, line 134: emit event for successful payout?

roxdanila: must burn / edit the cover nft

roxdanila: do we send the claimant the edited cover nft? or will they redeem it?

horodnicdragos: All of that should happen in performCoverBurn not here. Ignore events for now.

contracts/modules/v2/Assessment.sol
horodnicdragos: It implicitly reverts if the index of an array >= length.

contracts/libraries/Assessment/AssessmentIncidentsLib.sol
horodnicdragos: Do we have any guarantees that we won't have clashes with libraries from modules that have similar functionality? (i.e. 1 xLib in assessment and another xLib in cover)

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,286 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
-
-    AssessmentIncidentsLib.saveIncident(
-      incident,
-      incidents,
-      affectedToken,
-      tokenAffectedByIncident
-    );
-  }
-
   function depositStake(uint96 amount) external override onlyMember {
```
roxdanila, line 69: does it make sense to revert if the given amount is 0?

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,286 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
-    AssessmentClaimsLib.redeemClaimPayout(
-      config,
-      internalContracts,
-      claims,
-      claimants,
-      id
-    );
+  function withdrawStake(uint96 amount) external override onlyMember {
+    Stake storage stake = stakeOf[msg.sender];
+    require(stake.amount != 0, "No tokens staked");
```
roxdanila, line 138: check that `amount <= stake.amount` instead
also - does it make sense to check if amount > 0?

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,286 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+    uint voteCount = votesOf[msg.sender].length;
+    Vote memory vote = votesOf[msg.sender][voteCount - 1];
+    require(
+      block.timestamp > vote.timestamp + config.stakeLockupPeriodDays * 1 days,
+      "Stake is in lockup period"
+     );
+
+    nxm.transferFrom(address(this), msg.sender, amount);
```
roxdanila, line 146: use `operatorTransfer`

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,286 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
 
-  function redeemClaimPayout(uint104 id) external override {
-    AssessmentClaimsLib.redeemClaimPayout(
-      config,
-      internalContracts,
-      claims,
-      claimants,
-      id
-    );
+  function withdrawStake(uint96 amount) external override onlyMember {
```
roxdanila, line 136: move this function after `depositStake`

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,286 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+    for (uint i = stake.rewardsWithdrawnUntilIndex; i < withdrawUntilIndex; i++) {
+      vote = votesOf[user][i];
+      assessment = assessments[vote.assessmentId];
+      if (assessment.poll.end + config.payoutCooldownDays * 1 days >= block.timestamp) {
+        // Poll is not final
+        withdrawUntilIndex = i;
+        break;
+      }
+
+      withdrawn += assessment.totalReward * vote.tokenWeight /
```
roxdanila, line 127: don't you have to declare `withdrawn` first?

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,286 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+  function getRewards(address user) external view
+  returns (uint total, uint withdrawable, uint withdrawableUntilIndex) {
+    Stake memory stake = stakeOf[user];
+    Vote memory vote;
+    Assessment memory assessment;
+    for (uint i = stake.rewardsWithdrawnUntilIndex; i < votesOf[user].length; i++) {
+      vote = votesOf[user][i];
+      assessment = assessments[vote.assessmentId];
+      if (
+        withdrawableUntilIndex == 0 &&
```
roxdanila, line 84: I don't understand this condition, you'll only enter it once at most

contracts/modules/v2/Assessment.sol
roxdanila: I'd go for 
`stake` instead of `depositStake` and 
`unstake` instead of `withdrawStake`

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,286 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
-    AssessmentClaimsLib.redeemCoverForDeniedClaim(
-    config,
-    internalContracts,
-    claims,
-    claimants,
-    coverId,
-    claimId
-    );
-  }
+  function castVote(uint assessmentId, bool isAccepted) external override onlyMember {
```
roxdanila, line 165: `isAccepted` -> `verdict`

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,286 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+        keccak256(abi.encodePacked(assessor, lastFraudulentVoteIndex, burnAmount, fraudCount))
+      ),
+      "Invalid merkle proof"
     );
+
+    Stake memory stake = stakeOf[assessor];
+
+    // Make sure we don't burn beyond lastFraudulentVoteIndex
+    uint processUntil = stake.rewardsWithdrawnUntilIndex + voteBatchSize;
+    if ( processUntil >= lastFraudulentVoteIndex){
```
roxdanila, line 234: nit: spacing

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,286 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+    if (fraudCount == stake.fraudCount) {
+      // Make sure this doesn't revert if the stake amount is already subtracted due to a previous
+      // burn from a different merkle tree.
+      if (stake.amount >= burnAmount) {
+        stake.amount -= burnAmount;
+        nxm.burn(burnAmount);
+      } else {
+        stake.amount = 0;
+        nxm.burn(burnAmount - stake.amount);
+      }
```
roxdanila, line 279: can avoid the if-else

```
burnAmount = burnAmount > stake.amount ? stake.amount : burnAmount 

stake.amount -= burnAmount;
nxm.burn(burnAmount);

```

roxdanila: use TokenController to burn

contracts/modules/v2/Claims.sol
```diff
@@ -0,0 +1,350 @@
...
+      0,
+      coverId,
+      requestedAmount,
+      payoutAsset,
+      config.assessmentBaseDepositRatio,
+      false
+    );
+
+    // Calculate the expected in NXM using the NXM price at cover purchase time
+    uint expectedPayoutNXM = claim.amount * PRECISION / nxmPrice;
```
roxdanila, line 220: this should use `requestedAmount`, not `claim.amount`

contracts/modules/v2/Claims.sol
```diff
@@ -0,0 +1,350 @@
...
+    // [todo] Find out which price to use
+    uint dynamicDeposit;
+    if (payoutAsset > 0) {
+      // If the payout asset is an ERC20 use the currentNxmPrice in that asset
+      uint currentNxmPrice = pool().getTokenPrice(payoutAsset);
+      dynamicDeposit = totalReward * currentNxmPrice / PRECISION;
+    } else {
+      // If the payout asset is ETH use the nxmPrice at the time of cover purchase
+      dynamicDeposit = totalReward * nxmPrice / PRECISION;
+    }
```
roxdanila, line 235:
```
nxmPrice = payoutAsset > 0 ? pool().getTokenPrice(payoutAsset) : nxmPrice
dynamicDeposit = totalReward * nxmPrice / PRECISION;
```

contracts/modules/v2/Claims.sol
```diff
@@ -0,0 +1,350 @@
...
+  function redeemClaimPayout(uint104 claimId) external override {
+    Claim memory claim = claims[claimId];
+    (
+      IAssessment.Poll memory poll,
+      /*uint128 totalAssessmentReward*/,
+      uint assessmentDeposit
+    ) = assessment().assessments(claim.assessmentId);
+
+    require(
+      AssessmentLib._getPollStatus(poll) == IAssessment.PollStatus.ACCEPTED,
```
roxdanila, line 259: can we move this to the Assessment contract?

contracts/modules/v2/Claims.sol
```diff
@@ -0,0 +1,350 @@
...
+
+    // [todo] Replace payoutAddress with the member's address using the member id
+    IMemberRoles memberRolesContract = IMemberRoles(internalContracts[uint(IMasterAwareV2.ID.MR)]);
+    address payable payoutAddress = memberRolesContract.getClaimPayoutAddress(coverOwner);
+
+    bool succeeded = pool().sendClaimPayout(claim.payoutAsset, payoutAddress, claim.amount);
+    require(succeeded, "Claim payout failed");
+
+    {
+      uint assessmentDepositToRefund = 1 ether * uint(claim.assessmentBaseDepositRatio) / RATIO_BPS;
```
roxdanila, line 287: you must refund the dynamic deposit fee as well

contracts/modules/v2/Incidents.sol
```diff
@@ -0,0 +1,188 @@
...
+
+  function submitIncident(
+    uint24 productId,
+    uint96 priceBefore,
+    uint32 date
+  ) external override {
+    require(
+      memberRoles().checkRole(msg.sender, uint(IMemberRoles.Role.AdvisoryBoard)),
+      "Caller must be an advisory board member"
+    );
```
roxdanila, line 77: should this be a modifier to be consistent with `onlyMember`?

contracts/interfaces/IAssessment.sol
```diff
@@ -61,15 +39,13 @@ interface IAssessment is IERC721Receiver {
...
   struct Vote {
    // Identifier of the claim or incident
-    uint104 eventId;
+    uint80 assessmentId;
    // If the assessor votes to accept the event it's true otherwise it's false
     bool accepted;
    // Date and time when the vote was cast
     uint32 timestamp;
    // How many tokens were staked when the vote was cast
     uint96 tokenWeight;
```
roxdanila, line 48: 'weight' is misleading here. alternative suggestions:
- `stakeAmount`
- `tokensStaked`

contracts/modules/v2/Assessment.sol
horodnicdragos: Unsigned integers implicitly revert in this solidity version when underflowing. At the same time uints can never be <0 so there's no need to check the amount.

contracts/modules/v2/Assessment.sol
horodnicdragos: User handholding of sending zero amounts should be offloaded to the UI as long as there's no security benefits on the business side. Checks like this add up to the gas cost and the deployed bytecode size. Check out approve/transfer from OZ's ERC20, they allow 0 amounts.

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,288 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+  /// @param untilIndex  The index until which the rewards should be withdrawn. Used if a large
+  ///                    number of assessments accumulates and the function doesn't fir in one
+  ///                    block, thus requiring multiple batched transactions.
+  function withdrawRewards(address user, uint104 untilIndex) external override
+  returns (uint withdrawn, uint withdrawUntilIndex) {
+    uint104 rewardsWithdrawnUntilIndex = stakeOf[user].rewardsWithdrawnUntilIndex;
+    {
+      uint voteCount = votesOf[user].length;
+      withdrawUntilIndex = untilIndex > 0 ? untilIndex : voteCount;
+      require(rewardsWithdrawnUntilIndex < voteCount, "No withdrawable rewards");
```
roxdanila, line 132: should this condition be `rewardsWithdrawnUntilIndex < withdrawUntilIndex`?

roxdanila: or `rewardsWithdrawnUntilIndex < voteCount - 1`?

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,288 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+        break;
+      }
+
+      withdrawn += assessment.totalReward * vote.stakedAmount /
+        (assessment.poll.accepted + assessment.poll.denied);
+    }
+
+    // This is the index where the next withdrawReward call will start iterating from
+    stakeOf[user].rewardsWithdrawnUntilIndex = uint104(withdrawUntilIndex);
+    ITokenController(getInternalContractAddress(ID.TC)).mint(user, withdrawn);
```
roxdanila, line 152: be careful with minting rewards and switch membership

contracts/modules/v2/Assessment.sol
```diff
@@ -35,238 +27,288 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
-    config,
-    eventType,
-    id,
-    accepted,
-    stakeOf,
-    votesOf,
-    hasAlreadyVotedOn,
-    claims,
-    incidents
+  function castVote(uint assessmentId, bool isAccepted) external override {
```
roxdanila, line 171: also something to be careful with if the member switches addresses - we must make sure the stake is "transferred" to the new address

contracts/modules/v2/Claims.sol
```diff
@@ -0,0 +1,381 @@
...
+        coverPeriod,
+        /*uint80 nxmPrice*/
+      ) = cover().covers(coverId);
+      require(requestedAmount <= coverAmount, "Cannot claim more than the covered amount");
+    }
+
+    {
+      address owner = cover().ownerOf(coverId);
+      claimants.push(owner);
+      cover().transferFrom(owner, address(this), coverId);
```
roxdanila, line 219: do you need to check the transfer succeded?

contracts/modules/v2/Claims.sol
```diff
@@ -0,0 +1,381 @@
...
+   *                          When false, no ProofSubmitted event is emitted to save gas if the
+   *                          cover wording doesn't enforce a proof of loss.
+   *  @param ipfsProofHash    The IPFS hash required for proof of loss. It is ignored if hasProof
+   *                          is false
+   */
+  function submitClaim(
+    uint24 coverId,
+    uint96 requestedAmount,
+    bool hasProof,
+    string calldata ipfsProofHash
```
roxdanila, line 198: can you check if this is empty to avoid passing the extra hasProof flag?

contracts/modules/v2/Claims.sol
```diff
@@ -0,0 +1,381 @@
...
+      block.timestamp >= poll.end + payoutCooldownDays * 1 days,
+      "The claim is in cooldown period"
+    );
+
+    require(
+      block.timestamp < poll.end +
+      payoutCooldownDays * 1 days +
+      config.payoutRedemptionPeriodDays * 1 days,
+      "The redemption period has expired"
+    );
```
roxdanila, line 281: in this case, how do we send the deposit to the pool?

contracts/modules/v2/Claims.sol
roxdanila: seems like the check would be `bytes(string).length > 0`

roxdanila: I'm thinking someone could call this function with a valid proof, but the hadProof flag set to false or smth

contracts/modules/v2/Claims.sol
roxdanila: this also means that you invalidate the whole cover if one didn't redeem on time, even for a partial claim

contracts/modules/v2/Claims.sol
```diff
@@ -0,0 +1,381 @@
...
+      }
+      revert("A cover can be redeemed only if the claim was denied");
+    } else if (block.timestamp < poll.end + payoutCooldownDays * 1 days) {
+      // Sometimes the resolution of a fraudulent assessmnt could potentially change the outcome of
+      // the assessment, thus it is necessary to keep the cover NFT in this contract to be able to
+      // redeem the payout afterwards.
+      revert("The claim is in cooldown period");
+    }
+
+    cover().transferFrom(address(this), claimants[claimId], claim.coverId);
```
roxdanila, line 336: do you need to check this was successful?

contracts/modules/v2/Incidents.sol
```diff
@@ -0,0 +1,236 @@
...
+    uint96 activeCoverAmountInNXM = cover().activeCoverAmountInNXM(productId);
+
+    Incident memory incident = Incident(
+      0, // assessmentId
+      productId,
+      date,
+      priceBefore
+    );
+
+    uint expectedPayoutInNXM = activeCoverAmountInNXM * config.incidentExpectedPayoutRatio *
```
roxdanila, line 87: make sure Anatol is aware he needs to provide this from a staking standpoint

contracts/modules/v2/Claims.sol
horodnicdragos: As long as the cover is active or in grace period, yes.

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,43 @@ interface ICover is IERC721 {
...
+    uint16 productType;
     address productAddress;
+    uint16 capacityFactor;
+    /* supported payout assets bitmap TODO: explain */
+    uint payoutAssets;
+  }
+
+  struct ProductType {
+    string descriptionIpfsHash;
+    uint8 redeemMethod;
```
roxdanila, line 35: I don't see this used anywhere @horodnicdragos

contracts/modules/v2/Incidents.sol
horodnicdragos: @shark0der @danoctavian I think capacity will be stored in Cover.sol, right?

roxdanila (approved):

PR #128, opened 2021-09-06
Feature/cover v2

danoctavian:


contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+
+  Cover[] public override covers;
+  mapping(uint => StakingPool[]) stakingPoolsForCover;
+
+  Product[] public products;
+  mapping(uint => uint) capacityFactors;
+
+  mapping(uint => uint) initialPrices;
+  mapping(uint => mapping(address => uint)) lastPrices;
+  mapping(uint => mapping(address => uint)) lastPriceUpdate;
```
roxdanila, line 22: comments pls, unsure what these are

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+    uint24 productId,
+    uint8 payoutAsset,
+    uint8 deniedClaims,
+    uint96 amount,
+    uint32 period,
+    StakingPool[] memory stakingPools
+  ) internal returns (uint coverId, uint priceInAsset) {
+
+    // convert to NXM amount
+    uint amountToCover = uint(amount) * 1e18 / pool().getTokenPrice(pool().assets(payoutAsset));
```
roxdanila, line 91: `coverAmountInNXM`

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+    uint totalPrice = 0;
+
+    for (uint i = 0; i < stakingPools.length; i++) {
+      if (amountToCover == 0) {
+        break;
+      }
```
roxdanila, line 97: I don't understand this

contracts/interfaces/ICover.sol
```diff
@@ -9,7 +9,7 @@ interface ICover is IERC721 {
   /* ========== DATA STRUCTURES ========== */
 
   struct StakingPool {
-    uint160 id;
+    address poolAddress;
     uint96 bookedAmount;
```
roxdanila, line 13: is this the amount you want to buy from that pool?

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+
+      IStakingPool stakingPool = IStakingPool(stakingPools[i].poolAddress);
+      (uint coveredAmount, uint price) = buyCoverFromPool(stakingPool, productId, amountToCover, period);
+      amountToCover -= coveredAmount;
+      totalPrice += price;
+      stakingPoolsForCover[covers.length].push(StakingPool(address(stakingPool), uint96(coveredAmount)));
+    }
+    require(amountToCover == 0, "Not enough available capacity");
```
roxdanila, line 105: hmm there could actually be races here

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+
+    priceInAsset = totalPrice * pool().getTokenPrice(pool().assets(payoutAsset)) / 1e18;
```
roxdanila, line 107: coverPriceInAsset / coverPriceInCoveredAsset

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+      ));
+
+    coverId = covers.length - 1;
+    _safeMint(msg.sender, coverId);
+  }
+
+  function buyCoverFromPool(
+    IStakingPool stakingPool,
+    uint24 productId,
+    uint amountToCover,
```
roxdanila, line 126: coverAmount

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+  ) internal returns (uint, uint) {
+
+    uint availableCapacity = stakingPool.getAvailableCapacity(productId, capacityFactors[productId]);
+
+    uint coveredAmount;
+    if (amountToCover > availableCapacity) {
+      coveredAmount = availableCapacity;
+    } else {
+      coveredAmount = amountToCover;
+    }
```
roxdanila, line 137: use a ternary instead

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+      period,
+      capacityFactor
+    );
+
+    stakingPoolsForCover[covers.length].push(StakingPool(address(stakingPool), uint96(coveredAmount)));
+
+    return (coveredAmount, price);
+  }
+
+  function extendCover(
```
roxdanila, line 157: we need to walk through all scenarios here and see what happens.

This implementation supports:
- increasing the period
- increasing the amount
- decreasing the amount

It doesn't support:
- decreasing the period

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+
+  /* ========== PRODUCT CONFIGURATION ========== */
+
+  function setCapacityFactor(uint productId, uint capacityFactor) external onlyAdvisoryBoard {
+    capacityFactors[productId] = capacityFactor;
+  }
+
+  function setInitialPrice(uint productId, uint initialPrice) external onlyAdvisoryBoard {
+    initialPrices[productId] = initialPrice;
+  }
```
roxdanila, line 298: we also need a function to list new products, where you configure all settings in one tx

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+      // get price for already paid asset
+      uint priceToBePaid = priceInAsset - priceAlreadyPaid;
+      require(priceToBePaid <= maxPrice, "Cover: Price exceeds maxPrice");
+      retrievePayment(priceToBePaid, cover.payoutAsset);
+    }
+
+    return newCoverId;
+  }
+
+  function incrementDeniedClaims(uint coverId) external onlyInternal override {
```
roxdanila, line 193: to be discussed if it's safe to remove this limitations

contracts/modules/cover/Cover.sol
danoctavian: this assumes the decimal count for asset. need to actually know it.

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+    priceInAsset = totalPrice * pool().getTokenPrice(pool().assets(payoutAsset)) / 1e18;
+
+    covers.push(Cover(
+        productId,
+        payoutAsset,
+        deniedClaims,
+        uint96(amount),
+        uint32(block.timestamp + 1),
+        uint32(period),
+        uint96(priceInAsset)
```
roxdanila, line 116: rename to `premiumNXM`

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+    uint8 payoutAsset,
+    uint8 deniedClaims,
+    uint96 amount,
+    uint32 period,
+    StakingPool[] memory stakingPools
+  ) internal returns (uint coverId, uint priceInAsset) {
+
+    // convert to NXM amount
+    uint amountToCover = uint(amount) * 1e18 / pool().getTokenPrice(pool().assets(payoutAsset));
+    uint totalPrice = 0;
```
roxdanila, line 92: rename to `totalPremium`

contracts/interfaces/ICover.sol
roxdanila: what currency is this in?

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+import "hardhat/console.sol";
+
+
+contract Cover is ICover, ERC721, MasterAwareV2 {
+
+  Cover[] public override covers;
+  mapping(uint => StakingPool[]) stakingPoolsForCover;
+
+  Product[] public products;
+  mapping(uint => uint) capacityFactors;
```
horodnicdragos, line 18: Remove this since we  already have it in Product struct.

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+    uint premiumInAsset = totalPremiumInNXM * pool().getTokenPrice(pool().assets(cover.payoutAsset)) / 1e18;
+
+    require(premiumInAsset <= maxPrice, "Cover: Price exceeds maxPrice");
+    retrievePayment(premiumInAsset, cover.payoutAsset);
+
+    cover.amount += amount;
+    cover.price += uint96(premiumInAsset);
+  }
+
+  function addPeriod(uint coverId, uint32 extraPeriod, uint maxPrice) external {
```
roxdanila, line 224: what is maxPrice? looks like an absolute value in NXM from the implementation - doesn't look like what we want

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+    return 0; // TODO: implement
+  }
+
+  function _createCover(
+    address owner,
+    uint24 productId,
+    uint8 payoutAsset,
+    uint96 amount,
+    uint32 period,
+    StakingPool[] memory stakingPools
```
horodnicdragos, line 86: Use caldata

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+
+    return (coveredAmount, price);
+  }
+
+  function extendCover(
+    uint coverId,
+    uint32 period,
+    uint96 amount,
+    uint maxPrice,
+    StakingPool[] memory stakingPools
```
horodnicdragos, line 160: Use caldata

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    return newCoverId;
+  }
+
+  function addAmount(
+    uint coverId,
+    uint96 amount,
+    uint maxPrice,
+    StakingPool[] memory stakingPools
```
horodnicdragos, line 194: Use caldata

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+      retrievePayment(priceToBePaid, cover.payoutAsset);
+    }
+
+    return newCoverId;
+  }
+
+  function addAmount(
+    uint coverId,
+    uint96 amount,
+    uint maxPrice,
```
roxdanila, line 193: what is maxPrice here? we should either stick to premium % slippage, or absolute annual price % slippage

contracts/mocks/Assessment/AssessmentMockCover.sol
```diff
@@ -57,7 +54,6 @@ contract AssessmentMockCover is ICover, ERC721 {
     address owner,
     uint24 productId,
     uint8 payoutAsset,
-    uint8 deniedClaims,
     uint amount,
     uint period,
     StakingPool[] memory stakingPools
```
horodnicdragos, line 59: Use caldata

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+  /* === MUTATIVE FUNCTIONS ==== */
+
+  function buyCover(
+    address owner,
+    uint24 productId,
+    uint8 payoutAsset,
+    uint96 amount,
+    uint32 period,
+    uint maxPrice,
+    StakingPool[] memory stakingPools
```
horodnicdragos, line 59: Use caldata

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+      stakingPoolsForCover[covers.length].push(StakingPool(address(stakingPool), uint96(coveredAmount)));
+    }
+    require(amountToCover == 0, "Not enough available capacity");
+
+    uint premiumInAsset = totalPremiumInNXM * pool().getTokenPrice(pool().assets(cover.payoutAsset)) / 1e18;
+
+    require(premiumInAsset <= maxPrice, "Cover: Price exceeds maxPrice");
+    retrievePayment(premiumInAsset, cover.payoutAsset);
+
+    cover.amount += amount;
```
roxdanila, line 220: this doesn't looks right. if you do this, the new cover amount will apply to hacks that occurred in the past as well

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+    cover.price += uint96(premiumInAsset);
```
roxdanila, line 221: please keep the naming convention between premium and price

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+      period,
+      capacityFactor
+    );
+
+    stakingPoolsForCover[covers.length].push(StakingPool(address(stakingPool), uint96(coveredAmount)));
+
+    return (coveredAmount, price);
+  }
+
+  function extendCover(
```
roxdanila, line 155: please leave a comment with context on this generic function

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+    require(premiumInAsset <= maxPrice, "Cover: Price exceeds maxPrice");
+    retrievePayment(premiumInAsset, cover.payoutAsset);
+
+    cover.amount += amount;
+    cover.price += uint96(premiumInAsset);
+  }
+
+  function addPeriod(uint coverId, uint32 extraPeriod, uint maxPrice) external {
+    Cover memory cover = covers[coverId];
+    StakingPool[] memory stakingPools = stakingPoolsForCover[covers.length];
```
roxdanila, line 226: note: this assumes extending the cover by booking capacity in the same staking pools the original cover was sold from

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,364 @@
...
+  ) external payable override returns (uint /*coverId*/) {
+    require(initialPrices[productId] != 0, "Cover: product not initialized");
+
+    (uint coverId, uint priceInAsset) = _createCover(owner, productId, payoutAsset, amount, period, stakingPools);
+    require(priceInAsset <= maxPrice, "Cover: Price exceeds maxPrice");
+    retrievePayment(priceInAsset, payoutAsset);
+    return coverId;
+  }
+
+  function createCover(
```
roxdanila, line 69: do you need this?

contracts/modules/cover/Cover.sol
horodnicdragos: I like price btw

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,316 @@
...
+    uint amountToCover = uint(amount) * 1e18 / pool().getTokenPrice(pool().assets(payoutAsset));
+    uint totalPrice = 0;
+
+    for (uint i = 0; i < stakingPools.length; i++) {
+      if (amountToCover == 0) {
+        break;
+      }
+
+      IStakingPool stakingPool = IStakingPool(stakingPools[i].poolAddress);
+      (uint coveredAmount, uint price) = buyCoverFromPool(stakingPool, productId, amountToCover, period);
```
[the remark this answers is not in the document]
horodnicdragos: I assume this is determined at tx time and you'd need to know if you continue buying cover from other pools or the first one(s) get you covered and the rest can be ignored. In other words, the covered amount send by the router can be out of sync with what we've got on chain, hence the chain is the source of truth and that's what will eventually get stored in the stakingPoolsForCover. Correct me if I'm wrong or resolve conversation.

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,44 @@ interface ICover is IERC721 {
   /* ========== DATA STRUCTURES ========== */
 
   struct StakingPool {
-    uint160 id;
-    uint96 bookedAmount;
+    address poolAddress;
```
shark0der, line 12: Optimize using `poolId` and `CREATE2`. We'll need to work on the minimal proxy contract (what I have in mind is not quite standard).

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+   (productId, poolAddress) => lastPrice
+   Last base prices at which a cover was sold by a pool for a particular product.
+  */
+  mapping(uint => mapping(address => uint)) lastPrices;
+
+  /*
+   (productId, poolAddress) => lastPriceUpdate
+   Last base price update time.
+  */
+  mapping(uint => mapping(address => uint)) lastPriceUpdate;
```
shark0der, line 36: These 2 can be packed together. Maybe with `initialPrices` as well?

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+  modifier onlyAdvisoryBoard {
+    uint abRole = uint(IMemberRoles.Role.AdvisoryBoard);
+    require(
+      memberRoles().checkRole(msg.sender, abRole),
+      "Cover: Caller is not an advisory board member"
+    );
+    _;
+  }
+
+  constructor(string memory name_, string memory symbol_) ERC721(name_, symbol_) {
```
shark0der, line 59: We'll need to hardcode the `name` and `symbol` because they should be the same even after upgrades.

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    address owner,
+    uint24 productId,
+    uint8 payoutAsset,
+    uint96 amount,
+    uint32 period,
+    StakingPool[] memory stakingPools
+  ) internal returns (uint coverId, uint premiumInAsset) {
+
+    // convert to NXM amount
+    uint amountLeftToCoverInNXM = uint(amount) * 1e18 / pool().getTokenPrice(pool().assets(payoutAsset));
```
shark0der, line 91: `getTokenPrice` and `pool.assets()` are called multiple times, will need to store them in variables at the optimization stage.

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    activeCoverAmountInNXM[productId] += uint96(amountLeftToCoverInNXM);
```
shark0der, line 92: Todo: add 20% MCR check.

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+
+    uint totalPremiumInNXM = 0;
+
+    for (uint i = 0; i < stakingPools.length; i++) {
+      if (amountLeftToCoverInNXM == 0) {
+        break;
+      }
+
+      IStakingPool stakingPool = IStakingPool(stakingPools[i].poolAddress);
+      (uint coveredAmount, uint premiumInNXM) = buyCoverFromPool(stakingPool, productId, amountLeftToCoverInNXM, period);
```
shark0der, line 102: This currently takes the max amount of each pool instead of using the input and only buying a specified amount from each pool. However, when purchasing a limited capacity out of each pool, if a specified pool doesn't have enough it might revert the whole thing. We'll need to find a middle-ground here.

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    uint24 productId,
+    uint amountToCover,
+    uint32 period
+  ) internal returns (uint, uint) {
+
+    uint availableCapacity = stakingPool.getAvailableCapacity(productId, capacityFactors[productId]);
+
+    uint coveredAmount = amountToCover > availableCapacity ? availableCapacity : amountToCover;
+
+    uint capacityFactor = capacityFactors[productId];
```
shark0der, line 137: Move this to the calling function and pass it as a parameter.

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,44 @@ interface ICover is IERC721 {
   /* ========== DATA STRUCTURES ========== */
 
   struct StakingPool {
```
roxdanila, line 11: suggested rename in offline discussions: `StakingPoolAllocation` / something similar

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,44 @@ interface ICover is IERC721 {
...
     uint24 productId;
     uint8 payoutAsset;
-    uint8 deniedClaims;
     uint96 amount;
     uint32 start;
     uint32 period;  // seconds
+    uint96 premium;
   }
 
   struct Product {
```
roxdanila, line 26: needs and ipfs hash as well for things like product name, any potential exclusions / inclusions from the main cover wording of the product type

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,44 @@ interface ICover is IERC721 {
...
-    uint8 payoutAsset;
+    uint16 productType;
     address productAddress;
+    uint16 capacityFactor;
+    /* supported payout assets bitmap TODO: explain */
+    uint payoutAssets;
+  }
+
+  struct ProductType {
+    string descriptionIpfsHash;
```
roxdanila, line 35: needs a more comprehensive name - this will include name, cover wording, etc.

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,44 @@ interface ICover is IERC721 {
...
     uint96 amount;
     uint32 start;
     uint32 period;  // seconds
+    uint96 premium;
   }
 
   struct Product {
-    uint8 payoutAsset;
+    uint16 productType;
     address productAddress;
```
roxdanila, line 28: does this need to be of type address? seems like we're carrying legacy with us

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,44 @@ interface ICover is IERC721 {
...
+    uint16 capacityFactor;
+    /* supported payout assets bitmap TODO: explain */
+    uint payoutAssets;
+  }
+
+  struct ProductType {
+    string descriptionIpfsHash;
+    uint8 redeemMethod;
```
roxdanila, line 36: this doesn't seem used anywhere so far

contracts/modules/cover/Cover.sol
roxdanila: actually this can just be a global param `capacityFactor` initialized to 2x

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
+
```
roxdanila, line 1: let's keep consistent and have this in a V2 folder?

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+  mapping(uint => StakingPool[]) stakingPoolsForCover;
+
+  Product[] public override products;
+  mapping(uint => uint) capacityFactors;
+
+  ProductType[] public override productTypes;
+
+  mapping(uint => uint) initialPrices;
+
+  mapping(uint => uint96) public override activeCoverAmountInNXM;
```
roxdanila, line 24: please group mappings and arrays separately, for improved readability

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+  uint public REWARD_BPS = 5000;
+  uint public constant PERCENTAGE_CHANGE_PER_DAY_BPS = 100;
+  uint public constant BASIS_PRECISION = 10000;
+  uint public constant STAKE_SPEED_UNIT = 100000e18;
+  uint public constant PRICE_CURVE_EXPONENT = 7;
+  uint public constant MAX_PRICE_PERCENTAGE = 1e20;
+
+  /* === MODIFIERS ==== */
+
+  modifier onlyAdvisoryBoard {
```
roxdanila, line 50: MasterAwareV2 already defines this

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+      actualPrice = actualPrice * activeCover / capacity;
+    }
+    actualPrice = actualPrice / 8 + basePrice * activeCover;
+
+    return actualPrice;
+  }
+
+  /* ========== PRODUCT CONFIGURATION ========== */
+
+  function setCapacityFactor(uint productId, uint capacityFactor) external onlyAdvisoryBoard {
```
roxdanila, line 546: given this becomes a global param, it can be adjusted via governance

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+
+      if (remainder > 0) {
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool ok, /* data */) = address(msg.sender).call{value: remainder}("");
+        require(ok, "Cover: Returning ETH remainder to sender failed.");
+      }
+    } else {
+      address payoutAsset = pool().assets(payoutAssetIndex);
+      IERC20 token = IERC20(payoutAsset);
+      token.transferFrom(msg.sender, address(this), totalPrice);
```
roxdanila, line 468: use `safeTransferFrom`

roxdanila: you need to apply slippage to ERC20 covers as well, right?

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+
+    return newCoverId;
+  }
+
+  function performPayoutBurn(uint coverId, address owner, uint amount) external onlyInternal override {
+    Cover memory cover = covers[coverId];
+  }
+
+
+  function retrievePayment(uint totalPrice, uint8 payoutAssetIndex) internal {
```
roxdanila, line 454: `transferPremiumAsset`?

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+  ) external payable override returns (uint /*coverId*/) {
+    require(initialPrices[productId] != 0, "Cover: product not initialized");
+
+    (uint coverId, uint priceInAsset) = _createCover(owner, productId, payoutAsset, amount, period, stakingPools);
+    require(priceInAsset <= maxPrice, "Cover: Price exceeds maxPrice");
+    retrievePayment(priceInAsset, payoutAsset);
+    return coverId;
+  }
+
+  function _createCover(
```
roxdanila, line 81: inline this function, given now it's only called in one place

contracts/modules/cover/Cover.sol
roxdanila: `amountLeftToCoverInNXM` -> `coverAmountinNXM`

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    );
+    _;
+  }
+
+  constructor(string memory name_, string memory symbol_) ERC721(name_, symbol_) {
+  }
+
+  /* === MUTATIVE FUNCTIONS ==== */
+
+  function buyCover(
```
roxdanila, line 64: you should also mint 10% of the premium in NXM to the buyer - no locking

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    address owner,
+    uint24 productId,
+    uint8 payoutAsset,
+    uint96 amount,
+    uint32 period,
+    uint maxPrice,
+    StakingPool[] memory stakingPools
+  ) external payable override returns (uint /*coverId*/) {
```
roxdanila, line 72: only members should be allowed to buy cover

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    require(initialPrices[productId] != 0, "Cover: product not initialized");
+
+    (uint coverId, uint priceInAsset) = _createCover(owner, productId, payoutAsset, amount, period, stakingPools);
+    require(priceInAsset <= maxPrice, "Cover: Price exceeds maxPrice");
+    retrievePayment(priceInAsset, payoutAsset);
```
roxdanila, line 77: make sure `payoutAsset` is a supported currency for the cover product

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+        break;
+      }
+
+      IStakingPool stakingPool = IStakingPool(stakingPools[i].poolAddress);
+      (uint coveredAmount, uint premiumInNXM) = buyCoverFromPool(stakingPool, cover.productId, amountToCover, period);
+      amountToCover -= coveredAmount;
+      totalPremiumInNXM += premiumInNXM;
+
+      uint j = 0;
+      for ( ; j < currentStakingPools.length; j++) {
```
shark0der, line 237: Instead of searching the pools, pass an additional parameter to the function - an array of indexes, from off-chain, and just check whether at the provided index the pool address is the intended one. Could provide MAX_UINT for the ones that aren't currently used and do the push directly.

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,44 @@ interface ICover is IERC721 {
...
 
   struct StakingPool {
-    uint160 id;
-    uint96 bookedAmount;
+    address poolAddress;
+    uint96 coverAmount;
+    uint96 premiumInNXM;
   }
 
   struct Cover {
```
roxdanila, line 17: naming clash with the contract, we should rename one of them

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    return coverId;
+  }
+
+  function _addAmount(
+    uint coverId,
+    uint96 amount,
+    StakingPool[] memory stakingPools
+  ) internal returns (uint newCoverId, uint premiumInAsset) {
+
+    Cover memory previousCover = covers[coverId];
```
roxdanila, line 216: previousCover -> original cover

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    // clone the existing cover
+    Cover memory cover = covers[coverId];
```
roxdanila, line 218: refactor so you don't need this copy

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+
+    StakingPool[] storage currentStakingPools = stakingPoolsForCover[covers.length];
```
roxdanila, line 220: currentStakingPools -> originalPools

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+      stakingPoolsForCover[covers.length].push(
+        StakingPool(
+          address(stakingPool),
+          uint96(coveredAmount),
+          uint96(premiumInNXM)
+        ));
+    }
+    require(amountToCover == 0, "Not enough available capacity");
+
+    premiumInAsset = totalPremiumInNXM * pool().getTokenPrice(pool().assets(cover.payoutAsset)) / 1e18;
```
roxdanila, line 258: don't read the pool address and price twice. instead do this at the beginning of the function:

```
uint tokenPrice;
{
    IPool _pool = pool();
    tokenPrice = _pool.getTokenPrice(_pool.assets(originalCover.payoutAsset));
}
```

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    // convert to NXM amount
+    uint amountToCover = uint(amount) * 1e18 / pool().getTokenPrice(pool().assets(cover.payoutAsset));
+    uint totalPremiumInNXM = 0;
+    for (uint i = 0; i < stakingPools.length; i++) {
+      if (amountToCover == 0) {
+        break;
+      }
+
+      IStakingPool stakingPool = IStakingPool(stakingPools[i].poolAddress);
+      (uint coveredAmount, uint premiumInNXM) = buyCoverFromPool(stakingPool, cover.productId, amountToCover, period);
```
roxdanila, line 232: period should be renamed to remainingPeriod and calculated as 

`remainingPeriod = originalCover.start + originalCover.period - uint32(block.timestamp)`

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+
+    Cover memory previousCover = covers[coverId];
+    // clone the existing cover
+    Cover memory cover = covers[coverId];
+
+    StakingPool[] storage currentStakingPools = stakingPoolsForCover[covers.length];
+
+    uint32 period = uint32(block.timestamp) - cover.start;
+    // convert to NXM amount
+    uint amountToCover = uint(amount) * 1e18 / pool().getTokenPrice(pool().assets(cover.payoutAsset));
```
roxdanila, line 224: `amountToCover` -> `extraAmountNXM`

contracts/modules/cover/Cover.sol
roxdanila: this will underflow for expired covers, but we'll check that separately anyway

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    uint totalPremiumInNXM = 0;
```
roxdanila, line 225: `totalPremiumInNXM` -> `extraPremiumNXM`

contracts/modules/cover/Cover.sol
roxdanila: `buyCoverFromPool` -> `allocateCapacityFromPool` or smth similar; the pools doesn't "buy" a cover

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,44 @@ interface ICover is IERC721 {
...
     uint24 productId;
     uint8 payoutAsset;
-    uint8 deniedClaims;
     uint96 amount;
     uint32 start;
     uint32 period;  // seconds
+    uint96 premium;
```
shark0der, line 23: Keeping in mind that we already have the premium in NXM in the pools array, do we still need the premium in asset here? We could just emit an event with the paid amount if we want to calculate something off-chain, but if we're not using it anywhere on-chain - let's delete it. One less write, and would fit the cover in a single slot!

contracts/interfaces/ICover.sol
shark0der, line 17: The contract name and the struct name use the same name. This is a bit confusing and also confuses the autocomplete in the IDE. Let's rename one of them.

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+          address(stakingPool),
+          uint96(coveredAmount),
+          uint96(premiumInNXM)
+        ));
+    }
+    require(amountToCover == 0, "Not enough available capacity");
+
+    premiumInAsset = totalPremiumInNXM * pool().getTokenPrice(pool().assets(cover.payoutAsset)) / 1e18;
+
+    // make the previous cover expire at current block
```
roxdanila, line 260: you don't actually do this in storage
```
uint32 elapsedPEriod = originalCover.period - remainingPeriod;
covers[coverId].period = elapsedPeriod;
```

You should also update the premium of the original cover, proportional to the time elapsed:
```
updatedOriginalPremium = originalCover.premium * elapsedPeriod / originalCover.period;
covers[coverId].premium = updatedOriginalPremium;
```

contracts/interfaces/ICover.sol
shark0der: I vote for `CoverChunk`!

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    uint32 newPeriod = uint32(block.timestamp) - cover.start;
+    uint32 previousPeriod = covers[coverId].period;
+
+    cover.amount += amount;
+    cover.premium += uint96(premiumInAsset);
```
roxdanila, line 265: this is wrong. it should add the new premium to the "refunded" premium

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    cover.start = uint32(block.timestamp);
+    cover.period = uint32(block.timestamp) - previousCover.start;
```
roxdanila, line 267: use remainingPeriod defined above

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+
+    covers.push(cover);
```
roxdanila, line 269: you can just inline this:

```
covers.push(
      Cover(
        originalCover.productId,
        originalCover.payoutAsset,
        originalCover.amount + amount,
        uint32(block.timestamp), // start
        remainingPeriod,
        uint96(extraPremiumNXM + carriedPremium)
      )
    );
```

contracts/modules/cover/Cover.sol
roxdanila, line 1: stick to 100 chars / line

contracts/modules/cover/Cover.sol
roxdanila: don't forget to call the staking function to update rewards / capacity

contracts/modules/cover/Cover.sol
roxdanila: don't forget to call the staking function to update rewards / capacity

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+      // get price for already paid asset
+      uint priceToBePaid = priceInAsset - priceAlreadyPaid;
+      require(priceToBePaid <= maxPrice, "Cover: Price exceeds maxPrice");
+      retrievePayment(priceToBePaid, cover.payoutAsset);
+    }
+
+    return newCoverId;
+  }
+
+  function addAmount(
```
roxdanila, line 194: how about `increaseAmount`?

contracts/interfaces/ICover.sol
```diff
@@ -9,27 +9,44 @@ interface ICover is IERC721 {
...
+    uint16 capacityFactor;
+    /* supported payout assets bitmap TODO: explain */
+    uint payoutAssets;
+  }
+
+  struct ProductType {
+    string descriptionIpfsHash;
+    uint8 redeemMethod;
+    uint16 gracePeriodInDays;
+    uint16 burnRatio;
```
roxdanila, line 38: if we go with the latest suggestion from Hugh, to burn proportional to the capacityFactor, then we don't need to store this.

example:
Stake of 100 NXM
Total Capacity of 200 NXM (capacity factor of 2x)
Claim of 80 NXM (example)
Stake burn of 40 NXM

contracts/interfaces/ICover.sol
danoctavian: yes it's a type i negotiated with @horodnicdragos  based on his needs in assessment

contracts/modules/cover/Cover.sol
danoctavian: i want to covey that the payment is being pulled from the user ( maybe `chargePremium` ? ) that was my intention here

that works too but i feel it's too general

contracts/interfaces/ICover.sol
danoctavian: CoverData ?

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,570 @@
...
+    require(msg.sender == ERC721.ownerOf(coverId), "Cover: not cover owner");
+
+    (uint coverId, uint premiumInAsset) = _addAmount(coverId, amount, stakingPools);
+
+    require(premiumInAsset <= maxPrice, "Cover: Price exceeds maxPrice");
+    retrievePayment(premiumInAsset, covers[coverId].payoutAsset);
+    return coverId;
+  }
+
+  function _addAmount(
```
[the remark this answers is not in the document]
danoctavian: yes this check should be for all editing operations

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,588 @@
...
+
+    covers.push(cover);
+
+    newCoverId = covers.length - 1;
+
+    // mint the new cover
+    _safeMint(msg.sender, newCoverId);
+  }
+
+  function addPeriod(uint coverId, uint32 extraPeriod, uint maxPrice) external payable {
```
roxdanila, line 277: `addPeriod` -> `increasePeriod`

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,588 @@
...
+  function _addPeriod(uint coverId, uint32 extraPeriod) internal returns (uint) {
+
+    Cover storage cover = covers[coverId];
+    StakingPool[] storage stakingPools = stakingPoolsForCover[covers.length];
+
+    uint totalPremiumInNXM = 0;
+    for (uint i = 0; i < stakingPools.length; i++) {
+      IStakingPool stakingPool = IStakingPool(stakingPools[i].poolAddress);
+
+      uint capacityFactor = capacityFactors[cover.productId];
```
roxdanila, line 295: not used

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,588 @@
...
+        extraPeriod,
+        REWARD_BPS * (stakingPools[i].premiumInNXM + premiumInNXM) / BASIS_PRECISION,
+        cover.amount
+      );
+
+      totalPremiumInNXM += premiumInNXM;
+      stakingPools[i].premiumInNXM += uint96(premiumInNXM);
+    }
+
+    uint premiumInAsset = totalPremiumInNXM * pool().getTokenPrice(pool().assets(cover.payoutAsset)) / 1e18;
```
roxdanila, line 314: optimization: don't read the pool twice

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,588 @@
...
+
+    cover.period += extraPeriod;
+
+    return premiumInAsset;
+  }
+
+  function addAmountAndReducePeriod(
```
roxdanila, line 321: `addAmountAndReducePeriod` -> `increaseAmountAndReducePeriod`

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,588 @@
...
+      currentStakingPools[i].premiumInNXM = currentStakingPools[i].premiumInNXM * (cover.period - periodReduction) / cover.period;
+    }
+
+    uint refund = cover.premium * periodReduction / cover.period;
+
+    // reduce the cover period before purchasing additional amount
+    cover.period = cover.period - periodReduction;
+    cover.premium = cover.premium - uint96(refund);
+
+    (uint newCoverId, uint premiumInAsset) = _addAmount(coverId, amount, currentStakingPools);
```
roxdanila, line 359: don't you have to pass the input staking pools here?

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,588 @@
...
+    uint premiumInAsset = totalPremiumInNXM * pool().getTokenPrice(pool().assets(cover.payoutAsset)) / 1e18;
+
+    cover.period += extraPeriod;
+
+    return premiumInAsset;
+  }
+
+  function addAmountAndReducePeriod(
+    uint coverId,
+    uint32 periodReduction,
```
roxdanila, line 323: alternative: `periodReduction` -> `periodCut`

contracts/modules/cover/Cover.sol
roxdanila: I like `chargePremium`

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,588 @@
...
+        payoutAsset,
+        uint96(amount),
+        uint32(block.timestamp + 1),
+        uint32(period),
+        uint96(premiumInAsset)
+      ));
+
+    coverId = covers.length - 1;
+    _safeMint(owner, coverId);
+  }
```
roxdanila, line 124: this function is expected to return: `(uint coverId, uint priceInAsset) = _createCover(...)`, but it doesn't

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,588 @@
...
+
+  /* === MUTATIVE FUNCTIONS ==== */
+
+  function buyCover(
+    address owner,
+    uint24 productId,
+    uint8 payoutAsset,
+    uint96 amount,
+    uint32 period,
+    uint maxPrice,
```
roxdanila, line 70: what is `maxPrice`? `maxPremiumInAsset`?

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,588 @@
...
+
+    Cover storage cover = covers[coverId];
+    StakingPool[] storage stakingPools = stakingPoolsForCover[covers.length];
+
+    uint totalPremiumInNXM = 0;
+    for (uint i = 0; i < stakingPools.length; i++) {
+      IStakingPool stakingPool = IStakingPool(stakingPools[i].poolAddress);
+
+      uint capacityFactor = capacityFactors[cover.productId];
+      (uint basePrice, uint premiumInNXM) = getPrice(cover.productId, extraPeriod, cover.productId, stakingPool);
```
roxdanila, line 296: you should exclude the cover amount from the used capacity

roxdanila: you're passing the `productId` instead of the `amount` to the `getPrice` function

contracts/interfaces/ICover.sol
danoctavian: @roxdanila here we have 

```  struct ProductType {
    string descriptionIpfsHash;
    uint8 redeemMethod;
    uint16 gracePeriodInDays;
    uint16 burnRatio;
  }```
  
  where that hassh is contained.
  
  The product points to it through the productType

File: contracts/modules/cover/Cover.sol (hunk 60 above)
Comment (danoctavian) [reply contracts/modules/cover/Cover.sol:214]: @roxdanila  in the _addAmount case i'm adding new Pools in the array, however the existing ones stay unmodified (unless you buy again from an existing pool

calling buyCoverFromPool should update rewards accordingly in that case and there is further readjustment

File: contracts/interfaces/ICover.sol (hunk 59 above)
Comment (danoctavian) [reply contracts/interfaces/ICover.sol:23]: @shark0der  i use this to do calculations for refunds, remaining premiums etc.

https://github.com/NexusMutual/smart-contracts/blob/292ec224f6cc8a98b83ec94b75cb5783a2eed80c/contracts/modules/cover/Cover.sol#L318

this is one location

File: contracts/modules/cover/Cover.sol (hunk 33 above)
Comment (danoctavian) [reply contracts/modules/cover/Cover.sol:88]: we discussed having some sort of slippage for amount, but it seemed to me it's getting complicated when combined with the price slippage, plus it would be an edge case that occurs only when you're at the upper limit of capacity

Comment (roxdanila) [reply contracts/modules/cover/Cover.sol:88]: discussed offline

File: contracts/modules/cover/Cover.sol (hunk 60 above)
Comment (roxdanila) [reply contracts/modules/cover/Cover.sol:213]: you're right, my bad

File: contracts/interfaces/ICover.sol (hunk 36 above)
Comment (roxdanila) [reply contracts/interfaces/ICover.sol:31]: True, but that has generic "product type" information.
I argue that for info specific to the product ID (i.e. specific exclusions for sushiswap for example), we'd need something in the product struct as well.

File: contracts/modules/cover/Cover.sol (hunk 16 above)
Comment (danoctavian) [reply contracts/modules/cover/Cover.sol]: @horodnicdragos  i went back to using memory, since i need to make modifications to it as the loop iterates

File: contracts/modules/cover/Cover.sol (hunk 58 above)
Comment (horodnicdragos) [reply contracts/modules/cover/Cover.sol:188]: I actually prefer InNXM instead of NXM as a suffix.

File: contracts/modules/cover/Cover.sol [hunk 76]
```diff
@@ -0,0 +1,575 @@
...
+    CoverChunkRequest[] memory coverChunkRequests
+  ) external payable override onlyMember returns (uint /*coverId*/) {
+    require(initialPrices[productId] != 0, "Cover: product not initialized");
+
+    uint amountLeftToCoverInNXM;
+    uint tokenPrice;
+    {
+      IPool pool = pool();
+      // convert to NXM amount
+      tokenPrice = pool.getTokenPrice(payoutAsset);
```
horodnicdragos, line 74: pool.getTokenPrice -> pool().getTokenPrice
This means you can remove the lone above.

contracts/modules/cover/CoverNFT.sol
```diff
@@ -0,0 +1,24 @@
+
+import "@openzeppelin/contracts-v4/token/ERC721/ERC721.sol";
+import "../../interfaces/ICover.sol";
+
+contract CoverNFT is ERC721 {
+
+  ICover cover;
```
shark0der, line 7: `cover` is a state variable and no visibility is specified. There's no way to modify it so there's no point in having it in storage - let's make it immutable (cover is a proxy anyway) and let's make it public.

contracts/modules/cover/Cover.sol
```diff
@@ -0,0 +1,576 @@
...
+import "../../abstract/MasterAwareV2.sol";
+import "../../interfaces/IMemberRoles.sol";
+import "../../interfaces/ICoverNFT.sol";
+import "hardhat/console.sol";
+
+contract Cover is ICover, MasterAwareV2 {
+
+  uint public capacityFactor;
+  uint public coverCount;
+
```
shark0der, line 16: Make `capacityFactor` and `coverCount` uint32 and move them near `coverNFT` address so they can use the same slot (32+32+160 = 224).

contracts/interfaces/ICover.sol
```diff
@@ -13,24 +13,29 @@ interface ICover is IERC721 {
...
     uint24 productId;
     uint8 payoutAsset;
     uint96 amount;
     uint32 start;
     uint32 period;  // seconds
-    uint96 price;
+    uint96 premium;
   }
 
   struct Product {
```
shark0der, line 36: Let's pack `initialPrice` and `activeCoverAmountInNXM` in here as well.

PR #131, opened 2021-10-01
Claims ux and tests

horodnicdragos:


contracts/modules/capital/Pool.sol
```diff
@@ -230,15 +230,23 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
+      ok =  _safeTokenTransfer(asset, payoutAddress, amount);
+      uint rate = priceFeedOracle.getAssetToEthRate(asset);
+      require(rate > 0, "Pool: Zero rate");
+      amountInETH = amount.mul(rate).div(1e18);
     }
 
     if (ok) {
       emit Payout(payoutAddress, asset, amount);
+      uint totalAssetValue = getPoolValueInEth();
+      mcr.updateMCRInternal(totalAssetValue.sub(amountInETH), true);
```
roxdanila: you don't need to substract `amountInETH` here, `getPoolValueInEth()` should include it

contracts/modules/v2/Claims.sol
```diff
@@ -325,56 +345,25 @@ contract Claims is IClaims, IERC721Receiver, MasterAwareV2 {
...
+    if (claim.payoutAsset == 0) {
+      payoutSucceeded = pool().sendClaimPayout(
+        claim.payoutAsset,
+        coverOwner,
+        claim.amount + assessmentDeposit
+      );
+    } else {
+      bool depositRefundSucceeded = pool().sendClaimPayout(0, coverOwner, assessmentDeposit);
+      require(depositRefundSucceeded, "Assessment deposit refund failed");
+      payoutSucceeded = pool().sendClaimPayout(claim.payoutAsset, coverOwner, claim.amount);
```
roxdanila, line 363: read `pool()` only once

contracts/mocks/Disposables/DisposableCover.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+pragma solidity ^0.8.0;
+
+import "../../abstract/MasterAwareV2.sol";
+import "../../interfaces/ICover.sol";
+
+contract DisposableCover is MasterAwareV2 {
```
danoctavian, line 8: what's the reason for a DisposableCover ?

contracts/modules/cover/Cover.sol
```diff
@@ -47,18 +43,65 @@ contract Cover is ICover, MasterAwareV2 {
...
+    ) = quotationData.getCoverDetailsByCoverID1(coverId);
+    (
+      /*uint coverId*/,
+      uint8 status,
+      uint sumAssured,
+      uint16 coverPeriodInDays,
+      uint validUntil
+    ) = quotationData.getCoverDetailsByCoverID2(coverId);
+
+    require(msg.sender == coverOwner, "Cover can only be migrated by its owner");
```
danoctavian, line 80: so because of the check the end users need to call this through the distributor contract for covers boguht that way correct? arNFT covers

contracts/modules/cover/ProductsV1.sol
```diff
@@ -0,0 +1,1383 @@
...
+    //"name": "Idle USDT v4 (idleUSDTYield)",
+    //"site": "https://idle.finance/",
+    //"supportedChains": ["ethereum"],
+    //"type": "token",
+    //"symbol": "idleUSDTYield",
+    //"coveredToken": "0xF34842d05A1c888Ca02769A633DF37177415C2f8",
+    //"underlyingToken": "DAI",
+    //"dateAdded": "2021-07-05",
+    //"logo": "idleUSDTYield.svg"
+    if (legacyProductId == 0x0000000000000000000000000000000000000012)  {
```
danoctavian, line 1379: curious what's the gas cost of this last branching path (worst case)

contracts/modules/cover/Cover.sol
```diff
@@ -126,7 +169,7 @@ contract Cover is ICover, MasterAwareV2 {
         uint96(premiumInAsset)
       );
 
-    coverNFT.safeMint(owner, coverId);
+    ICoverNFT(coverNFT).safeMint(owner, coverId);
```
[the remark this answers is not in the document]
horodnicdragos: Storing it as an address allows you to use Cover.sol in contracts which use older solidity versions which do not allow non-primitive return values such as interfaces or structs without using experimental features. At the same time it could very well be stored in contract bytecode since coverNFT will be immutable.

contracts/modules/cover/ProductsV1.sol
horodnicdragos: 26866 gas units for the last one and 22116 gas units for the first one. Subtracting the base gas fee and adding the cost of a call of 2600 it means that the gas cost increase is 8466 and 3716 respectively.

contracts/modules/cover/Cover.sol
```diff
@@ -47,18 +43,65 @@ contract Cover is ICover, MasterAwareV2 {
...
+      /*uint coverId*/,
+      uint8 status,
+      uint sumAssured,
+      uint16 coverPeriodInDays,
+      uint validUntil
+    ) = quotationData.getCoverDetailsByCoverID2(coverId);
+
+    require(msg.sender == coverOwner, "Cover can only be migrated by its owner");
+    require(LegacyCoverStatus(status) != LegacyCoverStatus.Migrated, "Cover has already been migrated");
+    require(LegacyCoverStatus(status) != LegacyCoverStatus.ClaimAccepted, "A claim has already been accepted");
```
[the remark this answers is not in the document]
horodnicdragos: Added checks below this line.

contracts/modules/cover/Cover.sol
horodnicdragos: Correct. The distributor is the current owner of all covers bought through them.

contracts/modules/cover/Cover.sol
```diff
@@ -47,18 +43,65 @@ contract Cover is ICover, MasterAwareV2 {
...
 
-  constructor() {
+  constructor(IQuotationData _quotationData, IProductsV1 _productsV1) {
+    quotationData = _quotationData;
+    productsV1 = _productsV1;
   }
 
-  function initialize(ICoverNFT _coverNFT) public {
-    require(address(coverNFT) == address(0), "Cover: already initialized");
+  function initialize(address _coverNFT) public {
```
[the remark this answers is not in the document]
horodnicdragos: Storing it as an address allows you to use Cover.sol in contracts which use older solidity versions which do not allow non-primitive return values such as interfaces or structs without using experimental features. At the same time it could very well be stored in contract bytecode since coverNFT will be immutable.

contracts/mocks/Disposables/DisposableCover.sol
horodnicdragos: It's used as a shortcut to add products and product types without going through governance since that is not the scope of all tests that rely on the existence of products/product types.

contracts/modules/cover/ProductsV1.sol
danoctavian: ok so the gas costs would be lower if we just do SLOAD, but the relative cost increase to the base cost of the contract call is not that high

contracts/modules/cover/Cover.sol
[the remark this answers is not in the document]
danoctavian: is there a use case for this in our codebase? reading coverNFT from legacy code

PR #132, opened 2021-10-20
Feature/cover v2 updates

danoctavian:


contracts/modules/cover/Cover.sol
```diff
@@ -173,124 +191,111 @@ contract Cover is ICover, MasterAwareV2 {
...
 
-    premiumInAsset = totalPremiumInNXM * tokenPrice / 1e18;
+    premiumInPaymentAsset = totalPremiumInNXM * pool().getTokenPrice(params.paymentAsset) / 1e18;
 
     // make the previous cover expire at current block
     uint32 elapsedPeriod = originalCover.period - remainingPeriod;
-    uint96 updatedOriginalPremium = originalCover.premium * elapsedPeriod / originalCover.period;
-    uint96 carriedPremium = originalCover.premium - updatedOriginalPremium;
+    uint96 updatedOriginalPremium =
+      uint96(originalCover.premium * originalCover.amount / BASIS_PRECISION * elapsedPeriod / originalCover.period);
```
horodnicdragos, line 266: s/premium/premiumRatio

It took me a while to realize it was a ratio.

contracts/modules/cover/Cover.sol
```diff
@@ -173,124 +191,111 @@ contract Cover is ICover, MasterAwareV2 {
...
+    uint96 carriedPremium = originalCover.premium * originalCover.amount - updatedOriginalPremium;
```
horodnicdragos, line 267: s/carriedPremium/newCoverPremium

To be consistent with the variable naming above.

contracts/modules/cover/Cover.sol
```diff
@@ -173,124 +191,111 @@ contract Cover is ICover, MasterAwareV2 {
...
     // mint the new cover
     coverNFT.safeMint(msg.sender, newCoverId);
   }
 
-  function increasePeriod(uint coverId, uint32 extraPeriod, uint maxPremiumInAsset) external payable onlyMember {
+  function increasePeriod(uint coverId, uint32 extraPeriod, uint8 paymentAsset, uint maxPremiumInAsset) external payable onlyMember {
 
 
     CoverData memory cover = covers[coverId];
     require(cover.start + cover.period > block.timestamp, "Cover: cover expired");
```
horodnicdragos, line 290: Capital C

contracts/modules/cover/Cover.sol
```diff
@@ -322,64 +326,62 @@ contract Cover is ICover, MasterAwareV2 {
       coverChunks[i].premiumInNXM += uint96(premiumInNXM);
     }
 
-    uint premiumInAsset = extraPremiumInNXM * pool().getTokenPrice(cover.payoutAsset) / 1e18;
+    uint premiumInAsset = extraPremiumInNXM * pool().getTokenPrice(paymentAsset) / 1e18;
 
     cover.period += extraPeriod;
```
horodnicdragos, line 331: s/extraPeriod/periodExtension

contracts/modules/cover/Cover.sol
```diff
@@ -173,124 +191,111 @@ contract Cover is ICover, MasterAwareV2 {
...
     require(premiumInAsset <= maxPremiumInAsset, "Cover: Price exceeds maxPremiumInAsset");
 
     retrievePayment(premiumInAsset, covers[coverId].payoutAsset);
   }
 
-  function _increasePeriod(uint coverId, uint32 extraPeriod) internal returns (uint) {
+  function _increasePeriod(uint coverId, uint32 extraPeriod, uint8 paymentAsset) internal returns (uint) {
 
     CoverData storage cover = covers[coverId];
     CoverChunk[] storage coverChunks = coverChunksForCover[coverId];
```
horodnicdragos, line 301: It looks like increasing the amount allows you to add an unlimited amount of pools, while increasing the period only allows you to do so using the same pools. How would this work with unstake requests that diminish the capacity? Wouldn't this lead you to the outcome of not being able to increase the period due to the inability of providing a list of pools other than the existing ones?

contracts/modules/cover/Cover.sol
```diff
@@ -60,78 +83,72 @@ contract Cover is ICover, MasterAwareV2 {
...
+    updateActiveCoverAmountInNXM(params.productId, params.period, totalCoverAmountInNXM);
 
     uint coverId = coverCount++;
     covers[coverId] = CoverData(
-        productId,
-        payoutAsset,
-        uint96(amount),
+        params.productId,
+        params.payoutAsset,
+        uint96(totalCoverAmountInNXM * payoutAssetTokenPrice / 1e18),
```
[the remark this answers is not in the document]
horodnicdragos: Are there any scenarios where the the amounts would not match or the user would end up with an amount different than what he initially requested?

PR #137, opened 2021-10-27
Upgrade pool to solidity 0.8 and fix tests

horodnicdragos:


contracts/modules/capital/Pool.sol
```diff
@@ -191,15 +188,15 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
+  function setSwapDetails(
     address assetAddress,
     uint104 _min,
     uint104 _max,
     uint16 _maxSlippageRatio
   ) external onlyGovernance {
 
     require(_min <= _max, "Pool: min > max");
-    require(_maxSlippageRatio <= 1 ether, "Pool: Max slippage ratio > 1");
+    require(_maxSlippageRatio <= 100, "Pool: Max slippage ratio > 1");
```
danoctavian, line 199: to confirm 1 is 1% right?

contracts/modules/capital/Pool.sol
```diff
@@ -308,22 +299,22 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
     uint balance = token.balanceOf(address(this));
     uint transferableAmount = amount > balance ? balance : amount;
 
-    token.safeTransfer(destination, transferableAmount);
+    token.transfer(destination, transferableAmount);
```
danoctavian, line 302: why this change to just transfer

contracts/modules/capital/Pool.sol
horodnicdragos: 1 is 0.01%

shark0der: The original check was to make sure `_maxSlippageRatio` did not exceed 100% (not 1%). This would mean the check should be `_maxSlippageRatio <= 10000` (not 100).

contracts/modules/capital/Pool.sol
```diff
@@ -221,76 +221,34 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
-    // tokens that don't have return data
-    if (returndata.length == 0) {
-      return true;
-    }
+    emit Payout(payoutAddress, asset.assetAddress, amount);
+    uint totalAssetValue = getPoolValueInEth();
 
-    // tokens that have return data will return a bool
-    return abi.decode(returndata, (bool));
+    // [todo] consider adding a minimum sum for triggering an MCR update to save gas
```
shark0der, line 250: Payouts happen less often, and given the complexity to calculate the payout amount in every asset it's probably simpler (and maybe even cheaper) just to update it.

danoctavian (approved):
LGTM

PR #139, opened 2021-10-28
Feature/cover v2 general cover edit

danoctavian:


contracts/modules/cover/Cover.sol
```diff
@@ -415,119 +268,46 @@ contract Cover is ICover, MasterAwareV2 {
...
+      uint refundInPaymentAsset =
+      refundInCoverAsset * 1e18 / pool().getTokenPrice(cover.payoutAsset)
+      * (pool().getTokenPrice(buyCoverParams.payoutAsset) / 10 ** paymentAssetDecimals);
 
-      // write the new staking pool with modified parameters
-      coverChunksForCover[newCoverId].push(newCoverChunks[i]);
+      if (refundInPaymentAsset < premiumInPaymentAsset) {
+        // retrieve extra required payment
+        retrievePayment(premiumInPaymentAsset - refundInPaymentAsset, buyCoverParams.paymentAsset);
+      }
```
horodnicdragos, line 307: Should there be a 10% NXM mint for the paid difference (premiumInPaymentAsset - refundInPaymentAsset) just like in buyCover?

contracts/modules/cover/Cover.sol
```diff
@@ -153,6 +154,19 @@ contract Cover is ICover, MasterAwareV2 {
       "Cover: Payout asset is not supported"
     );
 
+    (uint coverId, uint premiumInPaymentAsset, uint totalPremiumInNXM) = _buyCover(params, coverChunkRequests);
+    require(premiumInPaymentAsset <= params.maxPremiumInAsset, "Cover: Price exceeds maxPremiumInAsset");
+    retrievePayment(premiumInPaymentAsset, params.payoutAsset);
```
horodnicdragos, line 159: What about paying with NXM?

contracts/modules/cover/Cover.sol
```diff
@@ -234,179 +243,23 @@ contract Cover is ICover, MasterAwareV2 {
...
-    );
-
-    // TODO: add check for max 20% MCR check for activeCoverAmountInNXM
-
-    CoverChunk[] storage originalCoverChunks = coverChunksForCover[params.coverId];
-
-    // reduce period
+    uint totalPreviousCoverAmountInNXM = 0;
+    // rollback previous cover
     for (uint i = 0; i < originalCoverChunks.length; i++) {
```
horodnicdragos, line 263: Store originalCoverChunks.length in memory before the loop so yout don't read from storage at every iteration.

contracts/modules/cover/Cover.sol
danoctavian: conifmed yes we want this

contracts/modules/cover/Cover.sol
```diff
@@ -156,11 +156,16 @@ contract Cover is ICover, MasterAwareV2 {
 
     (uint coverId, uint premiumInPaymentAsset, uint totalPremiumInNXM) = _buyCover(params, coverChunkRequests);
     require(premiumInPaymentAsset <= params.maxPremiumInAsset, "Cover: Price exceeds maxPremiumInAsset");
-    retrievePayment(premiumInPaymentAsset, params.payoutAsset);
 
-    // mint 10% of paid premium value to user
-    tokenController().mint(params.owner, totalPremiumInNXM / 10);
+    if (params.payWithNXM) {
+      tokenController().burnFrom(msg.sender, totalPremiumInNXM);
```
horodnicdragos, line 161: This should be 90%, shouldn't it?

danoctavian: we're not doing that anymore

PR #140, opened 2021-11-05
Add the transferToCommunityFund to SwapOperator

shark0der:


contracts/modules/capital/SwapOperator.sol
```diff
@@ -22,10 +22,19 @@ contract SwapOperator is ReentrancyGuard {
...
   }
 
-  ITwapOracle public twapOracle;
-  address public swapController;
-  INXMMaster master;
-  address public stETH;
+  /* storage */
+  bool public communityFundTransferExecuted;
+
+  /* immutables */
```
shark0der, line 28: The changes to the variables were:
1. Reorder so we'd have all variables/constants near each other
2. Make variables initialized in constructor immutable
3. Add `communityFundTransferExecuted` storage variable

contracts/modules/capital/SwapOperator.sol
```diff
@@ -290,6 +294,24 @@ contract SwapOperator is ReentrancyGuard {
     emit Swapped(ETH, stETH, amountIn, amountOut);
   }
 
+  function transferToCommunityFund() external onlySwapController {
```
shark0der, line 297: New function intended for a single use. `communityFundTransferExecuted` ensures it's called only once.
Deadline's role is to prevent re-executing the transaction in case we redeploy the swap operator after a while and forget to remove the function.

danoctavian: note: if we deploy within the deadline, you can still re-execute

shark0der: Yes, of course, but the chance that we'll remember is higher :)
It's intended for longer periods of time, ex redeploying in a few months (when we actually could forget haha)

danoctavian (approved):
LGTM

PR #146, opened 2021-11-17
Fix claim unit tests

horodnicdragos:


contracts/mocks/Assessment/ASMockClaims.sol
```diff
@@ -43,7 +42,7 @@ contract ASMockClaims is MasterAwareV2 {
       false // payoutRedeemed
     );
 
-    uint assessmentId = assessment().startAssessment(0, 0);
+    uint assessmentId = assessment().startAssessment(config.rewardRatio * requestedAmount / 10000, 0);
```
danoctavian, line 45: why was it 0 before? just not implemented yet?

contracts/modules/v2/Assessment.sol
```diff
@@ -70,34 +70,41 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+    for (uint i = rewardsWithdrawableFromIndex; i < voteCount; i++) {
       vote = votesOf[user][i];
       assessment = assessments[vote.assessmentId];
 
-      // If withdrawableUntilIndex has been assigned before, continue calculating the total accrued
+      // If hasReachedUnwithdrawableReward is true, skip and continue calculating the pending total
       // rewards.
       if (
-        withdrawableUntilIndex == 0 &&
+        !hasReachedUnwithdrawableReward &&
```
danoctavian, line 91: why does the condition here change?

before and after withdrawableUntilIndex = i so it would only enter the if when the index is 0, can the index be and you'd want to enter the if?

contracts/modules/v2/Claims.sol
```diff
@@ -86,17 +87,18 @@ contract Claims is IClaims, MasterAwareV2 {
...
-    // Calculate the expected in NXM using the NXM price at cover purchase time
+    // Calculate the expected payout in NXM using the NXM price at cover purchase time
     uint expectedPayoutInNXM = requestedAmount * PRECISION / nxmPriceInPayoutAsset;
 
     // Determine the total rewards that should be minted for the assessors based on cover period
-    uint totalReward = max(
-      config.maxRewardNXM * PRECISION,
-      expectedPayoutInNXM * config.rewardRatio * coverPeriod / 365 days / RATIO_BPS
+    uint totalReward = min(
+      uint(config.maxRewardNXM) * PRECISION,
```
danoctavian, line 95: would be more clear if maxRewardNXM is actually stored as 50e18 because you generally assume all token amounts are expressed in wei on-chain, and when we have things like the  cover amount expressed in units instead of wei it's prone to mistakes

contracts/modules/v2/Claims.sol
```diff
@@ -51,14 +50,16 @@ contract Claims is IClaims, MasterAwareV2 {
   function initialize(address masterAddress) external {
     // The minimum cover premium per year is 2.6%. 20% of the cover premium is: 2.6% * 20% = 0.52%
     config.rewardRatio = 130; // 0.52%
+    config.maxRewardNXM = 50; // 50 NXM
```
danoctavian, line 53: would store this as wei

test/unit/Claims/getAssessmentDepositAndReward.js
```diff
@@ -0,0 +1,193 @@
...
+    {
+      const [, /* deposit */ totalReward] = await claims.getAssessmentDepositAndReward(
+        parseEther('1'),
+        daysToSeconds(30),
+        ASSET.ETH,
+      );
+      assert(totalReward.lte(max));
+    }
+    {
+      const [, /* deposit */ totalReward] = await claims.getAssessmentDepositAndReward(
```
danoctavian, line 23: /* deposit */ should sit before the comma

contracts/modules/v2/Claims.sol
horodnicdragos: Would have made sense for something such as cover. This is only used internally and not meant as some piece of data other contracts rely on. The only time where user input could potentially be wrongly introduced is when the parameter is updated through governance. However given that it's an uint16 it would just revert for values > 65535, so passing 1e18 by mistake would not be possible, not even in the UI. Some methods require multiple parts of config and reading everything in one SLOAD saves gas.

contracts/modules/v2/Assessment.sol
horodnicdragos: The scenario is one withdrawable reward on only one vote (index 0). withdrawableUntilIndex should return voteCount in that case (1). A second bug here was that for two unwithdrawable rewards, it would not stop at index 0 and would have considered vote 0 withdrawable even though it wasn't.

contracts/mocks/Assessment/ASMockClaims.sol
horodnicdragos: I was not using the reward in this mock before.

contracts/modules/v2/Claims.sol
danoctavian: it's more about interpreting it as wei, when it's actually units

The default for any token amounts stored on-chain is to store the amount of wei, at the very let's look for a better name to indicate it's units not wei

danoctavian (approved):
LGTM

PR #156, opened 2022-01-25
Feature/cover tests

danoctavian:


contracts/modules/v2/StakingPool.sol
```diff
@@ -440,16 +446,25 @@ contract StakingPool is ERC20 {
...
+
+    // if the active cover ratio is already above SURGE_THRESHOLD (80%) then apply the surge loading to the entire
+    // value of the cover (surgeFraction = 1). Otherwise apply to the part of the cover that is above the threshold.
+    uint surgeFraction = activeCoverRatio >= SURGE_THRESHOLD_RATIO ? SURGE_DENOMINATOR : surgeLoadingRatio * capacity / amount;
+
+    // apply a base BASE_SURGE_LOADING_RATIO of 10% for each 1% of the surgeLoadingRatio
+    uint surgeLoading =
+      BASE_SURGE_LOADING_RATIO
+      * surgeLoadingRatio / (SURGE_DENOMINATOR / 100)
+      / 2 * surgeFraction / 1e18;
```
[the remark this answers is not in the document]
danoctavian: i'll just write the formula in the comments with explanation for all things

test/unit/Cover/setup.js
```diff
@@ -138,13 +140,13 @@ async function setup () {
   }
 
   // add products
-  await cover.connect(accounts.advisoryBoardMembers[0]).addProduct({
+  await cover.connect(accounts.advisoryBoardMembers[0]).addProducts([{
     productType: '1',
     productAddress: '0x0000000000000000000000000000000000000000',
     coverAssets: '1', // ETH supported
```
horodnicdragos, line 146: CoverMockPool is incorrectly initialized. You want ETH here but the assets array only contains DAI (index 0). So the value of coverAssets of 1 is actually DAI instead of ETH. Make sure you initialize the asset on index 0 to ETH just like in Pool.sol before adding other assets.

horodnicdragos (changes requested):

test/unit/StakingPool/getPrices.js
```diff
@@ -0,0 +1,119 @@
+const { assert } = require('chai');
+const { web3, ethers: { utils: { parseEther } } } = require('hardhat');
+const { time, expectRevert } = require('@openzeppelin/test-helpers');
+const { hex, bnEqual } = require('../utils').helpers;
+
+const accounts = require('../utils').accounts;
+const { calculatePrice, getPrices } = require('./helpers');
+
+const { toBN } = web3.utils;
```
horodnicdragos, line 9: Don't mix web3 with ethers BN types.

test/unit/StakingPool/getPrices.js
```diff
@@ -0,0 +1,119 @@
+const { assert } = require('chai');
+const { web3, ethers: { utils: { parseEther } } } = require('hardhat');
```
horodnicdragos, line 2: Importing both web3 and ethers?

test/unit/StakingPool/getPrices.js
```diff
@@ -0,0 +1,119 @@
...
+const { time, expectRevert } = require('@openzeppelin/test-helpers');
```
horodnicdragos, line 3: Use waffle for reverts. Avoid oz test-helpers since it is something which we're supposed to remove.

https://ethereum-waffle.readthedocs.io/en/latest/matchers.html?highlight=revert#revert-with-message

test/unit/Cover/setup.js
danoctavian: fixed

horodnicdragos:
LGTM

PR #157, opened 2022-01-28
Fix cover segments aware contracts

horodnicdragos:


contracts/interfaces/ICover.sol
```diff
@@ -98,7 +98,24 @@ interface ICover {
...
+    //uint24 productId,
+    //uint8 payoutAsset,
+    //uint96 amountPaidOut
+  //);
+
+  function coverData(uint id) external view returns (CoverData memory);
+
+  function coverSegmentsCount(uint coverId) external view returns (uint);
+
+  //function coverSegments(uint coverId, uint segmentId) external view returns (
```
danoctavian, line 111: is this left commented out to be enabled later?

contracts/modules/cover/CoverViewer.sol
```diff
@@ -0,0 +1,118 @@
...
+      uint segmentCount = cover().coverSegmentsCount(coverId);
+      if (segmentCount == 1) {
+        coverEnd = coverStart + firstSegment.period;
+        amountRemaining = firstSegment.amount;
+      } else {
+        uint lastSegmentStart;
+        uint lastSegmentPeriod;
+        ICover.CoverSegment memory lastSegment = cover().coverSegments(coverId, segmentCount - 1);
+        coverEnd = lastSegment.start + lastSegment.period;
+        amountRemaining = lastSegment.amount;
```
danoctavian, line 63: here the amountRemaining does not take into consideration what was paidOut.

Are you going to compute that in the frontend?

Easier perhaps to keep the same convention and compute the "real" amount here

contracts/interfaces/ICover.sol
danoctavian: if not let's delete

contracts/modules/cover/CoverViewer.sol
horodnicdragos: Look at coverSegments view in Cover.sol

contracts/modules/v2/Claims.sol
```diff
@@ -272,54 +269,46 @@ contract Claims is IClaims, MasterAwareV2 {
...
       (
         /*string descriptionIpfsHash*/,
         uint8 redeemMethod,
         uint16 gracePeriodInDays
       ) = coverContract.productTypes(productType);
       require(redeemMethod == uint8(ICover.RedeemMethod.Claim), "Invalid redeem method");
-      require(requestedAmount <= coverAmount, "Covered amount exceeded");
-      require(coverStart <= block.timestamp, "Cover starts in the future");
+      require(requestedAmount <= segment.amount, "Covered amount exceeded");
+      require(segment.start <= block.timestamp, "Cover starts in the future");
```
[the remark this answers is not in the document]
horodnicdragos: That's the current assumption. We don't know whether a bug allows it to or if we'll actually want this in the future. It shouln't add too much of a gas cost burden and I'd say it's good to have it just in case.

contracts/interfaces/ICover.sol
horodnicdragos: deleted

contracts/modules/cover/CoverViewer.sol
danoctavian: correct it's is already computed

danoctavian (approved):
LGTM

PR #158, opened 2022-01-31
Feature: vote power contract

shark0der:
Exposes an ERC20 interface and returns the amount of tokens held by the owner + balances of all members who delegated. Adds 1 NXM to everyone's balance as per spec. Can only be used against a snapshot as it doesn't check lock times hence doesn't prevent double voting if used on-chain.

### DIFF
--- a/contracts/modules/governance/VotePower.sol
+++ b/contracts/modules/governance/VotePower.sol
@@ -41,7 +41,7 @@ contract VotePower {
       }
     }
 
-    uint balance = tokenController.totalBalanceOf(member) + 1;
+    uint balance = tokenController.totalBalanceOf(member) + 1e18;
     uint[] memory delegationIds = governance.getFollowers(member);
 
     for (uint i = 0; i < delegationIds.length; i++) {
@@ -55,7 +55,7 @@ contract VotePower {
         continue;
       }
 
-      balance += tokenController.totalBalanceOf(follower) + 1;
+      balance += tokenController.totalBalanceOf(follower) + 1e18;
     }
 
     return balance;
--- a/contracts/modules/governance/VotePower.sol
+++ b/contracts/modules/governance/VotePower.sol
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+pragma solidity ^0.8.0;
+
+import "../../interfaces/IGovernance.sol";
+import "../../interfaces/IMemberRoles.sol";
+import "../../interfaces/INXMMaster.sol";
+import "../../interfaces/INXMToken.sol";
+import "../../interfaces/ITokenController.sol";
+
+contract VotePower {
+
+  string constant public name = "NXM balance with delegations";
+  string constant public symbol = "NXMD";
+  uint8 constant public decimals = 18;
+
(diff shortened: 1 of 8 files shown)


danoctavian:
What's the context for this? where is it needed?

contracts/modules/governance/VotePower.sol
```diff
@@ -0,0 +1,64 @@
...
+
+pragma solidity ^0.8.0;
+
+import "../../interfaces/IGovernance.sol";
+import "../../interfaces/IMemberRoles.sol";
+import "../../interfaces/INXMMaster.sol";
+import "../../interfaces/INXMToken.sol";
+import "../../interfaces/ITokenController.sol";
+
+contract VotePower {
```
danoctavian, line 11: have you validated balanceOf and totalSupply is enough for snapshot work ?

I assume it is

contracts/modules/governance/VotePower.sol
```diff
@@ -0,0 +1,64 @@
...
+    uint delegationId = governance.followerDelegation(member);
+
+    if (delegationId != 0) {
+      (, address leader,) = governance.allDelegation(delegationId);
+      if (leader != address(0)) {
+        return 0;
+      }
+    }
+
+    uint balance = tokenController.totalBalanceOf(member) + 1e18;
```
danoctavian, line 44: what is the reason for the `+ 1e18` ?

contracts/modules/governance/VotePower.sol
shark0der: `balanceOf` is the only function that's being called by the `erc20-balance-of` strategy:
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/erc20-balance-of/index.ts#L8-L10

contracts/modules/governance/VotePower.sol
shark0der: Required by the formula: each wallet votes with amount of nxm held + 1 nxm. In other words it's not a tech-related requirement but a governance rules one.

contracts/modules/governance/VotePower.sol
```diff
@@ -0,0 +1,64 @@
...
+    uint[] memory delegationIds = governance.getFollowers(member);
+
+    for (uint i = 0; i < delegationIds.length; i++) {
+
+      (address follower, address leader,) = governance.allDelegation(delegationIds[i]);
+
+      if (
```
danoctavian, line 51: there's an extra check in the `_setVoteTally` code:

https://github.com/NexusMutual/smart-contracts/blob/021f721e3e793fa527badf8dd0a297ab5ca74cd2/contracts/modules/governance/Governance.sol#L930

```(now - _lastUpd) > tokenHoldingTime;```

roxdanila: Snapshot deals with double voting by default, isn't it? So this isn't needed

roxdanila (approved):

contracts/modules/governance/VotePower.sol
danoctavian: lgtm

danoctavian (approved):
LGTM

PR #184, opened 2022-02-21
Add stake migration functions

horodnicdragos:


contracts/modules/claims/LegacyIncidents.sol
```diff
@@ -179,9 +179,9 @@ contract LegacyIncidents is ILegacyIncidents, MasterAware {
     return IPool(internalContracts[uint(ID.P1)]);
   }
 
-  function pooledStaking() internal view returns (IPooledStaking) {
-    return IPooledStaking(internalContracts[uint(ID.PS)]);
-  }
+  //function pooledStaking() internal view returns (IPooledStaking) {
```
danoctavian, line 182: if it's not used we should just delete it

roxdanila (approved):

PR #190, opened 2022-03-14
Staking contract

shark0der:


contracts/modules/v2/StakingPool.sol
```diff
@@ -249,63 +284,79 @@ abstract contract StakingPool is IStakingPool, ERC721 {
...
       period
     );
 
-    currentAllocation += newAllocation;
-    products[productId].currentAllocation = currentAllocation;
+    products[productId].allocatedStake = allocatedStake + newAllocation;
+    products[productId].lastBucket = currentBucket;
 
-    // ceil = fn(a, b) => (a + b - 1) / b
+    // divCeil = fn(a, b) => (a + b - 1) / b
```
horodnicdragos, line 359: Move this to a pure function instead.

contracts/interfaces/IStakingPool.sol
```diff
@@ -6,46 +6,93 @@ import "@openzeppelin/contracts-v4/token/ERC721/IERC721.sol";
...
+
+  /* structs for io */
+
+  struct WithdrawParams {
+    uint positionId;
+    uint[] groupIds;
+    uint flags;
+  }
+
+  struct ProductParams {
```
horodnicdragos, line 45: Move structs outside of the interface so they can be used by other contracts by importing this file and not having to specify the name of the interface for each of them.

contracts/modules/v2/StakingPool.sol
```diff
@@ -94,17 +98,48 @@ abstract contract StakingPool is IStakingPool, ERC721 {
   uint constant REWARDS_DENOMINATOR = 100;
   uint constant WEIGHT_DENOMINATOR = 100;
 
+  // product params flags
+  uint constant FLAG_PRODUCT_WEIGHT = 1;
```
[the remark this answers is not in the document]
shark0der: I was thinking to unify the `setWeights` and `setTargetPrice` in a single function with this signature:  
`function setProductDetails(ProductParams[] memory params)`.

The `ProductParams` contains both the weight and the target price and the uint flags would signal what updates that are intended to be applied:
```
struct ProductParams {
  uint productId;
  uint weight;
  uint targetPrice;
  uint flags; // for example: FLAG_PRODUCT_WEIGHT & FLAG_PRODUCT_PRICE
}
```

contracts/modules/v2/StakingPool.sol
shark0der: Will move when it will be used in more than one place, a comment is sufficient until then.

contracts/interfaces/IStakingPool.sol
shark0der: Done

contracts/modules/v2/StakingPool.sol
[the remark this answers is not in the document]
shark0der: FYI: I'm not sure if this will be the final design, still pondering if it's worth or not. I think it will become clear only after I'll work on storage packing because then we'll know if we're writing to the same slot or not or maybe something else will come up.

From the user point of view (pool manager), might be worth having them in a single function so he'd be able to update both in a single tx.

PR #191, opened 2022-03-14
CowSwapOperator - Using CoW protocol for pool asset swaps

antico5:
### CowSwapOperator

- This contract is responsible for swapping the pool's assets by using CoW protocol.
- Only one order can be placed at any time.
- The swaps are achieved by pulling funds from the pool, wrapping eth if needed, and setting a presignature for a given order.
- The swapped assets and quantities are validated against parameters set by governance (i.e. min and max amounts).
- The price for the swaps is checked against the PriceFeedOracle, which uses chainlink aggregators.
- The assets remain on the operator contract until a request is made to close the order.
  - To account for the missing assets in the pool valuation, a placeholder `swapValue`, denominated in eth, is set on the pool while the swap is ongoing.
- The order can be closed, returning the assets to the pool. This can happen regardless of the order being fulfilled or not.

### Pool

- `swapValue` variable introduced. It's used to account for missing assets when they are held by CowSwapOperator waiting to be swapped
- `swapValue` is set by operator when placing an order, and reset to 0 when closing the order and returning the assets to the pool
- The function `getPoolValueInEth()` adds `swapValue` to the calculation
- The function `getAssetSwapDetails()` was changed to return the SwapDetails instead of a tuple.

### PriceFeedOracle

- Constructor reworked to take an arbitrary list of assets, their Chainlink price aggregator and decimals.
- Added `getEthForAsset()` function that returns amount of eth for a given asset and amount.

## Deployment

CowSwapOperator can be deployed using existing create2-factory script. Must set the required env. vars before running;

```
HARDHAT_NETWORK=$DEPLOY_NETWORK node scripts/create2-deploy.js \
  -a $EXPECTED_ADDRESS \
  -f $DEPLOYER_ADDRESS \
  -c [\"0x9008D19f58AAbD9eD0D60971565AA8510560ab41\",\"$SWAP_CONTROLLER_ADDRESS\",\"$MASTER_ADDRESS\",\"$WETH_ADDRESS\"] \
  -s $DEPLOY_SALT \
  -b $DEPLOY_BASE_FEE \
  -p $DEPLOY_PRIORITY_FEE \
  -l 3000000 \
  CowSwapOperator
```

After deployment, the pool's swapOperator address must be set via governance. CowSwapOperator only works with the new versions of Pool and PriceFeedOracle contracts.

## Operation

- The swaps are performed using an interactive CLI tool on the nexus-workers repo, under cow-swap-controller package.
- CLI specific instructions are on the readme file on said package.

## Test coverage

- All contract code written is 100% unit tested using tools and conventions from the project
- CLI was heavily tested on rinkeby integration, both regular scenarios and edge cases.


contracts/modules/capital/CowSwapOperator.sol
```diff
@@ -0,0 +1,256 @@
...
+contract CowSwapOperator {
+  // Storage
+  ICowSettlement public immutable cowSettlement;
+  address public immutable cowVaultRelayer;
+  INXMMaster public immutable master;
+  address public immutable swapController;
+  IWeth public immutable weth;
+  IPriceFeedOracle public immutable priceFeedOracle;
+  bytes public currentOrderUID;
+  bytes32 public immutable domainSeparator;
```
shark0der, line 22: Immutable variables do use the storage, they're replaced in the contract bytecode directly (similar to constants). Move them to a separate "section" (storage, immutables, constants, events, etc)

contracts/modules/capital/CowSwapOperator.sol
```diff
@@ -0,0 +1,240 @@
...
+      "SwapOp: validTo must be at least 10 minutes in the future"
+    );
+    require(
+      order.validTo <= block.timestamp + MAX_VALID_TO_PERIOD,
+      "SwapOp: validTo must be at most 60 minutes in the future"
+    );
+  }
+
+  function approveVaultRelayer(IERC20 token, uint256 amount) internal {
+    token.approve(cowVaultRelayer, amount); // infinite approval
```
shark0der, line 226: 1. The comment is misleading
2. The return value of the approve function is not checked and depending on the token it may be missing. Use OZ's safeApprove from SafeERC20 because it checks the optionally returned boolean. Beware that a side effect of the safeApprove method is that it checks the allowance to be zero and reverts otherwise (unwanted side effect but I think it's fine for our scenario because we reset the allowance anyway).

contracts/modules/capital/CowSwapOperator.sol
```diff
@@ -0,0 +1,240 @@
...
+    address _master,
+    address _weth,
+    address _priceFeedOracle
+  ) {
+    cowSettlement = ICowSettlement(_cowSettlement);
+    cowVaultRelayer = cowSettlement.vaultRelayer();
+    master = INXMMaster(_master);
+    swapController = _swapController;
+    weth = IWeth(_weth);
+    priceFeedOracle = IPriceFeedOracle(_priceFeedOracle);
```
shark0der, line 51: Price feed oracle is not immutable in the Pool contract and may be updated. Instead of having it immutable or storage - read it from the pool contract each time a new order is placed.

contracts/modules/capital/CowSwapOperator.sol
```diff
@@ -0,0 +1,240 @@
...
+
+    // Check how much of the order was filled, and if it was fully filled
+    uint256 filledAmount = cowSettlement.filledAmount(currentOrderUID);
+    bool fullyFilled = filledAmount == order.sellAmount;
+
+    // Cancel signature and unapprove tokens
+    if (!fullyFilled) {
+      cowSettlement.setPreSignature(currentOrderUID, false);
+      approveVaultRelayer(order.sellToken, 0);
+    }
```
shark0der, line 172: Should we get rid of the check and perform these actions anyway? I'm thinking that in an extreme scenario, if the cowSettlement contract is lying (for example if it was compromised) we would want to remove the signature and reset the approval anyway - to be safe. Any counter-arguments?

contracts/modules/capital/CowSwapOperator.sol
```diff
@@ -0,0 +1,240 @@
...
+
+    if (balance == 0) {
+      return;
+    }
+
+    if (address(asset) == address(weth)) {
+      weth.withdraw(balance); // Unwrap WETH
+      payable(address(_pool())).transfer(balance); // Transfer ETH to pool
+    } else {
+      asset.transfer(address(_pool()), balance); // Transfer ERC20 to pool
```
shark0der, line 199: Use SafeERC20's safeTransfer

contracts/modules/capital/CowSwapOperator.sol
```diff
@@ -0,0 +1,240 @@
...
+  function returnAssetToPool(IERC20 asset) internal {
+    uint256 balance = asset.balanceOf(address(this));
+
+    if (balance == 0) {
+      return;
+    }
+
+    if (address(asset) == address(weth)) {
+      weth.withdraw(balance); // Unwrap WETH
+      payable(address(_pool())).transfer(balance); // Transfer ETH to pool
```
shark0der, line 197: Do not use transfer() for ETH, use the low-level call instead and check the return value.
https://solidity-by-example.org/sending-ether/

package.json
```diff
@@ -26,6 +26,7 @@
     "@openzeppelin/test-helpers": "^0.5.15",
     "@uniswap/v2-core": "^1.0.1",
     "@uniswap/v2-periphery": "^1.1.0-beta.0",
+    "axios": "^0.25.0",
```
shark0der, line 29: Remove axios and use note-fetch instead

shark0der (changes requested):

contracts/modules/capital/CowSwapOperator.sol
```diff
@@ -0,0 +1,240 @@
...
+    require(currentOrderUID.length == 0, "SwapOp: an order is already in place");
+
+    // Order UID verification
+    validateUID(order, orderUID);
+
+    // Validate basic CoW params
+    validateBasicCowParams(order);
+
+    // Validate feeAmount is not too high
+    require(order.sellAmount / order.feeAmount >= MIN_SELL_AMT_TO_FEE_RATIO, "SwapOp: Fee is above 1% of sellAmount");
```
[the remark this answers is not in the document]
antico5: In the past I offered both options (percentage and absolute) to Hugh and he thought the 1% was fine. But nothing stops us from also adding an absolute amount. Althought we would have to use the oracle and increase the gas cost a bit. Up to you. About the revert message, you're right I'll change it.

contracts/modules/capital/CowSwapOperator.sol
```diff
@@ -0,0 +1,240 @@
...
+    uint256 totalOutAmount = order.sellAmount + order.feeAmount;
+
+    if (isSellingEth(order)) {
+      // Validate min/max setup for buyToken
+      IPool.SwapDetails memory swapDetails = pool.getAssetSwapDetails(address(order.buyToken));
+      require(swapDetails.minAmount != 0 || swapDetails.maxAmount != 0, "SwapOp: buyToken is not enabled");
+      uint256 buyTokenBalance = order.buyToken.balanceOf(address(pool));
+      require(buyTokenBalance < swapDetails.minAmount, "SwapOp: can only buy asset when < minAmount");
+      require(buyTokenBalance + order.buyAmount <= swapDetails.maxAmount, "SwapOp: swap brings buyToken above max");
+      require(buyTokenBalance + order.buyAmount >= swapDetails.minAmount, "SwapOp: swap leaves buyToken below min");
```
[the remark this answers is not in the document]
antico5: I just checked the logic from the old SwapOperator and you are right, when exchanging eth for an asset it's not required that the final amount is > min. I think my rationale for this was to avoid the possibility of a compromised swap controller making a lot of small trades to burn fees (since I removed the swap frequency validation). Your argument is valid so I think we should remove that particular restriction. Do you think we can ignore the potential scenario of a bad swap controller executing a lot of small swaps? or should I add the frequency check to prevent it ?

contracts/modules/capital/CowSwapOperator.sol
```diff
@@ -0,0 +1,240 @@
...
+
+      // Validate minimum pool eth reserve
+      require(address(pool).balance - totalOutAmount >= pool.minPoolEth(), "SwapOp: Pool eth balance below min");
```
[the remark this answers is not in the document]
antico5: Correct, since at this point a) ether has not been transfered from the pool yet and b) swapValue was not set yet. And at every closeOrder() call, swapValue is set to 0.

contracts/modules/capital/CowSwapOperator.sol
antico5: This `if` statement is for saving a moderate amount of gas, since when the order is fully filled, there's no need to execute the 2 actions.

contracts/modules/capital/CowSwapOperator.sol
antico5: I just checked the safeApprove function and it seems it's deprecated in favor of safeIncreaseAllowance (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol#L45), althought the behavior is not the same. I favor the latter as well since if for any reason the allowance is not 0, the contract will be usable, What do you think?

contracts/modules/capital/CowSwapOperator.sol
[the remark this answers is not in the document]
horodnicdragos: I'd suggest using errors instead of strings for revert reasons as they are cheaper and allow the message to be customized with the actual values:

https://docs.soliditylang.org/en/v0.8.9/structure-of-a-contract.html#errors

package.json
antico5: This dependency and everything related to scripts will be removed from the code, since now it's on the CLI codebase (using node-fetch there).

contracts/modules/capital/CowSwapOperator.sol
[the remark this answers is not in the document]
shark0der: As per our discussion, let's implement a trade frequency check.

contracts/modules/capital/CowSwapOperator.sol
shark0der: Perhaps the security aspect is more important than the gas savings we'd gain here.

test/unit/CowOperator/closeOrder.js
```diff
@@ -0,0 +1,368 @@
...
+      expect(event.args.filledAmount).to.eq(order.sellAmount.div(2));
+    });
+
+    it('when order was fully filled', async function () {
+      await cowSettlement.fill(contractOrder, orderUID, order.sellAmount, order.feeAmount, order.buyAmount);
+
+      const tx = await swapOperator.closeOrder(contractOrder);
+      const rcp = await tx.wait();
+
+      const event = rcp.events.find(e => e.event === 'OrderClosed');
```
horodnicdragos, line 352: Hi! I'm curious whether you were aware of this or if you've encountered any inconveniences using it:

https://ethereum-waffle.readthedocs.io/en/latest/matchers.html?highlight=events#emitting-events

antico5: Hi! Yes, I'm aware of that matcher but was not able to properly match the arguments of the event. So the closest I got using that was: `await expect(swapOperator.closeOrder(contractOrder)).to.emit(swapOperator, 'OrderClosed').withArgs(Object.values(contractOrder), 0);` , but it doesn't work because the ordering of the order attributes is different. If I had a more powerful argument matcher I'd do `withArgs(arrayMatching(Object.values(contractOrder)), 0)`, but afaik we don't have that available (please correct me if I'm wrong). So instead of manually constructing the expectation (tuple with 12 elements), I first got the receipt and used an available matcher: `expect(event.args.order).to.deep.include.members(Object.values(contractOrder));`

horodnicdragos: I see. Curently node seems to sort the values in the order they are declared in the object so simply moving the receiver key-value pair right after buyToken in order would make the test pass using that matcher, but as far as I know the order of Array.values which is the same as for...in should not be relied upon as it's implementation specific. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#array_iteration_and_for...in)

My initial nitpicky concern towards using chai's include was that two keys could theoretically end up having the same value. It is however highly unlikely since the order comes from calldata and nothing alters it. 

But in other words, if someone were to trick the test on purpose it would be possible by modifying one of the properties in the Data struct from the emitted event such as feeAmount, making it equal to sellAmount for example, and the test would pass even though the order object is not the same as the one provided as a parameter.

If you think it's worth it you just need to get the values in the correct order by using a function that maps the ordered Data struct keys to the values in the object.

antico5: I see your point. Well, to be 100% safe I just updated the test to reflect your suggestion, now I build the order tuple with a function and use it for an exact match using `withArgs` matcher.

horodnicdragos (approved):

shark0der (approved):

patitonar (approved):

PR #193, opened 2022-03-22
Feature/fork deploy and migration

horodnicdragos:


contracts/mocks/Disposables/DisposableCover.sol
```diff
@@ -35,15 +35,25 @@ contract DisposableCover is MasterAwareV2 {
...
 
-  function addProducts(Product[] calldata newProducts) public {
+  function addProducts(
+    Product[] calldata newProducts,
+    string[] calldata ipfsMetadata
+  ) external {
+    uint initialProuctsCount = products.length;
     for (uint i = 0; i < newProducts.length; i++) {
       products.push(newProducts[i]);
+      emit ProductUpserted(initialProuctsCount + i, ipfsMetadata[i]);
```
danoctavian, line 45: spelling error here: initialProuctsCount -> initialProductsCount

danoctavian: why is it upserted ?

You upsert if you may also update an existing record, but here it just pushes one at the end, so it would be just ProductAdded

contracts/mocks/Disposables/DisposableCover.sol
```diff
@@ -35,15 +35,25 @@ contract DisposableCover is MasterAwareV2 {
...
 
-  function addProductTypes(ProductType[] calldata newProductTypes) public {
+  function addProductTypes(
+    ProductType[] calldata newProductTypes,
+    string[] calldata ipfsMetadata
+  ) public {
+    uint initialProuctTypesCount = productTypes.length;
     for (uint i = 0; i < newProductTypes.length; i++) {
       productTypes.push(newProductTypes[i]);
+      emit ProductTypeUpserted(initialProuctTypesCount + i, ipfsMetadata[i]);
```
danoctavian, line 56: same question as above with regards to `Upserted`

contracts/interfaces/IStakingPool.sol
```diff
@@ -21,7 +21,7 @@ struct ProductParams {
 
 struct ProductInitializationParams {
   uint productId;
-  uint weight;
+  uint8 weight;
```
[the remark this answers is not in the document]
horodnicdragos: It should be 0-100, each unit representing one percent. @shark0der let us know if anything is about to change here.

contracts/modules/capital/Pool.sol
```diff
@@ -59,95 +83,121 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
+    // Add investment assets
+    investmentAssets.push(Asset(stETHAddress, 18));
 
-      assets.push(asset);
-      swapDetails[asset.assetAddress].minAmount = _minAmounts[i];
-      swapDetails[asset.assetAddress].maxAmount = _maxAmounts[i];
-      swapDetails[asset.assetAddress].maxSlippageRatio = _maxSlippageRatios[i];
-    }
+    // Set DAI swap details
+    swapDetails[DAIAddress] = SwapDetails(
```
danoctavian, line 108: these used to be parameters.

This is definitely nice for clarity of what's being used, but it might make some tests cumbersome if we want to the test the Pool with different values.

contracts/modules/capital/Pool.sol
```diff
@@ -361,11 +490,11 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
     emit NXMBought(msg.sender, ethIn, tokensOut);
   }
 
-  /**
-   * @dev Sell NXM tokens and receive ETH.
-   * @param tokenAmount Amount of tokens to sell.
-   * @param  minEthOut Minimum amount of ETH to be received. Revert if ethOut falls below this number.
-   */
+  /// Sell NXM tokens and receive ETH.
```
danoctavian, line 493: why change the comment style on top of functions?

the /** */ is pretty standard across multiple languages and looks better/more familiar for most devs.

contracts/mocks/Disposables/DisposableCover.sol
horodnicdragos: The event is emitted whenever a product changes. This means both adding a new product and editing an existing one. It's purpose is to follow changes to any specific wordings similar to what we currently have in the UI as "excluded scenarios" which vary from one product to another. This information is stored on IPFS and the event is a reference to the cid referencing the file that contains it.

contracts/modules/capital/Pool.sol
horodnicdragos: We have setSwapDetails which can easily be used to test different values.

contracts/modules/capital/Pool.sol
horodnicdragos: Here's why:

1. You can easily copy or move chunks of the comment without having to worry about turning it back into a comment.
2. It doesn't make any assumptions about smart tooling that handles comments.
3. It's 2 lines smaller.
4. It's also standard across multiple languages.
5. It doesn't affect the code or other devs in any way, even if we're looking in subjective matters such as looks and familiarity. You don't have to stick to it, but it definitely makes my dev experience more pleasent.

contracts/modules/staking/PooledStaking.sol
```diff
@@ -497,8 +698,9 @@ contract PooledStaking is IPooledStaking, MasterAware {
 
       // i-- might underflow to MAX_UINT
       // but that's fine since it will be incremented back to 0 on the next loop
-      i--;
+      //unchecked{i--;}
```
danoctavian, line 701: is this still needed in this commented out form?

contracts/modules/staking/PooledStaking.sol
```diff
@@ -737,47 +957,519 @@ contract PooledStaking is IPooledStaking, MasterAware {
     }
   }
 
-  function migrateToNewV2Pool(address stakerAddress) external {
+  function getV1ProceForProduct(uint id) pure internal returns (uint) {
```
danoctavian, line 960: misspelling here i'm assuming 

getV1PercForProduct
getV1PercentageForProduct
getV1PriceForProduct

contracts/modules/staking/PooledStaking.sol
```diff
@@ -737,47 +957,519 @@ contract PooledStaking is IPooledStaking, MasterAware {
...
+      id == 108 ||
+      // Ribbon Finance v2
+      id == 109
+    ) {
+      return 2600000000000000000; // 2.6%
+    }
+
+    // DDEX
+    if (id == 6) {
+      return 56398373648265250000; // 56.39837364826525%
```
danoctavian, line 1090: are we going to keep updating these to match the latest number?

contracts/modules/staking/PooledStaking.sol
```diff
@@ -737,47 +957,519 @@ contract PooledStaking is IPooledStaking, MasterAware {
...
+    // Trader Joe
+    if (id == 111) {
+      return 7995085516794736000; // 7.9950855167947354%
+    }
+    // {V1_PRICES_HELPER_END}
+
+    revert("Invalid product id");
+  }
+
+  function migrateToNewV2Pool(address stakerAddress, uint groupId) external noPendingActions {
```
danoctavian, line 1382: we should measure gas here to see if there's any issue with max block size and whether we need iterations.

contracts/modules/staking/PooledStaking.sol
```diff
@@ -737,47 +957,519 @@ contract PooledStaking is IPooledStaking, MasterAware {
...
+      uint stakePositionNFTId;
+
+      {
+        // [todo] This might require an offset to start the first group at the time of v2 deploy
+        uint GROUP_SIZE = 91 days;
+        uint firstActiveGroupId = block.timestamp / GROUP_SIZE;
+
+        // Use the groupId provided as a paramtere if the user is migrating to v2 himself
+        // Use next id after the first active gorup id for those in the initial migration list
+        uint groupIdInEffect = stakerAddress == msg.sender ? groupId : firstActiveGroupId + 1;
```
danoctavian, line 1454: so for addresses that are not in the initial migration list, the groupId is self-defined (parameter).

Is stakingPool.deposit supposed to revert if that groupId is wrong (not in the right range)?

contracts/modules/capital/Pool.sol
danoctavian: Understood.

I would want things to be consistent so all comments in 1 way, i prefer the old style. I'll let the others chime in on this one.

contracts/modules/staking/PooledStaking.sol
```diff
@@ -669,14 +871,14 @@ contract PooledStaking is IPooledStaking, MasterAware {
...
         _contractStakers.pop();
-        i--;
         _stakerCount--;
 
         // since the stake is 0, there's no reward to give
         continue;
       }
 
       _actualRewardAmount = _actualRewardAmount + _stakerRewardAmount;
+      i++;
```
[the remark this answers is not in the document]
horodnicdragos: Try taking a look at the diff as a whole. There were incrementations and decrementations that were cancelling each other out, sometimes ending up with underflows. 

That's one of the reasons why I left some things commented (unchecked operations which would be required to allow underflows). Thus you won't be able to minimize changes but instead pick one of the two approaches. The one involving underflows seemed fragile to me. 

The test coverage should be pretty good for this part of the code.

contracts/modules/staking/PooledStaking.sol
horodnicdragos: Yes, `get-v1-cover-prices.js` does that automatically.

contracts/modules/staking/PooledStaking.sol
```diff
@@ -737,47 +957,519 @@ contract PooledStaking is IPooledStaking, MasterAware {
...
+    uint migratableCount = 0;
+
     for (uint i = 0; i < contractsCount; i++) {
       address oldProductId = stakers[stakerAddress].contracts[i];
-      uint productId = productsV1.getNewProductId(oldProductId);
+      uint productId;
+      try productsV1.getNewProductId(oldProductId) returns (uint v) {
+        productId = v;
+      } catch {
+        console.log("oldProductId %s not found in productsV1, likely deprecated", oldProductId);
```
danoctavian, line 1418: now that i see this, should we leave the console.log in the deploy to have this logging?

Or just have a custom event, which adds less to the bytecode

contracts/modules/staking/PooledStaking.sol
```diff
@@ -737,47 +957,519 @@ contract PooledStaking is IPooledStaking, MasterAware {
...
-    // stakingPool.stakeAllocation(products, weights);
+        // Move deposit to v2 pool
+        stakePositionNFTId = stakingPool.deposit(deposit, groupIdInEffect, 0);
+      }
+
+      // NFT transfer reverts if stakerAddress is a contract that doesn't implement IERC721Receiver
+      stakingPool.safeTransferFrom(address(this), stakerAddress, stakePositionNFTId);
+    }
 
+    // Finally set the v1 deosit to 0
```
danoctavian, line 1464: *deposit

contracts/modules/v2/CoverMigrator.sol
```diff
@@ -0,0 +1,36 @@
...
+import "../../interfaces/IIndividualClaims.sol";
+import "../../interfaces/IAssessment.sol";
+import "../../interfaces/IERC20Detailed.sol";
+import "../../interfaces/ICoverNFT.sol";
+
+import "../../abstract/MasterAwareV2.sol";
+
+/// Provides a way for contracts which don't use the Gateway.sol contract to migrate covers to V2
+/// using the submitClaim claim function.
+contract CoverMigrator is MasterAwareV2 {
```
danoctavian, line 18: is this replacing an existing contract?

contracts/modules/staking/PooledStaking.sol
horodnicdragos: I think it should. At the moment it looks like it only checks whether the group is in the future.

`    require(groupId >= firstActiveGroupId, "StakingPool: Requested group has expired");`

Added a todo in StakingPool.sol.

contracts/modules/v2/CoverMigrator.sol
horodnicdragos: Yes, Claims.sol. It's not explicitly mentioned in the comment above but it should be obvious. Some contracts use Claims.sol directly instead of Gateway.sol so we're leaving this CoverMigrator contract in place (CL) to allow cover migration for 1y. It can then be removed from Master entirely.

roxdanila (commented):
Adding a few more comments to the mix. Still need to dive a bit more in the staking migration

scripts/create-testnet-version-data.js
```diff
@@ -55,22 +55,22 @@ console.log({ addresses });
 fs.writeFileSync(`${deployPath}/${network}-addresses.json`, JSON.stringify(addresses, null, 2));
 
 const artifactPathOfContractCode = {
-  LCD: 'modules/claims/LegacyClaimsData.sol/LegacyClaimsData.json',
-  LCL: 'modules/claims/LegacyClaims.sol/LegacyClaims.json',
-  LCR: 'modules/claims/LegacyClaimsReward.sol/LegacyClaimsReward.json',
-  LCP: 'modules/claims/LegacyClaimProofs.sol/LegacyClaimProofs.json',
+  // LCD: 'modules/claims/LegacyClaimsData.sol/LegacyClaimsData.json',
```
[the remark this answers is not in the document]
horodnicdragos: I think we need that in the UI to display v1 claims.

contracts/modules/governance/MemberRoles.sol
```diff
@@ -186,25 +189,23 @@ contract MemberRoles is IMemberRoles, Governed, LegacyMasterAware {
 
   }
 
+  /**
+   * @dev Change the wallet address which receive Joining Fee
+   */
+  function changeJoiningFeeWallet(address payable _address) external onlyInternal {
```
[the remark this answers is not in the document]
horodnicdragos: Not sure if this was (or still is) a legal requirement. I'll remove it if it's no longer needed.

roxdanila (approved):

roxdanila:
Please fix/skip failing tests before merging: https://app.circleci.com/pipelines/github/NexusMutual/smart-contracts/1381/workflows/b4bcda3d-5352-4821-bad7-9f976a949e8f/jobs/2817

contracts/modules/staking/PooledStaking.sol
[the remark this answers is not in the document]
shark0der: The underflow was intended. The patch looks correct to me and indeed, the tests should cover this. The edge case here and in the burn function as well is when there's a single staker left and his stake is zero so we're removing the entry from the `_contractStakers` array.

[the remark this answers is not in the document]
shark0der: processBurns.js: the "should remove stakers when burning 0-deposit stakers" test for example covers this edge-case in the burn function.

PR #194, opened 2022-03-23
Staking

shark0der:


contracts/modules/v2/StakingPool.sol
```diff
@@ -408,28 +437,40 @@ contract StakingPool is IStakingPool, ERC721 {
   /* pool management */
 
   function setProductDetails(ProductParams[] memory params) external onlyManager {
-
+    // silence compiler warnings
+    params;
+    activeStake = activeStake;
+    revert("Not implemented");
```
horodnicdragos, line 443: I think it's less likely to block others when integrating these function if you use a todo comments instead of a revert. It's quite hard to miss if we forget implementing something here.

contracts/modules/staking/PooledStaking.sol
```diff
@@ -763,19 +763,22 @@ contract PooledStaking is IPooledStaking, MasterAware {
...
     // IStakingPool stakingPool = IStakingPool(cover.createStakingPool(stakerAddress));
-    revert("Not implemented");
 
     // [todo] Add the corresponding calls after a IStakingPool is available
     // uint stakingPositionNFT = stakingPool.deposit(deposit);
     // stakingPool.stakeAllocation(products, weights);
 
     stakers[stakerAddress].deposit = 0;
+
+    revert("Not implemented");
```
horodnicdragos, line 773: Leave this the way it was before. It will conflict with unmerged changes.

PR #196, opened 2022-03-23
burn function

danoctavian:


contracts/modules/cover/Cover.sol
```diff
@@ -437,15 +439,37 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
 
     CoverData storage cover = _coverData[coverId];
-    cover.amountPaidOut += SafeUintCast.toUint96(amount);
+    cover.amountPaidOut += SafeUintCast.toUint96(burnAmount);
+
+    CoverSegment memory segment = coverSegments(coverId, segmentId);
+    PoolAllocation[] storage allocations = coverSegmentAllocations[coverId][segmentId];
+
+    // burn an
+    uint burnRatio = burnAmount * BURN_DENOMINATOR / segment.amount;
```
horodnicdragos, line 456: denominator - mathematics : the part of a fraction that is below the line and that functions as the divisor of the numerator

BURN_DENOMINATOR is not a denominator but a precision factor. I've simply named the constant PRECISION whenever I had divisions between NXM amounts.

contracts/modules/cover/Cover.sol
```diff
@@ -437,15 +439,37 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
+
+    for (uint i = 0; i < allocations.length; i++) {
```
horodnicdragos, line 458: Cache the length of the array. This would hit the storage at every iteration.

contracts/modules/cover/Cover.sol
```diff
@@ -437,15 +439,37 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
+
+      PoolAllocation storage allocation = allocations[i];
+      IStakingPool _stakingPool = stakingPool(allocation.poolId);
+
+      uint nxmBurned = allocation.coverAmountInNXM * burnRatio / BURN_DENOMINATOR;
```
horodnicdragos, line 463: BURN_DENOMINATOR makes sense here.

contracts/modules/cover/Cover.sol
danoctavian: solved by removing that variable entirely

PR #197, opened 2022-04-06
Separate pool assets

horodnicdragos:
Pool assets are now split in two categories:
- Payout assets (used by covers when claims needs to be paid out)
- Investment assets (used solely for investments)

Payout assets cannot be removed once they are added as cover data
references them by indexes. They can however be ignored if calculating
the total ETH worth of all assets becomes expensive making the iteration
skip the indexes that are no longer used by covers.

Investment assets on the other hand can be added and removed freely.

This commit also updates Master, MCR, VotePower, Proxy,
UpgradeabilityProxy, OwnedUpgradeabilityProxy, and any related
mocks or dispoable contracts to solidity 0.8.9.

Tests and coverage will be fixed in a future PR.


contracts/interfaces/IPool.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-pragma solidity >=0.5.0;
+pragma solidity >=0.8.9;
```
shark0der, line 3: Use the minimum version in interfaces (`>=0.8.0`) so they're usable by other contracts/projects with a lower patch version (ex. `0.8.4`).

contracts/modules/capital/Pool.sol
```diff
@@ -58,95 +76,107 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
-    require(assetAddresses.length == _minAmounts.length, "Pool: Length mismatch");
-    require(assetAddresses.length == _maxAmounts.length, "Pool: Length mismatch");
-    require(assetAddresses.length == _maxSlippageRatios.length, "Pool: Length mismatch");
-    require(assetAddresses.length == assetDecimals.length, "Pool: Length mismatch");
-
-    for (uint i = 0; i < assetAddresses.length; i++) {
+    // The order of payoutAssets should never change between updates. Do not remove the following
+    // lines!
+    payoutAssets.push(Asset(ETH, 18));
+    payoutAssets.push(Asset(DAIAddress, 18));
```
shark0der, line 92: This might be error prone as it will be checked manually. Would be better to have a mechanism that will read the values from the previous pool, similar to how we copy over MCR data. Probably doesn't make sense to implement that now because the interface changes, leave a todo note?

contracts/modules/capital/Pool.sol
```diff
@@ -166,44 +196,93 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
+    uint payoutAssetsCount = payoutAssets.length;
+    for (uint i = 0; i < payoutAssetsCount; i++) {
+      require(assetAddress != payoutAssets[i].assetAddress, "Pool: Asset exists");
+    }
+
+    // Check whether the new asset already exists as an investment asset
+    uint investmentAssetsCount = investmentAssets.length;
+    for (uint i = 0; i < investmentAssetsCount; i++) {
+      require(assetAddress != investmentAssets[i].assetAddress, "Pool: Asset exists");
     }
```
shark0der, line 216: This function prevents us from adding a payout asset if it is already used as an investment asset. For example if we'd (hypothetically) want to add stETH to be used as a payout asset - we'd be unable to do so.

horodnicdragos: You can if you remove it from the investmentAssets.

contracts/modules/capital/Pool.sol
```diff
@@ -166,44 +196,93 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
+      require(assetId < payoutAssets.length, "Pool: Payout asset does not exist");
+      require(ignoredPayoutAssetsBitmap & (1 << assetId) != 0, "Pool: Payout asset is ignored");
+
+      // Remove swap details
+      address assetAddress = payoutAssets[assetId].assetAddress;
+      delete swapDetails[assetAddress];
+
+      // Ignore asset which makes getPoolValueInEth skip it when the function loops through
+      // payments assets
+      ignoredPayoutAssetsBitmap |= SafeUintCast.toUint32(1 << assetId);
```
shark0der, line 257: This design has a terrible limitation that it will indeed prevent cover buys in the said asset but will also immediately exclude them from the pool value calculation.

contracts/modules/capital/Pool.sol
```diff
@@ -19,7 +21,8 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
   uint16 constant MAX_SLIPPAGE_DENOMINATOR = 10000;
 
   /* storage */
-  Asset[] public override assets;
+  Asset[] public override payoutAssets;
```
shark0der, line 24: This array is used for two purposes:
- Detect if the user can pay the cover premium using this asset
- Use this asset in a payout
The current naming only suggest the later and fails to signal the first one, for this reason I suggest using the `coverAssets` name instead.

contracts/modules/capital/Pool.sol
```diff
@@ -34,6 +37,21 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
+  // Examples:
+  // 1 (10) = 00000000000000000000000000000001 (2)
+  //                                         ^
+  //                                         payoutAssets[0] is ignored
+  //
+  // 9 (10) = 00000000000000000000000000001001 (2)
+  //                                      ^  ^
+  //                                      payoutAssets[0] and payoutAssets[3] are both ignored
+  //
+  uint32 public ignoredPayoutAssetsBitmap;
```
shark0der, line 53: It seems to me that this flag is only used to mark the "payout" assets as being deprecated and got me initially confused to think that it means that the asset is present in both arrays and should be skipped when iterating over one of them. If this flag's only purpose is deprecation I'd suggest calling it `deprecatedPayoutAssetsBitmap` or `deprecatedCoverAssetsBitmap` if we proceed with the payout -> cover rename.

contracts/modules/capital/Pool.sol
horodnicdragos: It's as "terrible" as the current mainnet implementation, where the same thing happens when you remove an asset. That's intended for both cover buys and pool value calculation. I don't see any actionables nor constructive arguments in this comment. Please clarify if necessary!

shark0der: Let's say tomorrow we want to prevent cover buys with DAI (for whatever reason). Currently we have over $2m DAI in pool, this will instantly impact the pool value. We need to be able to prevent cover buys while still using the asset in the pool value calculation. An additional issue would be that we still have to be able to be able to send payouts in that asset for 1 year + grace period. After this period we should be able to sell the asset from the pool and only then remove it.

It is indeed true that this issue is present in the current mainnet implementation, seems like we didn't notice until now, however that's not a reason to keep it that way.

horodnicdragos: 2 things here:

1. As stated in the natspec, the flag doesn't prevent cover buys. That's done separately in Cover.sol (coverAssetsFallback and coverAssets in the Product struct).
2. Added a check for balance requiring it to be 0 before removing the token. If the balanceOf function reverts it is considered 0.

contracts/modules/capital/MCR.sol
```diff
@@ -96,13 +92,13 @@ contract MCR is IMCR, MasterAware {
     // IPriceFeedOracle priceFeed = pool.priceFeedOracle();
     // address daiAddress = priceFeed.daiAddress();
 
-    // uint ethAmount = qd.getTotalSumAssured("ETH").mul(1e18);
-    // uint daiAmount = qd.getTotalSumAssured("DAI").mul(1e18);
+    // uint ethAmount = qd.getTotalSumAssured("ETH") * 1e18;
```
danoctavian, line 95: why do we have this commented out instead of removed?

horodnicdragos: Removed ✅

test/fork/migration-v2.js
```diff
@@ -109,7 +110,7 @@ describe('v2 migration', function () {
     this.claimsData = await factory('CD');
   });
 
-  it('impersonate AB members', async function () {
+  it('impersonate AB members', async function() {
```
danoctavian, line 113: across the js codebase we have a space between the function keyword or the function name and the parenthesis.

If we make this change we need to make it everywhere, in all the files (iteratively is fine). want to make sure everyone is on board with this @roxdanila @shark0der

danoctavian: note: this is done automatically from prettier

roxdanila: I'd prefer not to make this change at this point.

test/unit/Pool/index.js
```diff
@@ -1,7 +1,7 @@
 const { takeSnapshot, revertToSnapshot } = require('../utils').evm;
 const setup = require('./setup');
 
-describe('Pool unit tests', function () {
+describe.only('Pool unit tests', function () {
```
roxdanila, line 4: remove only

contracts/modules/governance/MemberRoles.sol
```diff
@@ -109,7 +96,7 @@ contract MemberRoles is IMemberRoles, Governed, LegacyMasterAware {
    * @dev to set address of kyc authentication
    * @param _add is the new address
    */
-  function setKycAuthAddress(address _add) external onlyInternal {
```
roxdanila, line 112: why was this onlyInternal before?

roxdanila, line 112: why was this onlyInternal before?

contracts/modules/capital/Pool.sol
```diff
@@ -3,23 +3,29 @@
...
+
+import "../../utils/SafeUintCast.sol";
 import "../../abstract/MasterAware.sol";
 import "../../interfaces/IMCR.sol";
 import "../../interfaces/INXMToken.sol";
 import "../../interfaces/IPool.sol";
 import "../../interfaces/IPriceFeedOracle.sol";
 import "../../interfaces/IQuotation.sol";
 import "../../interfaces/ITokenController.sol";
+import "hardhat/console.sol";
```
roxdanila, line 18: remove

roxdanila, line 18: remove

contracts/modules/governance/MemberRoles.sol
```diff
@@ -285,7 +272,6 @@ contract MemberRoles is IMemberRoles, Governed, LegacyMasterAware {
     cover.transferCovers(msg.sender, newAddress, coverIds);
 
     stakingPools;
-    revert("Staking pool migration not implemented");
```
roxdanila, line 288: add a todo at least?

roxdanila, line 288: add a todo at least?

test/fork/migration-v2.js
roxdanila: @horodnicdragos please undo the "space" change on this file

roxdanila: @horodnicdragos please undo the "space" change on this file

roxdanila (changes requested):

PR #199, opened 2022-04-13
Fix/reduce cover size - cover at 20k

danoctavian:


contracts/modules/cover/StakingPoolCreator.sol
```diff
@@ -0,0 +1,63 @@
...
+    address manager,
+    ProductInitializationParams[] calldata params
+  ) external returns (address stakingPoolAddress) {
+
+    stakingPoolAddress = address(
+      new MinimalBeaconProxy{ salt: bytes32(uint(stakingPoolCounter)) }(coverAddress)
+    );
+    IStakingPool(stakingPoolAddress).initialize(manager, params);
+
+    stakingPoolCounter++;
```
horodnicdragos, line 50: Would this counter better suit us if it was stored in the Cover proxy? It's unlikely that we'd upgrade this contract that often but it now implies that we need to be careful about copying the previous value to continue from where it left. Making this a library (delegate call) has the same gas cost as a contract call but allows you to store the counter in the proxy.

contracts/modules/cover/Cover.sol
```diff
@@ -575,28 +526,8 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
 
   /* ========== Staking Pool creation ========== */
 
-  function createStakingPool(
```
horodnicdragos, line 578: Moving this function entirely from Cover.sol means that PooledStaking needs to call the pool creator contract instead. Since the pool creator contract is not internal and stored as immutable it means that if there's the need to ever change something, we also need to redeploy Cover.sol and PooledStaking (at least until we decomission it ~3-4 months).

horodnicdragos (changes requested):

contracts/modules/cover/Cover.sol
shark0der: We still need to proxy through Cover.sol because of the pool counter which will be stored in the Cover contract.

danoctavian: yes

horodnicdragos (approved):

contracts/modules/cover/Cover.sol
```diff
@@ -92,22 +96,17 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
-      abi.encodePacked(
-        type(MinimalBeaconProxy).creationCode,
-        abi.encode(coverProxy)
-      )
-    );
-    stakingPoolImplementation =  _stakingPoolImplementation;
     coverNFT = _coverNFT;
+
+    stakingPoolProxyCodeHash = CoverUtilsLib.calculateProxyCodeHash();
+    stakingPoolImplementation = _stakingPoolImplementation;
```
shark0der, line 109: Use `_stakingPoolImplementation.codehash` instead of `CoverUtilsLib.calculateProxyCodeHash()`

shark0der: LE: not applicable here.

PR #203, opened 2022-04-29
Feature/expiry based active cover amount

danoctavian:


contracts/modules/cover/Cover.sol
```diff
@@ -588,6 +556,35 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
+
+    require(coverAmountTrackingEnabled, "Cover expiring not enabled");
+
+    CoverData memory cover = _coverData[coverId];
+    require(!cover.expired, "Cover: Cover is already expired.");
+
+    uint lastCoverSegmentIndex = _coverSegments[coverId].length - 1;
+    CoverSegment memory lastCoverSegment = _coverSegments[coverId][lastCoverSegmentIndex];
+    if (lastCoverSegment.period + lastCoverSegment.start > block.timestamp) {
+      cover.expired = true;
```
horodnicdragos, line 577: Wouldn't it be better to have the expired property on the segment, more precisely on the last segment? That makes it possible for covered vaults to reuse the same NFT even if they decide to not cover it for a certain period of time. It makes things easier to track and implement in UIs and dashboards, and also makes getting cover again on the same product + payout asset pair cheaper for all users since they don't have to mint a new NFT.

contracts/modules/cover/Cover.sol
```diff
@@ -588,6 +556,35 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
+
+
+      if (coverAmountTrackingEnabled) {
+        totalActiveCoverForAsset[cover.payoutAsset] -= lastCoverSegment.amount;
```
horodnicdragos, line 581: Careful! 🚨🚨🚨

You haven't used the public `coverSegments` view which accounts for the amount that's already paid out. You can potentially subtract way more than you should in the following scenario:

Buy 200 ETH cover, claim 100 ETH, `amountPaidOut` becomes 100, edit cover and increase it back to 200 ETH which internally sets the amount to 300 ETH, however only 200 ETH is covered (300 ETH "virtual" amount - 100 ETH from `amountPaidOut`.

Make sure that you use the view directly wherever it makes sense. Public view are supposed to be optimized by the compiler and there's no gas benefit by accessing the internal state instead. Only use the internal state when you need to work with the "virtual" amount, such as when you edit a cover.

horodnicdragos (changes requested):
⚙️

contracts/modules/cover/Cover.sol
danoctavian: i thought it would be strange because it's a redundant field in all but the last segment.

In the cover it makes more logical sense

horodnicdragos (approved):
LGTM

PR #205, opened 2022-05-09
Feature: Improved sign up

horodnicdragos:
Allow paying the joining fee after KYC approval

    This deprecates refundEligible and removes the need to set and unset it
    thus making the payJoiningFee function cheaper and denying KYC no longer
    needs an onchain transaction, it would simply not call the signUpMember
    function.

    This commit also removes the onlyOwner modifier and the functions that
    were using it. All of these functions were no longer used on mainnet and
    neither they were tested or used in the script that deploys the whole
    system from zero.


roxdanila (changes requested):

contracts/modules/governance/MemberRoles.sol
```diff
@@ -33,14 +38,14 @@ contract MemberRoles is IMemberRoles, Governed, LegacyMasterAware {
...
   uint public launchedOn;
 
-  // This slot was previously used as a mapping (address => address payable)
-  mapping (address => address payable) public _unused2;
-  mapping(address => bool) public refundEligible;
+  mapping(address => address payable) public _unused3;
+  mapping(address => bool) public _unused4;
+  mapping(bytes32 => bool) public usedMessageHashes;
 
   modifier checkRoleAuthority(uint _memberRoleId) {
```
roxdanila, line 50: looks like modifier can go away as well

contracts/mocks/Disposables/DisposableMemberRoles.sol
```diff
@@ -43,10 +41,39 @@ contract DisposableMemberRoles is MemberRoles {
...
-    _unused2[0x181Aea6936B407514ebFC0754A37704eB8d98F91] = payable(0x1337DEF18C680aF1f9f45cBcab6309562975b1dD);
+    _unused3[0x181Aea6936B407514ebFC0754A37704eB8d98F91] = payable(0x1337DEF18C680aF1f9f45cBcab6309562975b1dD);
+  }
+
+  /**
+   * @dev to add initial member roles
+   * @param _firstAB is the member address to be added
+   * @param memberAuthority is the member authority(role) to be added for
+   */
+  function _addInitialMemberRoles(address _firstAB, address memberAuthority) internal {
```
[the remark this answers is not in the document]
horodnicdragos: It's a disposable contract and that particula function is required for initializing the member roles when deploying on testnets and setting up integration tests environment. It was move from the MemberRoles where we already have the state initialized on mainnet.

contracts/modules/governance/MemberRoles.sol
```diff
@@ -129,125 +109,125 @@ contract MemberRoles is IMemberRoles, Governed, LegacyMasterAware {
...
+    require(
+      msg.value == joiningFee,
+      "MemberRoles: The transaction value should equal to the joining fee"
+    );
 
-  }
+    // Extract the nonce and the compact signature representation.
+    // See: https://eips.ethereum.org/EIPS/eip-2098
+    (uint nonce) = abi.decode(data[:32], (uint));
+    bytes memory signature = data[32:96];
```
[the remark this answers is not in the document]
horodnicdragos: I could do that but if we'll ever want to remove them we'll have the two types in the function parameters forever to avoid breaking changes. The intention here is to avoid bricking any integrations relying on the `signUp` function that would be caused by changes in the function selector.

A similar pattern which gives us this flexibility already exists in Gateway.sol in functions such as `getCoverPrice`, `buyCover` etc. This means that if we reserve this last parameter to be an arbitrary array of byters, we can change the behavior of the function without changing the signature. There would also be no need for an additional gateway contract which essentially does exactly that.

contracts/modules/governance/MemberRoles.sol
horodnicdragos: Why?

roxdanila: discussed offline, this is a broader discussion and out of scope for this PR

roxdanila (approved):
Fix conflicts, and good to go

PR #215, opened 2022-05-16
Keep only the v2 internal contracts

horodnicdragos:


contracts/interfaces/IMasterAwareV2.sol
```diff
@@ -4,8 +4,17 @@ pragma solidity >=0.5.0;
 
 interface IMasterAwareV2 {
 
-  // [todo] Are there any missing contracts here?
-  enum ID {GW, GV, MR, CL, CR, MC, P1, QT, QD, PC, PS, TC, TD, IC, AS, CO}
+  enum ID {
```
roxdanila, line 7: ProposalCategory.sol missing

roxdanila (changes requested):

contracts/interfaces/IMasterAwareV2.sol
horodnicdragos: I actually removed all contracts which were not used. If we ever decommission any of them not having things shifting around is better. If we ever need them in a v2 master aware contract we can add them at the very end.

roxdanila (approved):

PR #223, opened 2022-05-20
Implement new reward shares formula & fix extendDeposit

horodnicdragos:


contracts/modules/v2/StakingPool.sol
```diff
@@ -617,8 +673,8 @@ contract StakingPool is IStakingPool, ERC721 {
 
 
     // Move the user's shares from the initial deposit to the new one.
-    newDeposit.rewardsShares += initialDeposit.rewardsShares;
-    newDeposit.stakeShares += initialDeposit.stakeShares;
+    newDeposit.rewardsShares += initialDeposit.rewardsShares + newRewardsShares;
```
danoctavian, line 676: isn't this supposed to be:

```newDeposit.rewardsShares += newRewardsShares;````

the += is confusing.

if ```newDeposit.rewardsShares``` is zero i'd just do +=

File: contracts/modules/v2/StakingPool.sol [hunk 2]
```diff
@@ -348,21 +356,55 @@ contract StakingPool is IStakingPool, ERC721 {
...
+  /// @param initialStakeShares  The amount of stake shares that the deposit is already entitled
+  ///                            to.
+  /// @param newStakeShares      The amount o new stake shares that the deposit will entitled to
+  ///                            on top of the existing ones.
+  /// @param initialTrancheId    The initial id of the tranche that defines the deposit period.
+  /// @param newTrancheId        The new id of the tranche that will define the deposit period.
+  /// @param blockTimestamp      The timestamp of the block when the new shares are recalculated.
+  function calculateNewRewardShares(
+    uint initialStakeShares,
+    uint newStakeShares,
```
danoctavian, line 379: So `newStakeShares` is on top of `initialStakeShares`

`extraStakeShares` could be more clear

contracts/modules/v2/StakingPool.sol
horodnicdragos: The new deposit can be a deposit that already exists which is topped up with the new rewards shares and the reward shares from the initial deposit which is extended.

horodnicdragos: Would renaming `new` to `target` dispell the confusion and make it more obvious that the deposit is not guaranteed to be 0?

contracts/modules/v2/StakingPool.sol
horodnicdragos: The term "newShares" is already used throughout the contract and I do agree it doesn't clearly tell whether it is the final state after the transaction or something that only adds up to the initial state resulting in the final state.

I'd suggest `newStakeShares` -> `stakeSharesIncrease` and apply this pattern it consistently.

danoctavian: `stakeSharesIncrease` good enough

contracts/modules/v2/StakingPool.sol
danoctavian: ```The new deposit can be a deposit that already exists which is topped up with the new rewards shares and the reward shares from the initial deposit which is extended.```

how about a comment on this. this makes sense

PR #224, opened 2022-05-24
Fix burn function

danoctavian:


contracts/modules/cover/CoverViewer.sol
```diff
@@ -99,4 +99,61 @@ contract CoverViewer {
     }
     return coverViews;
   }
+
+  /* ========== COVER PRICING VIEWS ========== */
+
+  function getPoolAllocationPriceParametersForProduct(
```
danoctavian, line 105: moved this here to lower Cover.sol size back under 24k

contracts/modules/token/TokenController.sol
```diff
@@ -24,6 +25,9 @@ contract TokenController is ITokenController, LockHandler, LegacyMasterAware {
   IAssessment public assessment;
   IGovernance public governance;
 
+  ICover public cover;
+  uint pooledStakingNXMBalance;
```
danoctavian, line 29: if we track this as such, we would need to add all NXM balance changes, including deposits an wtihdrawals from staking pools here.

contracts/interfaces/ICover.sol
```diff
@@ -31,12 +31,12 @@ enum CoverUintParams {
...
 
 struct PoolAllocationRequest {
-  uint64 poolId;
+  uint40 poolId;
   uint coverAmountInAsset;
 }
 
 struct PoolAllocation {
-  uint64 poolId;
+  uint40 poolId;
```
danoctavian, line 39: initially i had this lowered to fit the globalRewardsRatio parameter but that should be associated with the segment.

However i don't see a downside to making it smaller and potentially fitting another data field in there later.

contracts/modules/cover/Cover.sol
```diff
@@ -29,22 +29,18 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
-  uint public constant PRICE_DENOMINATOR = 10000;
-  uint public constant COMMISSION_DENOMINATOR = 10000;
-  uint public constant CAPACITY_REDUCTION_DENOMINATOR = 10000;
-  uint public constant GLOBAL_CAPACITY_DENOMINATOR = 10_000;
-
-  uint public constant BURN_DENOMINATOR = 1e18;
+  uint private constant PRICE_DENOMINATOR = 10000;
+  uint private constant COMMISSION_DENOMINATOR = 10000;
+  uint private constant CAPACITY_REDUCTION_DENOMINATOR = 10000;
+  uint private constant GLOBAL_CAPACITY_DENOMINATOR = 10_000;
```
horodnicdragos, line 35: Shouldn't the separator be `100_00` instead of `10_000` (100.00% vs 10.000%)?

contracts/modules/cover/Cover.sol
```diff
@@ -29,22 +29,18 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
+  uint private constant REWARD_DENOMINATOR = 10_000;
 
   uint public constant MAX_COVER_PERIOD = 365 days;
-  uint public constant MIN_COVER_PERIOD = 30 days;
+  uint private constant MIN_COVER_PERIOD = 30 days;
```
horodnicdragos, line 39: Shouldn't this be 28 days now?

contracts/modules/cover/Cover.sol
```diff
@@ -423,7 +436,10 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
       IStakingPool _stakingPool = stakingPool(allocation.poolId);
 
-      uint nxmBurned = allocation.coverAmountInNXM * burnAmount / segment.amount;
+      uint nxmBurned = allocation.coverAmountInNXM
+        * burnAmount / segment.amount
+        * GLOBAL_CAPACITY_DENOMINATOR / globalCapacityRatio;
+
       _stakingPool.burnStake(cover.productId, segment.start, segment.period, nxmBurned);
 
       allocation.coverAmountInNXM -= SafeUintCast.toUint96(nxmBurned);
```
horodnicdragos, line 445: I think there's something fishy here. Assuming a 2x global capacity ratio and a cover worth 10 NXM. If a payout on a claim is 10 NXM, the coverAmountInNXM should be 0 aftewards, not 5 NXM. I believe this partial burn only makes sense in the context of staking when calling burnStake.

contracts/interfaces/ICover.sol
horodnicdragos: Have you thought about adding a minimum staking deposit when creating a pool to avoid a potential attack where a spammer depletes all possible ids?

horodnicdragos (changes requested):

contracts/interfaces/ICover.sol
danoctavian: on the L1  gas costs should be a deterrant for that.

BUT i don't see any downside to that, it can be a fixed NXM value. 

Something like 100 NXM?

contracts/modules/token/TokenController.sol
```diff
@@ -24,6 +25,10 @@ contract TokenController is ITokenController, LockHandler, LegacyMasterAware {
   IAssessment public assessment;
   IGovernance public governance;
 
+  ICover public cover;
+  mapping(uint => uint) stakingPoolNXMRewards;
```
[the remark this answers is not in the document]
danoctavian: So here I actually want to verify if the caller is a staking pool.

To do this i ask the caller to provide its pool id, and then use the stakingPool() function to verifiy msg.sender is the pool id with that id.

If i don't do this, and i don't ask for the id, i'd have to keep some register of pool staking addresses.

I think this is simpler and achieves authentication. What i would think here is: is there any way to circumvent this authentication? seem like not because all valid staking pools can only be created by cover.sol

contracts/modules/v2/StakingPool.sol
```diff
@@ -89,6 +92,8 @@ contract StakingPool is IStakingPool, ERC721 {
   // token id => tranche id => deposit data
   mapping(uint => mapping(uint => Deposit)) public deposits;
 
+  ITokenController public tokenController;
```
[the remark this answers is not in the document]
danoctavian: cool, once we get that merged

contracts/modules/token/TokenController.sol
```diff
@@ -24,6 +25,10 @@ contract TokenController is ITokenController, LockHandler, LegacyMasterAware {
...
+  mapping(uint => uint) stakingPoolDeposits;
```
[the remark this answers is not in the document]
danoctavian: They can but you don't really use the 2 values at the same time, does it matter?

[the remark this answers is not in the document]
horodnicdragos: As an example you (are supposed to) use both values in the withdraw function from staking pool which must decrease either one or both of them.

contracts/modules/cover/Cover.sol
```diff
@@ -442,7 +442,8 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
 
       _stakingPool.burnStake(cover.productId, segment.start, segment.period, nxmBurned);
 
-      allocation.coverAmountInNXM -= SafeUintCast.toUint96(nxmBurned);
+      uint coverAmountInNXMRemoved = allocation.coverAmountInNXM * burnAmount / segment.amount;
```
horodnicdragos, line 445: This is basically `payoutAmountInNXM`.

horodnicdragos (approved):

contracts/modules/cover/Cover.sol
```diff
@@ -240,12 +235,20 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
         requestedCoverAmountInNXM
       );
 
+      // apply the global rewards ratio and the total Rewards in NXM
+      uint totalRewardsInNXM = premiumInNXM * globalRewardsRatio / REWARD_DENOMINATOR;
```
shark0der, line 239: I'm doing the same calculation in the staking contract. Let's return it from the staking contract so we'd have it calculated in a single place, less room for error.

contracts/modules/cover/Cover.sol
```diff
@@ -327,9 +332,17 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
...
+            lastCoverSegment.globalRewardsRatio
           );
+
           originalPoolAllocations[i].premiumInNXM *= (lastCoverSegment.period - remainingPeriod) / lastCoverSegment.period;
+
+          // compute NXM rewards deallocated
+          uint deallocatedRewardsInNXM = originalPoolAllocations[i].premiumInNXM
+          * remainingPeriod / lastCoverSegment.period
+          * lastCoverSegment.globalRewardsRatio / REWARD_DENOMINATOR;
+          tokenController().burnPooledStakingNXMRewards(deallocatedRewardsInNXM, originalPoolAllocations[i].poolId);
```
shark0der, line 345: Let's create 2 arrays, one with nxm amounts, one with pool ids, and pass them in 1 call to TokenController

contracts/modules/cover/Cover.sol
```diff
@@ -521,14 +538,19 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
 
     // [todo] handle the creation of NFT 0 which is the default NFT owned by the pool manager
     return CoverUtilsLib.createStakingPool(
```
shark0der, line 540: The purpose of this library/function was to move the proxy bytecode out of the Cover contract to keep it small but we ended up moving a lot of logic in there, and the number of parameters keeps growing. I'd opt for having minimal required code in the library and keep the core logic here. This will also avoid some redundant memory copying.

contracts/modules/token/TokenController.sol
```diff
@@ -347,6 +352,35 @@ contract TokenController is ITokenController, LockHandler, LegacyMasterAware {
...
+
+  function burnPooledStakingNXMRewards(uint amount, uint poolId) external {
+
+    require(msg.sender == address(cover), "TokenController: only Cover allowed");
+    burnFrom(address(this), amount);
+    stakingPoolNXMBalances[poolId].rewards -= amount.toUint96();
+  }
+
+  function depositStakedNXM(address from, uint amount, uint poolId) external {
+    require(msg.sender == address(cover.stakingPool(poolId)), "TokenController: msg.sender not staking pool");
```
shark0der, line 371: We should have an internal library with a hardcoded proxy bytecode hash that will calculate the address of a staking pool given its id. having an external call for this is really unnecessary. See uniswap's library as an example: https://github.com/Uniswap/v2-periphery/blob/master/contracts/libraries/UniswapV2Library.sol#L17

contracts/modules/token/TokenController.sol
```diff
@@ -347,6 +352,35 @@ contract TokenController is ITokenController, LockHandler, LegacyMasterAware {
     token.transfer(user, totalAmount);
   }
 
+
+  function mintPooledStakingNXMRewards(uint amount, uint poolId) external {
+
+    require(msg.sender == address(cover), "TokenController: only Cover allowed");
+    mint(address(this), amount);
+    stakingPoolNXMBalances[poolId].rewards += amount.toUint96();
```
shark0der, line 360: Deposits and rewards are 128 bits.

shark0der: Note: fix other occurrences as well.

contracts/modules/v2/StakingPool.sol
```diff
@@ -448,8 +457,7 @@ contract StakingPool is IStakingPool, ERC721 {
 
       uint withdrawable = stakeToWithdraw + rewardsToWithdraw;
 
-      // TODO: use TC instead
-      nxm.transfer(ownerOf(tokenId), withdrawable);
+      tokenController.withdrawStakedNXM(ownerOf(tokenId), withdrawable, poolId);
```
shark0der, line 460: `withadrawable` includes stake (deposits) and rewards but the function handles only deposits. Let's make it accept both.

contracts/modules/cover/Cover.sol
danoctavian: purpose more generally is to reduce bytecode size in Cover.sol if we can move logic there so it stays under 24k

so i intentionally moved logic there

contracts/modules/cover/Cover.sol
danoctavian: discussed off-band. we can keep it with 1 param instead of array, you save around 100 gas and it's a rare case that you have multiple allocations.

contracts/modules/token/TokenController.sol
danoctavian: in theory it sounds good but this function needs to 2 parameters that are implicit there

`stakingPoolProxyCodeHash`

`address(cover)```` which is ```address(this)```

if we just make a library they need to be read anyway and passed in feels more awkward. the first one is from Cover.sol

roxdanila: Let's implement this post V2

PR #226, opened 2022-05-30
Feature/staking manager voting power

horodnicdragos:


hardhat.config.js/solidity.js
```diff
@@ -1,16 +1,13 @@
...
+  '0.5.7': { settings: compilerSettings, version: '0.5.7' }, // nexus mutual v1
   '0.5.17': { settings: compilerSettings, version: '0.5.17' }, // nexus mutual v1
   '0.8.9': { settings: compilerSettings, version: '0.8.9' }, // nexus mutual v2
 };
 
 const contracts = {
-  '0.5.7': [
-    'contracts/modules/governance/Governance.sol',
-  ],
+  '0.5.7': ['contracts/modules/governance/Governance.sol'],
```
shark0der, line 10: Please revert the modifications to the lines that weren't modified.

contracts/modules/v2/StakingPool.sol
```diff
@@ -747,15 +759,53 @@ contract StakingPool is IStakingPool, ERC721 {
...
+   * @dev See {IERC721-safeTransferFrom}.
+   */
+  function safeTransferFrom(
+      address from,
+      address to,
+      uint256 tokenId,
+      bytes memory _data
+  ) public override(ERC721, IERC721) {
+      _beforeTokenTransfer(from, to, tokenId);
+      super.safeTransferFrom(from, to, tokenId, _data);
```
shark0der, line 808: You correctly identified the fact that the hooked is not invoked, my bad. However the fix is to call `super._beforeTokenTransfer(from, to, tokenId);` inside the `_beforeTokenTransfer` function, not calling it in all transfer functions. https://docs.openzeppelin.com/contracts/4.x/extending-contracts#rules_of_hooks

contracts/modules/v2/StakingPool.sol
```diff
@@ -747,15 +759,53 @@ contract StakingPool is IStakingPool, ERC721 {
...
     uint256 tokenId
   ) internal view override {
-    require(
-      tokenId != 0 || nxm.isLockedForMV(from) < block.timestamp,
-      "StakingPool: Locked for voting in governance"
-    );
-    // todo: track owned zero-id nfts in TC
+    if(tokenId == 0) {
+      require(
+        nxm.isLockedForMV(manager()) < block.timestamp,
```
shark0der, line 768: Can you explain the change from `from` to `manager()`?

contracts/modules/v2/StakingPool.sol
```diff
@@ -36,7 +37,7 @@ import "../../utils/Math.sol";
 // │                               │
 // ╰───────────────────────────────╯
 
-contract StakingPool is IStakingPool, ERC721 {
+contract StakingPool is ERC721, IStakingPool {
```
shark0der, line 40: Why was the order changed?

contracts/modules/v2/StakingPool.sol
```diff
@@ -411,6 +413,15 @@ contract StakingPool is IStakingPool, ERC721 {
     WithdrawParams[] memory params
   ) public returns (uint stakeToWithdraw, uint rewardsToWithdraw) {
 
+    // Prevent withdrawals while the manager is locked in governance
+    {
+      uint managerLockedInGovernanceUntil = nxm.isLockedForMV(manager());
+      require(
+        managerLockedInGovernanceUntil > block.timestamp,
```
shark0der, line 420: Shouldn't the operator be `<`? The condition should say `lockedForMV` timestamp to be in the past.

contracts/modules/v2/StakingPool.sol
```diff
@@ -769,6 +819,10 @@ contract StakingPool is IStakingPool, ERC721 {
 
   /* views */
 
+  function governance() internal view returns (IGovernance) {
+    return IGovernance(getInternalContractAddress(ID.GV));
+  }
+
```
shark0der, line 825: Leftovers?

contracts/modules/v2/StakingPool.sol
horodnicdragos: The manager's lock is the one that matters. When the manager is locked, his manager NFT is non-transferable for 3 days. Other NFTs remain transferable since all withdrawals are halted so there's no double voting.

hardhat.config.js/solidity.js
horodnicdragos: Formatted according to the rules found in `.prettierrc`.

contracts/modules/governance/Governance.sol
```diff
@@ -865,8 +865,11 @@ contract Governance is IGovernance, LegacyMasterAware {
           stakingPool.manager() == msg.sender,
           "One or more staking pools are not managed by the sender"
         );
-        uint activeStake = stakingPool.activeStake();
-        voteWeight = voteWeight.add(activeStake);
+        (
+          uint stakingRewards,
+          uint stakingDeposits
+        ) = tokenInstance.stakingPoolNXMBalances(managedStakingPoolIds[i]);
+        voteWeight = voteWeight.add(stakingRewards).add(stakingDeposits);
```
shark0der, line 872: Rewards are not supposed to be used because they can be withdrawn at any time regardless of the voting lock.

horodnicdragos: Not anymore, this was changed today. The `withdraw` function should always revert while the manager is locked.

horodnicdragos: On a second thought I see no reason why we wouldn't allow reward withdrawals since rewards are negligible given the 15% voting power cap, so never mind. See the commit below.

contracts/modules/v2/StakingPool.sol
shark0der: We're inside the if condition. The `from` has to be the manager address so there's no point in fetching it from the storage. If the `from` is not the manager it will revert inside the ERC721 transfer function.

horodnicdragos: What if it's an approved contract?

roxdanila (approved):

PR #232, opened 2022-06-08
Feature/extra cover tests

danoctavian:


contracts/modules/cover/Cover.sol
```diff
@@ -416,7 +444,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
     );
   }
 
-  function performPayoutBurn(
+  function performStakeBurn(
```
horodnicdragos, line 447: What's the rationale behind this rename? Isn't "burnAmount" supposed to be the amount paid out (i.e. ETH/DAI, not NXM)? It definitely isn't the stake (in NXM) that is burned and we don't necessarily burn NXM when we make a payout, that's just a consequence of the payout.

contracts/modules/cover/Cover.sol
```diff
@@ -301,7 +316,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon {
     uint coverId,
     BuyCoverParams memory buyCoverParams,
     PoolAllocationRequest[] memory poolAllocations
-  ) external payable onlyMember whenNotPaused {
+  ) external payable onlyMember whenNotPaused nonReentrant {
```
roxdanila, line 319: remove the require below:

`require(lastCoverSegment.start + lastCoverSegment.period > block.timestamp, "Cover: cover expired");`

danoctavian: removed.

Changed condition in the IF to check for that.

contracts/modules/cover/Cover.sol
danoctavian: Discussed off-band.

roxdanila (approved):

roxdanila (approved):

PR #238, opened 2022-06-17
move function to library

danoctavian:


contracts/modules/cover/CoverUtilsLib.sol
```diff
@@ -13,6 +13,8 @@ import "./MinimalBeaconProxy.sol";
 
 library CoverUtilsLib {
 
+  uint private constant GLOBAL_CAPACITY_DENOMINATOR = 10_000;
```
horodnicdragos, line 16: Since we have this in Cover, CoverUtilsLib, CoverViewere and StakingPool and it's the same value, wouldn't it be better to keep constants in a single library and import them from there?

contracts/modules/cover/CoverUtilsLib.sol
```diff
@@ -160,4 +162,36 @@ library CoverUtilsLib {
...
+
+    // increase amountPaidOut only *after* you read the segment
+    cover.amountPaidOut += SafeUintCast.toUint96(burnAmount);
+
+    uint allocationCount = allocations.length;
+    for (uint i = 0; i < allocationCount; i++) {
+
+      PoolAllocation memory allocation = allocations[i];
+
+      uint nxmBurned = allocation.coverAmountInNXM
```
horodnicdragos, line 185: Would `burnAmountInNXM` be more in line with `payoutAmountInNXM`?

danoctavian: nxmBurned is how much its burned because you apply all the factors, think this is ok

danoctavian: ah i see what you mean

danoctavian: we did a rename for payoutAmountInNXM

There is no variable called burnAmountInNXM

Maybe you want to rename burnAmount?

contracts/modules/cover/CoverUtilsLib.sol
danoctavian: the issue i think here is that it'ss basically 1 or 2 common constants across those...

If you do a common lib i guess it helps with code tracking but you'd still do a full redeploy of all

horodnicdragos (approved):

PR #255, opened 2022-07-25
Fix/active cover track payout

danoctavian:


danoctavian:
Still need a test for active cover amount decrease on payout.

shark0der (approved):

shark0der (approved):

PR #256, opened 2022-07-26
Feature: v2 local deploy

shark0der:


contracts/modules/capital/MCR.sol
```diff
@@ -229,8 +234,8 @@ contract MCR is IMCR, MasterAware {
    * @param code parameter code
    * @param val new value
    */
-  function updateUintParameters(bytes8 code, uint val) public {
-    require(master.checkIsAuthToGoverned(msg.sender));
```
roxdanila, line 233: is this function used anywhere else? prolly worth getting rid of it

contracts/modules/token/TokenController.sol
```diff
@@ -36,6 +36,10 @@ contract TokenController is ITokenController, LockHandler, LegacyMasterAware {
     claimsReward = claimsRewardAddress;
   }
 
+  function unlistClaimsReward() external {
+    token.removeFromWhiteList(claimsReward);
```
roxdanila, line 40: added an issue to manage this in master instead

contracts/modules/capital/MCR.sol
roxdanila: added an issue for this, 1 more place where it needs to be replaced

roxdanila (approved):

contracts/modules/capital/MCR.sol
```diff
@@ -73,12 +72,18 @@ contract MCR is IMCR, MasterAware {
...
 
+    // copy over values
     mcrFloor = previousMCR.mcrFloor();
-
     mcr = previousMCR.mcr();
     desiredMCR = previousMCR.desiredMCR();
     lastUpdateTime = previousMCR.lastUpdateTime();
+
+    // copy over parameters
+    mcrFloorIncrementThreshold = previousMCR.mcrFloorIncrementThreshold();
```
danoctavian, line 82: we should add to the v2 fork test to validate all values are moved correctly.

We will be upgrading MCR.sol to get the new logic for the MCR gearing active cover amount computation.

contracts/modules/token/TokenController.sol
danoctavian: any particular reason to do this in master and not here?

danoctavian (approved):
LGTM

contracts/modules/token/TokenController.sol
roxdanila: I guess there are multiple contracts that require this

PR #258, opened 2022-08-02
Fix claim assessment timelock

shark0der:


danoctavian:
worth having a unit/integration test for this new logic

contracts/modules/claims/ClaimsReward.sol
```diff
@@ -529,10 +528,25 @@ contract ClaimsReward is IClaimsReward, LegacyMasterAware {
...
+      uint lockTimePerVote = td.lockCADays();
+      uint reduction = lockTimePerVote.mul(counter);
+      uint minExpiration = block.timestamp.add(lockTimePerVote);
+      uint currentExpiration = tc.getLockedTokensValidity(msg.sender, "CLA");
+
+      // reduce lock time only if current lock time is greater than min lock time
+      if (currentExpiration > minExpiration) {
+
+        // and never reduce to less than min lock time
+        if (currentExpiration.sub(reduction) < minExpiration) {
```
danoctavian, line 542: so this means that you will always have to wait at least min lock time correct?

There's no way to get this lock time down to block.timestamp

contracts/modules/claims/ClaimsReward.sol
```diff
@@ -529,10 +528,25 @@ contract ClaimsReward is IClaimsReward, LegacyMasterAware {
...
+          reduction = currentExpiration - minExpiration;
```
danoctavian, line 543: use .sub here

roxdanila (changes requested):
Fix looks good.
Please add a test as discussed offline and then loop in Iosiro for the final review.

roxdanila (approved):

PR #336, opened 2022-09-01
Create pull_request_template.md

### LINKED ISSUE #285: Introduce PR templates - smart contracts repo

roxdanila:
Open for feedback. The goal is to standardize how we handle PRs and make reviewers' life easier.
Closes #285

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,42 @@
+## Description
+
+Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
+
+Fixes # (issue)
+
+### Type of change
+
+Please delete options that are not relevant.
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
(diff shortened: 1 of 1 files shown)


mixplore (approved):
LGTM!

shark0der:
Overall looks very good.

Inspired by [DFE Digital](https://github.com/DFE-Digital/find-a-lost-trn/blob/main/.github/pull_request_template.md) I've got some suggestions:

1. Let's split the description in two sections: "Context" and "Changes proposed in this pull request" so we'd start with the "why" and then the "what" while also preventing us from skipping one of the two.
2. Just like DFE Digital, I would suggest to start using rebase more often. This will result in a much cleaner history and avoid tran rails style git log graph so we can actually follow the history and make sense of it. To avoid conflicts and issues it's important to rebase often so we'd catch bugs and changes early on. PRs should also be relatively small.
3. I'm slightly against heavy commented code. At times the comments (especially when there're walls of text) do not reflect the code anymore. I think the code should be clear enough by itself (proper var/fn names etc) so it's self documented. Comments should be added if the intention is not obvious.

roxdanila:
Addressed @shark0der's comments

PR #356, opened 2022-09-05
remove amount as a field. Only allocations are relevant

danoctavian:


test/unit/Cover/buyCover.js
```diff
@@ -572,7 +572,60 @@ describe('buyCover', function () {
...
+        amount: tooLargeExpectedAmount,
+        period,
+        maxPremiumInAsset: expectedPremium,
+        paymentAsset: payoutAsset,
+        payWitNXM: false,
+        commissionRatio: parseEther('0'),
+        commissionDestination: ZERO_ADDRESS,
+        ipfsData: ''
+      },
+      [{ poolId: '0', coverAmountInAsset: amount.toString() }],
```
shark0der, line 625: Never cast to string when calling contract functions. Pass the BN instead.

test/unit/Cover/buyCover.js
```diff
@@ -572,7 +572,60 @@ describe('buyCover', function () {
...
+      {
+        value: expectedPremium,
+      },
+    )).to.be.revertedWith('Cover Insufficient cover amount');
```
shark0der, line 629: :hammer: The revert reason wording suggests that the user didn't buy enough cover when in reality we ran out of capacity on some of the provided pools. Let's change the reason.

I'd suggest something like `Cover: The selected pools ran out of capacity`

shark0der (changes requested):
Needs to address the comments and a rebase after the other PR is merged.

contracts/modules/cover/Cover.sol
```diff
@@ -229,16 +228,18 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
   ) internal returns (uint totalPremiumInNXM, uint) {
 
     // convert to NXM amount
-    uint nxmPriceIncoverAsset = pool().getTokenPrice(params.coverAsset);
```
danoctavian, line 232: this was missed on the previous find and replace refactor (the capitalization)

contracts/modules/cover/Cover.sol
```diff
@@ -182,7 +182,6 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     require(params.period >= MIN_COVER_PERIOD, "Cover: Cover period is too short");
     require(params.period <= MAX_COVER_PERIOD, "Cover: Cover period is too long");
     require(params.commissionRatio <= MAX_COMMISSION_RATIO, "Cover: Commission rate is too high");
-    require(params.amount > 0, "Cover: amount = 0");
```
shark0der, line 185: We should keep this require statement because the allocations array could be empty and the final cover amount will end up being zero.

shark0der (approved):
Looks good :+1:

Issue #314, opened 2022-08-31
Audit - Pool 3.3.1 Oracle availability not verified when adding new assets


roxdanila:
[Pool.sol](https://github.com/NexusMutual/smart-contracts/blob/14a7ed6f5b6df877a2cb7eb9fd390c9cb216ccb2/contracts/modules/capital/Pool.sol#L207)

**Description**
The function addAsset() does not confirm whether a valid priceOracle exists when adding that asset to the system. If an asset without a valid oracle was added, all subsequent actions that require awareness of the full system value would revert, as the system would be unable to calculate the value for the added asset.

For example, when calls are made to getPoolValueInEth() in Pool.sol, getAssetValueInEth() is called which then calls priceFeedOracle.getAssetToEthRate(), supplying the desired asset's address. PriceFeedOracle.sol will then obtain the aggregator address from the contract state which, if no aggregator has been preconfigured, will return the zero address. This zero address will then cause the operation to revert, as _getAssetToEthRate() in PriceFeedOracle.sol includes a require, reverting if the aggregator is the zero address.

**Remediation**
The PriceFeedOracle.sol contract does not currently contain a setter function to add new aggregators for assets. It is recommended that such a function be added, and a modifier be included that only permits Pool.sol to call it. With this functionality in place, the addAsset() function in Pool.sol should also accept the aggregator address for the newly-created asset, which it can then use to update PriceFeedOracle.sol accordingly.

roxdanila:
Fixed in https://github.com/NexusMutual/smart-contracts/pull/362

PR #362, opened 2022-09-07
Check that when an asset is added to the pool it has an oracle

danoctavian:
## Context

Fixes issue: https://github.com/NexusMutual/smart-contracts/issues/314

We want to prevent addition of assets don't have a registered oracle.

This prevents any issue with pool value calculations that may lock the system.

Add a require to the addAsset function in Pool.sol

## Test plan

Upgrade all tests that interact with addAsset to now have a oracle available.

`npm run test-unit`

## Checklist
- [x] Rebased the base branch

### DIFF
--- a/contracts/modules/capital/Pool.sol
+++ b/contracts/modules/capital/Pool.sol
@@ -218,6 +218,9 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
     require(_max >= _min, "Pool: max < min");
     require(_maxSlippageRatio <= MAX_SLIPPAGE_DENOMINATOR, "Pool: Max slippage ratio > 1");
 
+    (address aggregator, ) = priceFeedOracle.assets(assetAddress);
+    require(aggregator != address(0), "Pool: Asset lacks oracle");
+
     // Check whether the new asset already exists as a cover asset
     uint coverAssetsCount = coverAssets.length;
     for (uint i = 0; i < coverAssetsCount; i++) {
--- a/contracts/modules/capital/Pool.sol
+++ b/contracts/modules/capital/Pool.sol
@@ -2,7 +2,6 @@
 
 pragma solidity ^0.8.16;
 
-import "@openzeppelin/contracts-v4/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts-v4/utils/Address.sol";
 import "@openzeppelin/contracts-v4/token/ERC20/utils/SafeERC20.sol";
 import "@openzeppelin/contracts-v4/security/ReentrancyGuard.sol";
@@ -741,14 +740,10 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
       for (uint i = 0; i < coverAssetsCount; i++) {
         Asset memory asset = coverAssets[i];
 
-        (address aggregator, ) = priceFeedOracle.assets(asset.assetAddress);
-
-        IERC20Detailed erc20 = IERC20Detailed(asset.assetAddress);
-
         if (asset.assetAddress != ETH) {
-          require(aggregator != address(0), "Pool: Asset lacks oracle");
+          (address aggregator, ) = priceFeedOracle.assets(asset.assetAddress);
+          require(aggregator != address(0), "Pool: Oracle lacks asset");
         }
-
       }
 
       priceFeedOracle = IPriceFeedOracle(value);
--- a/contracts/modules/capital/Pool.sol
+++ b/contracts/modules/capital/Pool.sol
@@ -13,6 +13,7 @@ import "../../interfaces/INXMToken.sol";
 import "../../interfaces/IPool.sol";
 import "../../interfaces/IPriceFeedOracle.sol";
 import "../../interfaces/ITokenController.sol";
+import "../../interfaces/IERC20Detailed.sol";
 import "../../libraries/SafeUintCast.sol";
 
 contract Pool is IPool, MasterAware, ReentrancyGuard {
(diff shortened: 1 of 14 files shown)


roxdanila (changes requested):
Looks like tests are failing. If this isn't ready for review, please make it draft PR.

test/unit/CowOperator/setup.js
```diff
@@ -46,12 +46,14 @@ async function setup() {
...
+  await usdcAggregator.setLatestAnswer(0.0002 * 1e18); // 1 usdc = 0.0002 eth, 1 eth = 5000 dai
 
   // Deploy PriceFeedOracle
   const priceFeedOracle = await PriceFeedOracle.deploy(
-    [dai.address, stEth.address],
-    [daiAggregator.address, stethAggregator.address],
-    [18, 18],
+    [dai.address, stEth.address, usdc.address],
+    [daiAggregator.address, stethAggregator.address, usdcAggregator.address],
+    [18, 18, 18],
```
shark0der, line 56: `usdc` was deployed with 6 decimals (see line 32).

test/unit/CowOperator/setup.js
```diff
@@ -46,12 +46,14 @@ async function setup() {
   await daiAggregator.setLatestAnswer(0.0002 * 1e18); // 1 dai = 0.0002 eth, 1 eth = 5000 dai
   const stethAggregator = await ChainlinkAggregatorMock.deploy();
   await stethAggregator.setLatestAnswer(parseEther('1')); // 1 steth = 1 eth
+  const usdcAggregator = await ChainlinkAggregatorMock.deploy();
+  await usdcAggregator.setLatestAnswer(0.0002 * 1e18); // 1 usdc = 0.0002 eth, 1 eth = 5000 dai
```
shark0der, line 50: `usdc` was deployed with 6 decimals (see line 32).

test/unit/Pool/addAsset.js
```diff
@@ -63,11 +83,21 @@ describe('addAsset', function () {
...
 
+    const chainlinkNewAsset = await ChainlinkAggregatorMock.new();
+    await chainlinkNewAsset.setLatestAnswer(new BN((1e18).toString()));
+    const priceFeedOracle = await PriceFeedOracle.new(
+      [dai.address, stETH.address, token.address],
+      [chainlinkDAI.address, chainlinkSteth.address, chainlinkNewAsset.address],
+      [18, 18, 18],
+    );
+
+    await pool.updateAddressParameters(hex('PRC_FEED'), priceFeedOracle.address, { from: governance });
```
shark0der, line 99: The new price feed does not contain the asset that you added to the pool at the very top of this file (inside beforeEach). The solution would be that whenever we set a new price feed - the pool should check that it has aggregators for all its assets. This would also prevent us from pushing a bad address for the price feed (typo in address resulting in adding an EOA for the price feed that will always revert).

test/unit/Pool/addAsset.js
```diff
@@ -1,17 +1,37 @@
...
+    const chainlinkNewAsset = await ChainlinkAggregatorMock.new();
+    await chainlinkNewAsset.setLatestAnswer(new BN((1e18).toString()));
+
+    const priceFeedOracle = await PriceFeedOracle.new(
+      [dai.address, stETH.address, assetAddress],
+      [chainlinkDAI.address, chainlinkSteth.address, chainlinkNewAsset.address],
+      [18, 18, 18],
+    );
+
+    await pool.updateAddressParameters(hex('PRC_FEED'), priceFeedOracle.address, { from: governance });
```
shark0der, line 32: The operations in this beforeEach should happen inside setup.js, and setup the price feed correctly from the start, not create one in the setup and then replace it before each test.

test/unit/CowOperator/setup.js
danoctavian: because this is USDC/ETH and not ETH/USDC the number of decimals here is 18 (value is in ETH)

Check the chainlink prod oracle as an example https://data.chain.link/ethereum/mainnet/stablecoins/usdc-eth#operator-alpha_chain

contracts/modules/capital/Pool.sol
```diff
@@ -731,6 +735,22 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
     if (code == "PRC_FEED") {
+
+      uint coverAssetsCount = coverAssets.length;
+
+      for (uint i = 0; i < coverAssetsCount; i++) {
+        Asset memory asset = coverAssets[i];
+
+        (address aggregator, ) = priceFeedOracle.assets(asset.assetAddress);
+
+        IERC20Detailed erc20 = IERC20Detailed(asset.assetAddress);
```
shark0der, line 746: Can you please explain why do you need the `IERC20Detailed erc20` here?

contracts/modules/capital/Pool.sol
```diff
@@ -731,6 +735,22 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
     }
 
     if (code == "PRC_FEED") {
+
+      uint coverAssetsCount = coverAssets.length;
+
+      for (uint i = 0; i < coverAssetsCount; i++) {
+        Asset memory asset = coverAssets[i];
+
+        (address aggregator, ) = priceFeedOracle.assets(asset.assetAddress);
```
shark0der, line 744: This call seems redundant if the asset is ETH. Move this call inside the if condition below.

test/unit/CowOperator/setup.js
shark0der, line 50: Although you have changed the decimals below, the exchange rate set here still uses 18 decimals. Shouldn't have this failed some tests?

contracts/modules/capital/Pool.sol
```diff
@@ -731,6 +735,22 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
+
+        IERC20Detailed erc20 = IERC20Detailed(asset.assetAddress);
+
+        if (asset.assetAddress != ETH) {
+          require(aggregator != address(0), "Pool: Asset lacks oracle");
```
shark0der, line 749: The revert reason should be `Pool: Oracle lacks asset`.

shark0der (changes requested):
There doesn't seem to be any test for the 2 new require statements.

shark0der (approved):

PR #365, opened 2022-09-07
Chore: setup eslint prettier

shark0der:
This PR adds eslint and prettier and configures eslint to use prettier for formating. From the developer's point of view one only needs to run `eslint` in order to lint the code. `eslint --fix` can be used to automatically fix issues but it can't always fix everything and manual intervention is required from time to time.

It is possible to configure your IDE to run `eslint --fix` on save or add a git commit hook to do that if you wish - personally I'm running it manually every time. If one forgets to fix the issues, this will be detected by the new lint CI stage.

There were a few issues in the tests besides the stylistic ones that I fixed. All the tests that were running before are still running and are passing.

PS: Make sure to run `npm i` if you want to test it locally


kyledewy:
LGTM! :rocket:

test/unit/Cover/helpers.js
```diff
@@ -71,35 +57,39 @@ async function assertCoverFields (
...
+    stakingPoolManager,
+    targetPriceRatio,
   );
 
-  const expectedPremium = amount.mul(targetPriceRatio).div(priceDenominator).mul(period).div(3600 * 24 * 365);
+  const expectedPremium = amount
+    .mul(targetPriceRatio)
+    .div(priceDenominator)
+    .mul(period)
+    .div(3600 * 24 * 365);
```
mixplore, line 88: 🍰 Use `daysToSeconds` helper method here maybe?

test/integration/setup.js
```diff
@@ -22,16 +20,16 @@ const web3ToEthers = (x, signers) => {
...
 };
 
-async function setup () {
+async function setup() {
   // external
   const ERC20BlacklistableMock = artifacts.require('ERC20BlacklistableMock');
   const OwnedUpgradeabilityProxy = artifacts.require('OwnedUpgradeabilityProxy');
   const ChainlinkAggregatorMock = artifacts.require('ChainlinkAggregatorMock');
-  const Lido = artifacts.require('P1MockLido');
+  // const Lido = artifacts.require('P1MockLido');
```
mixplore, line 32: Should this and the lines below be commented out?

mixplore (approved):
Left a few non-blocking comments, but other than that it looks good! 👍  Great work!! 🚀

test/unit/YieldTokenIncidents/redeemPayout.js
```diff
@@ -1,4 +1,4 @@
-const { ethers } = require('hardhat');
+const { ethers, network } = require('hardhat');
 const { expect } = require('chai');
 
 const { daysToSeconds, setTime, ASSET, signPermit } = require('./helpers');
```
roxdanila, line 4: same as above

test/integration/setup.js
shark0der: The varibles were unused but I didn't want to remove the code because it might be needed later (some of the tests are a work in progress). Commented out to keep the code but remove the warning.

shark0der:
I'll summarize here a reply for the most comments:
- `daysToSeconds` is such a small function that it takes less characters to define it in the same file than having it in a library and importing it every time. For small functions I really think it's better to have them in-place, which will also provide the ability to quicker jump to the definition and check what it does - its going to be in the same file.
- I will not address the `ZERO_ADDRESS` in this PR because it's out of scope. I did address some of them only because the tests in that file already used ethers and oz was the last extraneous dep.
- The reason I've changed some of the test titles was because the line was too long and I found a way to shorten that. Skipping/unskipping tests is not in the scope of this PR.
- In some place double quotes are used instead of single quotes because the string contains a single quote already. This change was done automatically by prettier. `"singleQuote": true` is already in the prettier config, therefore this is the way to go.

roxdanila (approved):
<3

PR #369, opened 2022-09-08
Replace OZ base NFT with solmate in CoverNFT

### LINKED ISSUE #295: Cover: Use solmate for Cover NFT

kyledewy:
## Context
resolves #295 

This PR replaces the OZ 721 implementation with solmates 721 implementation. 

## Test plan
No further tests were added.

### DIFF
--- a/contracts/mocks/Claims/CLMockUnknownNFT.sol
+++ b/contracts/mocks/Claims/CLMockUnknownNFT.sol
@@ -13,8 +13,7 @@ contract CLMockUnknownNFT is ERC721 {
     _mint(to, tokenId);
   }
 
-  function tokenURI(uint id) public view override returns (string memory) {
-    id; 
+  function tokenURI(uint) public view override returns (string memory) {
     return "";
   }
 
--- a/contracts/mocks/Claims/CLMockUnknownNFT.sol
+++ b/contracts/mocks/Claims/CLMockUnknownNFT.sol
@@ -2,7 +2,7 @@
 
 pragma solidity ^0.8.9;
 
-import "@openzeppelin/contracts-v4/token/ERC721/ERC721.sol";
+import "solmate/src/tokens/ERC721.sol";
 
 
 contract CLMockUnknownNFT is ERC721 {
@@ -11,6 +11,11 @@ contract CLMockUnknownNFT is ERC721 {
 
   function mint(address to, uint tokenId) external {
     _mint(to, tokenId);
+  }
+
+  function tokenURI(uint id) public view override returns (string memory) {
+    id; 
+    return "";
   }
 
 }
--- a/contracts/mocks/Cover/CoverMockStakingPool.sol
+++ b/contracts/mocks/Cover/CoverMockStakingPool.sol
@@ -46,10 +46,6 @@ contract CoverMockStakingPool is IStakingPool, ERC721Mock {
     memberRoles = _memberRoles;
   }
 
-  function nameWithPoolId() public view returns (string memory) {
-    return string(abi.encodePacked(name, " ", Strings.toString(poolId)));
-  }
-
   function initialize(
     address _manager,
     bool _isPrivatePool,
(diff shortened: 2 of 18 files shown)


shark0der:
1. It seems that solmate was moved from [rari to transmissions11's account](https://github.com/Rari-Capital/solmate) and the repo is automatically redirected. The npm package is now [deprecated](https://www.npmjs.com/package/@rari-capital/solmate). Let's migrate to the new npm package `solmate`.
2. There are still mock contracts using the OZ version. Let's switch those to solmate.
3. Solmate implementation lacks the tokenURI implementation if I'm not mistaken, and that one should be generated using the base uri and the token id appended, let's implement that one as well.

roxdanila:
@kyledewy can you double check there is no .only left in tests? The coverage should be 70+% afaik

contracts/mocks/Cover/CoverMockStakingPool.sol
```diff
@@ -43,13 +41,13 @@ contract CoverMockStakingPool is IStakingPool, ERC721 {
...
     address _memberRoles
-  ) ERC721("Nexus Mutual Staking Pool", "NMSPT")
+  ) ERC721Mock("Nexus Mutual Staking Pool", "NMSPT")
   {
     memberRoles = _memberRoles;
   }
 
-  function name() public view override returns (string memory) {
-    return string(abi.encodePacked(super.name(), " ", Strings.toString(poolId)));
+  function nameWithPoolId() public view returns (string memory) {
```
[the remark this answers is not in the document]
shark0der: I think the `name()` function in this mock was only present because it wasn't implemented in the inherited contract at the time so it wasn't deployable. We probably need to delete this function from the mock, but search for usages first to make sure we don't actually use it.

contracts/modules/cover/Cover.sol
```diff
@@ -2,7 +2,7 @@
 
 pragma solidity ^0.8.9;
 
-import "@openzeppelin/contracts-v4/token/ERC721/ERC721.sol";
+import "solmate/src/tokens/ERC721.sol";
```
shark0der, line 5: I don't think ERC721 is used in this contract - remove this import

contracts/mocks/Claims/CLMockUnknownNFT.sol
```diff
@@ -13,4 +13,9 @@ contract CLMockUnknownNFT is ERC721 {
     _mint(to, tokenId);
   }
 
+  function tokenURI(uint id) public view override returns (string memory) {
+    id;
```
shark0der, line 17: Since this is a mock and we don't need var names explicitly here, we can remove the variable name from the argument list to silence the error:
`function tokenURI(uint id)` => `function tokenURI(uint)`

contracts/mocks/Cover/CoverMockStakingPool.sol
shark0der, line 49: Same as the before ^

contracts/modules/cover/CoverNFT.sol
```diff
@@ -38,7 +38,23 @@ contract CoverNFT is ERC721 {
...
+            _balanceOf[from]--;
+
+            _balanceOf[to]++;
+        }
+
+        _ownerOf[tokenId] = to;
+
+        delete getApproved[tokenId];
+
+        emit Transfer(from, to, tokenId);
```
shark0der, line 57: nitpick: can we remove some of the extra new lines? ex:

```solidity
        require(from == _ownerOf[tokenId], "WRONG_FROM");
        require(to != address(0), "INVALID_RECIPIENT");

        // Underflow of the sender's balance is impossible because we check for
        // ownership above and the recipient's balance can't realistically overflow.
        unchecked {
            _balanceOf[from]--;
            _balanceOf[to]++;
        }

        _ownerOf[tokenId] = to;
        delete getApproved[tokenId];

        emit Transfer(from, to, tokenId);
```

contracts/mocks/Cover/CoverMockStakingPool.sol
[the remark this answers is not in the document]
kyledewy: Ok. Looks like it's not being called.

Issue #364, opened 2022-09-07
Get segments method to retrieve all segments for a cover


mixplore:
We need a method that returns all the NFT segments corresponding to a cover.

PR #371, opened 2022-09-08
Feat: Add cover segments view

danoctavian:
## Context

Fixes Issue: https://github.com/NexusMutual/smart-contracts/issues/364

This is needed for displaying all cover segments in the UI.

 It's easier to add it in the CoverViewer.sol to not bloat Cover.sol size so a function is added there for returning a list of CoverSegment items.

## Test plan

Added test for CoverViewer

```npm run test-unit```

### DIFF
--- a/contracts/modules/cover/CoverViewer.sol
+++ b/contracts/modules/cover/CoverViewer.sol
@@ -111,6 +111,17 @@ contract CoverViewer {
     return coverViews;
   }
 
+  function getCoverSegments(uint coverId) external view returns (CoverSegment[] memory segments) {
+
+    ICover _cover = cover();
+    uint count = _cover.coverSegmentsCount(coverId);
+    segments = new CoverSegment[](count);
+
+    for (uint i = 0; i < count; i++) {
+      segments[i] = _cover.coverSegments(coverId, i);
+    }
+  }
+
   /* ========== COVER PRICING VIEWS ========== */
 
   function getPoolAllocationPriceParametersForProduct(
--- a/test/unit/CoverViewer/views.js
+++ b/test/unit/CoverViewer/views.js
@@ -1,15 +1,10 @@
 const { assert, expect } = require('chai');
 const { ethers } = require('hardhat');
-const { utils: { parseEther } } = ethers;
-
 const {
-  constants: { ZERO_ADDRESS },
-} = require('@openzeppelin/test-helpers');
-const { BigNumber } = require('ethers');
-const { bnEqual } = require('../utils').helpers;
+  utils: { parseEther },
+} = ethers;
 
 describe('views', function () {
-
   it('getCoverSegments returns segments', async function () {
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: contracts/modules/cover/CoverViewer.sol [hunk 1]
```diff
@@ -111,6 +111,17 @@ contract CoverViewer {
     return coverViews;
   }
 
+  function getCoverSegments(uint coverId) external view returns (CoverSegment[] memory segments) {
```
roxdanila, line 114: @mixplore is this enough for your usecase? or should we implement this to take as input an array of cover IDs, and return a mapping coverID -> segments?

test/unit/CoverViewer/views.js
```diff
@@ -0,0 +1,34 @@
+const { assert, expect } = require('chai');
+const { ethers } = require('hardhat');
+const { utils: { parseEther } } = ethers;
+
+const {
+  constants: { ZERO_ADDRESS },
+} = require('@openzeppelin/test-helpers');
```
roxdanila, line 7: let's not use this anymore, replace with the equivalent ethers constant

test/unit/CoverViewer/views.js
```diff
@@ -0,0 +1,34 @@
...
+    };
+
+    const coverId = 0;
+
+    await cover.addSegments(coverId, [coverSegment1]);
+
+    const segments = await coverViewer.getCoverSegments(coverId);
+
+    assert.equal(segments.length, 1);
+    assert(segments[0].amount.toString(), coverSegment1.amount.toString());
```
roxdanila, line 32: assert.equal?

roxdanila (changes requested):

contracts/modules/cover/CoverViewer.sol
mixplore: 🤔  Now that you're mentioning, yeah, it would probably be better to be able to get the segments for multiple covers. I think we'll need at least the active segment of the cover (with period and amount details) in order to display the data in `My covers` table. The way I see it coming from an API is described in this ticket [#4](https://github.com/NexusMutual/cover-router/issues/4) if that helps.

roxdanila:
Also, please follow the title convention: Github case, prefixed with Feat:

contracts/modules/cover/CoverViewer.sol
danoctavian: understood, DMing about this

roxdanila: after offline discussions, we're keeping this implementation as is.

roxdanila (commented):
@danoctavian please address the comments in the test file + rebase and this is good to go.

nitpick: please update the title of the PR to follow the convention `Feat: Add cover segments view'

roxdanila (approved):

PR #396, opened 2022-09-27
Feature/edit existing product cover

danoctavian:
## Context

Fixes: https://github.com/NexusMutual/smart-contracts/issues/286

Adds a function for editing fields of an existing product.

Add a solidity function editProducts()

## Test plan

Added tests to `test/unit/Cover/editProducts.js`

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/modules/cover/Cover.sol
```diff
@@ -647,6 +647,16 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     }
   }
 
+  function editProducts(
```
danoctavian, line 650: caveats 

Still have a separate call to change the IPFS Metadata hashes.



Reasoning: i made them separate because if i'm not touching the metadata value I don't want an event emitted.

Caveat 2


All fields are editable this way. In theory some fields may not be editable, like initialPrice, but this implementation is easier to make.

Otherwise I'd need to make an ProductEditChanges struct. I can see the idea of it but given this is AB controlled i don't see a big benefit.

contracts/modules/cover/Cover.sol
```diff
@@ -637,12 +637,14 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
 
-  function editProductsIpfsMetadata(
+  function editProducts(
     uint[] calldata productIds,
+    Product[] calldata newProducts,
     string[] calldata ipfsMetadata
   ) external override onlyAdvisoryBoard {
 
     for (uint i = 0; i < productIds.length; i++) {
+      _products[productIds[i]] = newProducts[i];
```
shark0der, line 647: The `Product` struct contains the following fields:
- `uint16 productType`
- `address productAddress`
- `uint32 coverAssets`
- `uint16 initialPriceRatio`
- `uint16 capacityReductionRatio`

As a result:
1. I don't find it OK to allow to change `productType` and `productAsset`.
2. There's no validation whatsoever on the correctness of the other parameters.

shark0der (changes requested):
See comment above

kyledewy (approved):

shark0der (approved):

PR #397, opened 2022-09-28
Feature/edit product types

danoctavian:
## Context

Fixed issue: https://github.com/NexusMutual/smart-contracts/issues/288

Add an editProductTypes call.

Add editProductTypes function.

## Test plan

test/unit/editProductTypes.js

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -638,19 +638,19 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     uint initialProuctTypesCount = _productTypes.length;
     for (uint i = 0; i < newProductTypes.length; i++) {
       _productTypes.push(newProductTypes[i]);
-      emit ProductTypeUpserted(initialProuctTypesCount + i, ipfsMetadata[i]);
+      emit ProductTypeSet(initialProuctTypesCount + i, ipfsMetadata[i]);
     }
   }
   
   function editProductTypes(
     uint[] calldata productTypeIds,
-    ProductType[] calldata newProductTypes,
+    uint16[] calldata gracePeriodsInDays,
     string[] calldata ipfsMetadata
   ) external override onlyAdvisoryBoard {
 
     for (uint i = 0; i < productTypeIds.length; i++) {
-      _productTypes[productTypeIds[i]] = newProductTypes[i];
-      emit ProductTypeUpserted(productTypeIds[i], ipfsMetadata[i]);
+      _productTypes[productTypeIds[i]].gracePeriodInDays = gracePeriodsInDays[i];
+      emit ProductTypeSet(productTypeIds[i], ipfsMetadata[i]);
     }
   }
 
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -641,6 +641,18 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
       emit ProductTypeUpserted(initialProuctTypesCount + i, ipfsMetadata[i]);
     }
   }
+  
+  function editProductTypes(
+    uint[] calldata productTypeIds,
+    ProductType[] calldata newProductTypes,
+    string[] calldata ipfsMetadata
+  ) external override onlyAdvisoryBoard {
+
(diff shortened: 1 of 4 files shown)


contracts/modules/cover/Cover.sol
```diff
@@ -637,6 +637,18 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
   }
 
+  function editProductTypes(
+    uint[] calldata productTypeIds,
+    ProductType[] calldata newProductTypes,
+    string[] calldata ipfsMetadata
+  ) external override onlyAdvisoryBoard {
+    
+    for (uint i = 0; i < productTypeIds.length; i++) {
+      _productTypes[productTypeIds[i]] = newProductTypes[i];
```
shark0der, line 647: `ProductType` contains `claimMethod` and `gracePeriodInDays`. I don't think it's okay to change the claim method, hence instead of passing a `ProductType[]` we could only pass a `uint[]` with new grace periods.

contracts/modules/cover/Cover.sol
```diff
@@ -637,6 +637,18 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+      emit ProductTypeUpserted(productTypeIds[i], ipfsMetadata[i]);
```
shark0der, line 648: Can we please please please change the "upsert" part? It's a term that's highly related to databases and it's not the case to use it here. A couple of ideas: `ProductTypeMetadataUpdated`, `ProductTypeMetadataSet`, `ProductTypeSet`, `ProductTypeChanged`

test/unit/Cover/editProductTypes.js
```diff
@@ -0,0 +1,41 @@
...
+    };
+
+    await cover
+      .connect(accounts.advisoryBoardMembers[0])
+      .editProductTypes([productTypeId], [productTypeData], ['my ipfs hash']);
+
+    const storedProduct = await cover.productTypes(productTypeId);
+
+    expect(storedProduct.claimMethod).to.be.equal(productTypeData.claimMethod);
+    expect(storedProduct.gracePeriodInDays).to.be.equal(productTypeData.gracePeriodInDays);
```
shark0der, line 21: - Add test for ipfs hash
- Add test for emitted event

test/unit/Cover/editProductTypes.js
```diff
@@ -0,0 +1,41 @@
...
+    const productTypeData = {
+      claimMethod: 2,
+      gracePeriodInDays: 45,
+    };
+
+    await cover
+      .connect(accounts.advisoryBoardMembers[0])
+      .editProductTypes([productTypeId], [productTypeData], ['my ipfs hash']);
+
+    const storedProduct = await cover.productTypes(productTypeId);
```
shark0der, line 18: `storedProductType`*

shark0der (changes requested):
See comments above

contracts/modules/cover/Cover.sol
danoctavian: Yeah that's fine i think it was chosen because you can use it for both adding and updating but i agree it's confusing

shark0der (approved):

test/unit/Cover/editProductTypes.js
```diff
@@ -0,0 +1,37 @@
+const { expect } = require('chai');
+
+describe('editProductTypes', function () {
+  it('should edit existing product', async function () {
+    const { cover, accounts } = this;
+
+    const productTypeId = 0;
+
+    const gracePeriodsInDays = [45];
```
shark0der, line 9: Passed parameter becomes a 2-dimensional array.

shark0der (changes requested):
Investigate.

test/unit/Cover/editProductTypes.js
danoctavian: Seems to be something to do with how arrays of uints are serialized.

Adding the array nesting or removing it has the same effect.

```cover
        .connect(accounts.advisoryBoardMembers[0])
        .editProductTypes([[productTypeId]], gracePeriodsInDays, [ipfsHash]),```
        
        
       This call where [[productTypeId]] is nested in an array also works correctly.

shark0der (approved):

PR #408, opened 2022-10-04
Chore/Rename bucketStartTime to bucketEndTime

### LINKED ISSUE #388: Staking Code Review

kyledewy:
## Context
Closes #388  To align the naming between start/end, when expiring buckets and tranches.

This PR renames bucketStartTime to bucketEndTime to match the tranche expiry.

## Test plan
Just a semantic change, no tests were added.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/staking/StakingPool.sol
+++ b/contracts/modules/staking/StakingPool.sol
@@ -269,14 +269,14 @@ contract StakingPool is IStakingPool, ERC721 {
         // each bucket contains a reward reduction - we subtract it when the bucket *starts*!
 
         ++_firstActiveBucketId;
-        uint bucketStartTime = _firstActiveBucketId * BUCKET_DURATION;
-        uint elapsed = bucketStartTime - _lastAccNxmUpdate;
+        uint bucketEndTime = _firstActiveBucketId * BUCKET_DURATION;
+        uint elapsed = bucketEndTime - _lastAccNxmUpdate;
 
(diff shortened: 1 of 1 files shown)


roxdanila (approved):

shark0der (changes requested):
The first bucket and the first tranche (ie bucket id and tranche id = 0) start at zero unix time.

When you deposit to the tranche zero, your deposit is locked while the tranche is active. The tranche zero starts at `trancheStartTime = trancheId * TRANCHE_DURATION = 0` and expires at `trancheEndTime = (trancheId + 1) * TRANCHE_DURATION = TRANCHE_DURATION`. That's why we're using `trancheEndTime` in our code, not start time.

For the buckets, let's say that current unix time = 0. Say someone buys a cover for a period equal to exactly one bucket. We increase the current reward per second and write down that we should decrease the reward per second at a particular time in the future. That particular time is when the bucket id = 1 starts. For this reason we're using `bucketStartTime = bucketId * BUCKET_DURATION`.

So technically just renaming `bucketStartTime` to `bucketEndTime` is incorrect, because that's not what it represents. If we want to use `bucketEndTime`, we will need to reflect this change in the code as well, and use `bucketEndTime = bucketId * BUCKET_DURATION` where `bucketId` will be the expiring bucket id BUT ALSO on each cover buy store the reward amount to be reduced in the last bucket that the cover will be active, instead of the first bucket where the cover is expired.

PR #436, opened 2022-10-21
Use v1 cover sumAssured as it is (already in wei)

mixplore:
## Context

The cover migrate transaction for a v1 cover got reverted with reason string `SafeCast: value doesn't fit in 96 bits` on the amount param.

I noticed that the `sumAssured` value of a v1 cover (at runtime) was already in wei (had 18 0s) so there was no need to convert it further. The change in this PR removes that conversion.

## Test plan

I've only tested it locally, using the deploy script and calling the migrate contract method from the FE app. With this change, the v1 cover gets migrated successfully.

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback


shark0der:
N/A

Issue #437, opened 2022-10-21
Get new v2 cover id after v1 cover migration


mixplore:
### Context
When submitting a claim for an old v1 cover, we first need to migrate the cover to a v2 one and then call the submit claim method. 
After the cover migration, the new cover has a new `coverId`. 
We need that id returned from the migration method so we can use it further as param for submit claim method.

danoctavian:
We need an event here. for cover migration runs.

It is indeed missing now.

danoctavian:
we'll add a CoverMigrated event for simplicity.

PR #444, opened 2022-10-25
Fix: Add event for migrateCover

danoctavian:
## Context
Solves issue: https://github.com/NexusMutual/smart-contracts/issues/437

There's no easy way to know what's the new cover id of the cover you just migrated. This is necessary in the UI to know what use as a parameter for claim submission.

This PR adds an event for each migrated cover.

1 new event type CoverMigrated and an emit call.

## Test plan

We do not have tests for `migrateCovers`. We can add a verification for the event when we get there.

We can add the check when those tests are written. 

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -128,9 +128,13 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
   ///
   /// @param coverIds    Legacy (V1) cover identifiers
   /// @param toNewOwner  The address for which the V2 cover NFT is minted
-  function migrateCovers(uint[] calldata coverIds, address toNewOwner) external override {
+  function migrateCovers(
+    uint[] calldata coverIds,
+    address toNewOwner
+  ) external override returns (uint[] memory newCoverIds) {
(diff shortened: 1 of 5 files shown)


contracts/modules/cover/Cover.sol
```diff
@@ -174,6 +174,8 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
       _coverData,
       _coverSegments
     );
+
+    emit CoverMigrated(coverId, fromOwner, toNewOwner, _coverData.length - 1);
```
danoctavian, line 178: note i'm doing an extra read here.

Reason: returning the cover id from the Cover `CoverUtilsLib.migrateCoverFromOwner` causes stack too deep.

gas_cost = 100 (warm access)

mixplore (approved):
Thank you!!  🚀

roxdanila (changes requested):
@danoctavian return the coverId as well, so another contract (i.e. Armor) can use it as well

shark0der (approved):
:+1:

Issue #268, opened 2022-08-30
V2 Cover unit tests: buyCover


roxdanila:
`buyCover`

- [x] should purchase new cover using 1 staking pool: 1000 ETH for 30 days, pay in ETH
- [x] should purchase new cover using 2 staking pools: 1000 ETH for 30 days, pay in ETH
- [x] should purchase new cover using NXM with commission: 1000 ETH for 30 days, pay in NXM
- [x] should purchase new cover using DAI with commission: 1000 DAI for 30 days, pay in DAI
- [x] should purchase new cover using USDC with commission: 1M USDC for 30 days, pay in USDC
- [x] should revert for unavailable product
- [x] should revert for unsupported payout asset
- [x] should revert for period shorter than the minimum (28 days)
- [x] should revert for period longer than the maximum (364 days)
- [x] should revert for commission rate too high (25%)
- [x] should revert when cover amount is 0

- [x] Should revert if system is paused
- [x] Should revert if caller is not member
- [x] Should revert if reentrant (Is reentrant protection needed?)
- [x] Should revert if owner is address zero
- [x] Should revert if productId is not valid
- [ ] Should revert if product is not initialized
- [x] Should revert if payout asset is not supported by the product
- [x] Should revert if amount is 0
- [x] Should revert if period is not within allowed bounds 
- [x] Should revert if commission ratio is bigger than max value 
- [x] Should revert if paymentAsset is not supported
- [x] Should revert if paymentAsset is deprecated 
- [x] Should revert if calculated premium is bigger than maxPremiumInAsset
- [x] Should revert if empty array of allocationRequests?
- [x] Should revert if pools out of capacity
- [x] Should retrieve payment in ERC20 paymentAsset from caller
- [x] Should allow to pay with NXM
- [x] Should allow to pay with ETH
- [x] Should allow to pay with DAI
- [x] Should pay correct commission to commissionDestination
- [x] Should store cover data
- [x] Should mint NFT to owner
- [ ] Should emit CoverBought event
- [x] Should work if owner is not a member
- [ ] Should correctly calculate premium
- [x] Should correctly allocate capacity to pools
- [x] Should allow to buy against multiple staking pools (Is there a max amount of pools that should be considered?)
- [x] Should correctly mint rewards to pools
- [x] Should correctly create segments

roxdanila:
- _Should revert if product is not initialized_ - will be treated as part of issue #412 
- _ Should emit CoverBought event_: TODO 
[...]

PR #449, opened 2022-10-26
Test/ Add buyCover revert unit tests

patitonar:
## Context

Partial implementation of #268

Adds new revert unit tests for `buyCover`. Implemented some of the listed test cases, especially the reverts one.
 
I couldn't find a way to test `Cover: Product not initialized`, it seems that when a product is listed it also initializes the parameters so I couldn't reproduce a state of listed but with `initialPriceRatio = 0` as the code is checking.

Also, I removed the usage of `bnEqual` in the `buyCover` tests as it triggers a warning `Warning: Use Hardhat Chai Matchers instead of bnEqual: https://hardhat.org/hardhat-chai-matchers/docs/overview#big-numbers`

## Test plan

New tests were added in `test/unit/Cover/buyCover.js`

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/test/unit/Cover/buyCover.js
+++ b/test/unit/Cover/buyCover.js
@@ -5,43 +5,53 @@ const { createStakingPool, assertCoverFields } = require('./helpers');
 
 const { parseEther } = ethers.utils;
 const { AddressZero } = ethers.constants;
-const { BigNumber } = ethers;
 const gracePeriodInDays = 120;
 
+const buyCoverFixture = {
+  productId: 0,
+  coverAsset: 0, // ETH
+  coverId: 0,
+  period: 3600 * 24 * 30, // 30 days
+  amount: parseEther('1000'),
+  targetPriceRatio: 260,
+  priceDenominator: 10000,
+  activeCover: parseEther('8000'),
+  capacity: parseEther('10000'),
+  capacityFactor: 10000,
+  expectedPremium: parseEther('1000').mul(260).div(10000), // amount * targetPriceRatio / priceDenominator
+};
+
 describe('buyCover', function () {
-  it('should purchase new cover using 1 staking pool', async function () {
+  beforeEach(async function () {
     const { cover } = this;
 
     const {
       governanceContracts: [gv1],
-      members: [coverBuyer, stakingPoolManager],
+      members: [stakingPoolManager],
     } = this.accounts;
 
-    const productId = 0;
-    const coverAsset = 0; // ETH
-    const period = 3600 * 24 * 364; // 30 days
-
-    const amount = parseEther('1000');
-    const targetPriceRatio = '260';
-    const priceDenominator = '10000';
-    const activeCover = parseEther('8000');
-    const capacity = parseEther('10000');
-    const capacityFactor = '10000';
-
-    await cover.connect(gv1).updateUintParameters([0], [capacityFactor]);
+    await cover.connect(gv1).updateUintParameters([0], [buyCoverFixture.capacityFactor]);
 
     await createStakingPool(
       cover,
-      productId,
-      capacity,
-      targetPriceRatio,
-      activeCover,
+      buyCoverFixture.productId,
+      buyCoverFixture.capacity,
+      buyCoverFixture.targetPriceRatio,
+      buyCoverFixture.activeCover,
       stakingPoolManager,
       stakingPoolManager,
-      targetPriceRatio,
+      buyCoverFixture.targetPriceRatio,
     );
+  });
 
-    const expectedPremium = amount.mul(targetPriceRatio).div(priceDenominator);
+  it('should purchase new cover using 1 staking pool', async function () {
+    const { cover } = this;
+
+    const {
+      members: [coverBuyer],
+    } = this.accounts;
+
+    const { amount, targetPriceRatio, productId, coverAsset, period, expectedPremium, coverId } = buyCoverFixture;
 
     const tx = await cover.connect(coverBuyer).buyCover(
       {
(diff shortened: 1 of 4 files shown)


test/unit/Cover/buyCover.js
```diff
@@ -712,4 +711,372 @@ describe('buyCover', function () {
...
+      governanceContracts: [gv1],
+      members: [coverBuyer, stakingPoolManager],
+    } = this.accounts;
+
+    const productId = 0;
+    const coverAsset = 0; // ETH
+    const period = 3600 * 24 * 364; // 30 days
+
+    const amount = parseEther('1000');
+    const targetPriceRatio = '260';
```
danoctavian, line 728: just checked and this pool fixture and the createStakingPool call ar all the same for all the tests currently.

It makes sense to have  a `beforeEach` section which is just creating this default pool to avoid this code duplication.

We will add tests for buying across multiple pools. in those tests we can add the extra pool there but we always need at least this 1 pool.

initializing all these values in a fixture to be used in the rest of the test code would strip this section a code from each test.

danoctavian: So let's do a beforeEach with all this code including the `createStakingPool` 

create a fixture for the values for the pool so that values are not initialized in every test

test/unit/Cover/buyCover.js
```diff
@@ -2,7 +2,6 @@ const { expect } = require('chai');
 const { ethers } = require('hardhat');
```
danoctavian, line 2: i'm not seeing a test for 

`Should revert if product is not initialized`

should revert with `Cover: Product not initialized`

danoctavian: I saw your comment.

The way to test is to refer to some product id that's not used.

Look at how setup adds a product which will have index 0.

https://github.com/NexusMutual/smart-contracts/blob/568e2f6525147789bb54924afa976dc69de899b2/test/unit/Cover/setup.js#L188

Refer to the one with productId: 1 to make it crash

danoctavian: https://github.com/NexusMutual/smart-contracts/blob/568e2f6525147789bb54924afa976dc69de899b2/test/unit/Cover/setup.js#L188

To make it revert with that message add a product like in the setup, which will have productId = 1

and set initialPriceRatio: 0

test/unit/Cover/buyCover.js
```diff
@@ -314,10 +313,10 @@ describe('buyCover', function () {
     const commissionDaiBalanceAfter = await dai.balanceOf(commissionReceiver.address);
 
     const difference = daiBalanceBefore.sub(daiBalanceAfter);
-    bnEqual(difference, expectedPremium);
+    expect(difference).to.be.equal(expectedPremium);
```
danoctavian, line 316: thanks for fixing these

test/unit/Cover/helpers.js
```diff
@@ -50,11 +49,11 @@ async function assertCoverFields(
   const segment = await cover.coverSegments(coverId, segmentId);
   assert.equal(storedCoverData.productId, productId);
   assert.equal(storedCoverData.coverAsset, coverAsset);
-  bnEqual(storedCoverData.amountPaidOut, amountPaidOut);
+  expect(storedCoverData.amountPaidOut.toString()).to.be.equal(amountPaidOut);
```
danoctavian, line 52: you shouldn't need that toString() here, i expect

correct me if i'm wrong

test/unit/Cover/helpers.js
```diff
@@ -50,11 +49,11 @@ async function assertCoverFields(
...
   assert.equal(segment.gracePeriodInDays, gracePeriodInDays);
   assert.equal(segment.period, period);
   assert.equal(segment.amount.toString(), amount.toString());
-  bnEqual(segment.priceRatio, targetPriceRatio);
+  expect(segment.priceRatio.toString()).to.be.equal(targetPriceRatio);
```
danoctavian, line 56: same here

danoctavian (changes requested):
a few changes required.


Also expecting the rest of the tests are coming in a different PR?

Or what's the plan for those unticked ones not covered?

test/unit/Cover/buyCover.js
patitonar: Thanks for the help!

I tried doing that but it reverts with `Cover: initialPriceRatio < GLOBAL_MIN_PRICE_RATIO`  which is set as a constant `GLOBAL_MIN_PRICE_RATIO = 100`.

Unless I'm missing something, it looks like products could not be in an uninitialized state, so in that case, Is that validation needed? I think @roxdanila mentioned something about new functionality to deprecate products in which case once introduced it would make sense to validate that the product was not deprecated

danoctavian: Ah i see you're right. That validation was added very recently and that require is not useful anymore.

Thanks for clarifying.

test/unit/Cover/buyCover.js
patitonar: Thanks for the suggestion! I refactored the code, now the tests code is cleaner

test/unit/Cover/helpers.js
patitonar: I was throwing an error, I updated the code so `toString()` is not longer needed

patitonar:
> Also expecting the rest of the tests are coming in a different PR?

Yes, the rest of the tests will be added in another PR

danoctavian (approved):
LGTM

PR #458, opened 2022-10-28
Fix pool todos

danoctavian:
## Context
Fixes https://github.com/NexusMutual/smart-contracts/issues/343

Add a check for orderInProgress() when upgrading SwapOperator when it's being replaced.

Remove getWei function.

Move `minPoolEth` to SwapOperator where it's the only place where it's used.

Pack   `uint96 public swapValue;` with `deprecatedCoverAssetsBitmap`.

## Test plan

Change failing tests so they pass given the new updates.

No new tests added.
## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/capital/Pool.sol
+++ b/contracts/modules/capital/Pool.sol
@@ -94,6 +94,7 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
 
     // [todo] After this contract is deployed it might be worth modifying upgradeCapitalPool to
     // copy the assets on future upgrades instead of having them hardcoded in the constructor.
+    // issue: https://github.com/NexusMutual/smart-contracts/issues/473
 
     // The order of coverAssets should never change between updates. Do not remove the following
     // lines!
--- a/contracts/modules/capital/Pool.sol
+++ b/contracts/modules/capital/Pool.sol
@@ -32,7 +32,6 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
   IMCR public mcr;
 
   // parameters
-  address public swapController;
   IPriceFeedOracle public override priceFeedOracle;
   address public swapOperator;
 
@@ -92,6 +91,9 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
     master = INXMMaster(_master);
     priceFeedOracle = IPriceFeedOracle(_priceOracle);
     swapOperator = _swapOperator;
+
+    // [todo] After this contract is deployed it might be worth modifying upgradeCapitalPool to
+    // copy the assets on future upgrades instead of having them hardcoded in the constructor.
 
     // The order of coverAssets should never change between updates. Do not remove the following
     // lines!
--- a/contracts/modules/capital/Pool.sol
+++ b/contracts/modules/capital/Pool.sol
@@ -462,17 +462,6 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
     return true;
   }
 
-
-  /// @dev DEPRECATED, use calculateNXMForEth function instead! Returns the amount of wei a seller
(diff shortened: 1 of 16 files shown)


contracts/modules/capital/Pool.sol
```diff
@@ -461,17 +462,6 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
   }
 
-
-  /// @dev DEPRECATED, use calculateNXMForEth function instead! Returns the amount of wei a seller
-  /// will get for selling NXM
-  ///
-  /// @param amount     Amount of NXM to sell
-  /// @return weiToPay  Amount of wei the seller will get
-  /// [todo] Is it safe to remove this?
-  function getWei(uint amount) external view returns (uint weiToPay) {
```
roxdanila, line 471: I think this was used by Armor / some distributor contract, and that's why we kept it. Pls double check before removing it.

danoctavian: It's used by the yInsure contract

https://github.com/EaseDeFi/arNFT/blob/master/contracts/externals/yInsure.sol#L580

https://etherscan.io/address/0x181aea6936b407514ebfc0754a37704eb8d98f91#code

The issue is even though in theory removing this would brick the sellNXMTokens function, in practice the function is not callable anyway because MCR is at 100% so you can't sell.

The way, to pull out the NXM out of that contract is to run nxmTokenApprove
And transfer them out by the owner (https://etherscan.io/address/0x2D407dDb06311396fE14D4b49da5F0471447d45C) who is Andre i believe.

contracts/modules/capital/Pool.sol
```diff
@@ -93,9 +92,6 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
     priceFeedOracle = IPriceFeedOracle(_priceOracle);
     swapOperator = _swapOperator;
 
-    // [todo] After this contract is deployed it might be worth modifying upgradeCapitalPool to
```
danoctavian, line 96: I'll create an issue to track this.

shark0der: Let's keep the todo but add the issue id.

contracts/modules/capital/SwapOperator.sol
```diff
@@ -522,4 +525,8 @@ contract SwapOperator {
 
     asset.transfer(address(pool), balance);
   }
+
+  function orderInProgress() public returns (bool) {
```
shark0der, line 529: This should be view?

contracts/modules/capital/Pool.sol
```diff
@@ -699,11 +684,6 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
   }
 
   function updateUintParameters(bytes8 code, uint value) external onlyGovernance {
-    if (code == "MIN_ETH") {
-      minPoolEth = value;
-      return;
-    }
-
```
shark0der, line 706: Remove the function entirely?

contracts/modules/capital/Pool.sol
```diff
@@ -35,8 +32,6 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
 
   // parameters
   address public swapController;
```
shark0der, line 34: `swapController` is unused?

shark0der (changes requested):
Looking good! Just a few minor comments

contracts/modules/capital/Pool.sol
danoctavian: We have a Governance proposal for this i would keep it in case a parameter comes up

shark0der (approved):

PR #461, opened 2022-10-31
Feature:  Merge cover product management function into 2 functions: setProducts() and setProductTypes()

### LINKED ISSUE #412: Cover Product: Merge product management function into 2 functions: `setProducts()` and `setProductTypes()`

kyledewy:
## Context

Close #412 

- Add isDeprecated flag to products struct
- Replace add/editProduct function with a single `setProducts` function
      - `productId == uint256.max` will create a new product, otherwise can pass the productId to be edited
- Rename staking.ProductParams to staking.StakedProductParams for consistency with setStakedProducts()
- Explicitly prevent all cover buys and cover edits for deprecated products

## Test plan

New tests were added at `test/unit/Cover/setProducts.js` and `test/unit/Cover/setProductTypes.js`

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/modules/cover/Cover.sol
```diff
@@ -638,64 +639,44 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+        _products.push(param.product);
+      } else {
+        // existing product
+        require(param.productId < _products.length, "Cover: Product doesnt exist. Set id to uint256.max to add it");
+        Product storage newProductValue = _products[param.productId];
+        newProductValue.isDeprecated = param.product.isDeprecated;
+        newProductValue.coverAssets = param.product.coverAssets;
+        newProductValue.initialPriceRatio = param.product.initialPriceRatio;
+        newProductValue.capacityReductionRatio = param.product.capacityReductionRatio;
+        emit ProductSet(param.productId, param.ipfsMetadata);
```
shark0der, line 661: If the product update did not intend to edit the metadata - we'll end up with a bunch of events that did not in fact change the ipfs metadata. This makes it harder for the UI to figure out when the cover metadata was in fact changed as it would have a bunch of redundant events, and also error prone because it's harder to fetch events compared to reading a product from the state at the lastest block. A transient error in event fetching could accidentally downgrade the ipfs metadata.

How about allowing the ipfs metadata to be empty and use it as a signal not to update it?

```solidity
if (param.ipfsMetadata.length > 0) {
  emit ProductSet(param.productId, param.ipfsMetadata);
}
```

contracts/modules/cover/Cover.sol
```diff
@@ -638,64 +639,44 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-      _products[productIds[i]] = newProductValue;
-      emit ProductSet(productIds[i], ipfsMetadata[i]);
+      // New product has id == uint256.max
+      if (param.productTypeId == type(uint256).max) {
+        emit ProductTypeSet(_productTypes.length, param.ipfsMetadata);
+        _productTypes.push(param.productType);
+      } else {
+        require(param.productTypeId < _productTypes.length, "Cover: ProductType doesnt exist. Set id to uint256.max to add it");
+        _productTypes[param.productTypeId].gracePeriodInDays = param.productType.gracePeriodInDays;
+        emit ProductTypeSet(param.productTypeId, param.ipfsMetadata);
```
shark0der, line 678: Same comments as for the products: add `continue` to drop an indent level and allow empty metadata

shark0der (changes requested):
Looking good overall, just a few comments!

PR #479, opened 2022-11-07
Feature: Allow managers to lower targetWeights even if effectiveWeight is exceeded

### LINKED ISSUE #454: Staking: Add public function to recalculate effectiveWeights and allow for lower of targetWeight regardless of current weight

kyledewy:
## Context

Closes #454  

This PR changes the `stakingPool.setProducts()` function so the target weight of a product can be lowered, even if the pool has exceeded it's max effective weight. 

A function is also added, so that any user can recalculate the effective weights of the pool.

## Test plan
There isn't much to test until the burn functionality is implemented. Will add more on that PR.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/staking/StakingPool.sol
+++ b/contracts/modules/staking/StakingPool.sol
@@ -1315,8 +1315,8 @@ contract StakingPool is IStakingPool, ERC721 {
   function recalculateEffectiveWeights(uint[] calldata productIds) external {
     (
     uint globalCapacityRatio,
-    uint globalMinPriceRatio,
-    uint[] memory initialPriceRatios,
+    /* globalMinPriceRatio */,
+    /* initialPriceRatios */,
     uint[] memory capacityReductionRatios
     ) = ICover(coverContract).getPriceAndCapacityRatios(productIds);
 
--- a/contracts/modules/staking/StakingPool.sol
+++ b/contracts/modules/staking/StakingPool.sol
@@ -1645,7 +1645,7 @@ function setProducts(StakedProductParam[] memory params) external onlyManager {
       : 0;
 
     effectiveWeight = actualWeight > type(uint16).max
-      ? uint16(WEIGHT_DENOMINATOR)
-      : (Math.max(targetWeight, actualWeight)).toUint16();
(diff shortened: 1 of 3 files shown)


contracts/modules/staking/StakingPool.sol
```diff
@@ -1250,9 +1250,36 @@ contract StakingPool is IStakingPool, ERC721 {
...
+    ) = ICover(coverContract).getPriceAndCapacityRatios(productIds);
+
+    uint _totalEffectiveWeight = totalEffectiveWeight;
+
+    for (uint i = 0; i < productIds.length; i++) {
+      uint productId = productIds[i];
+      StakedProduct memory _product = products[productId];
+
+      uint8 previousEffectiveWeight = _product.lastEffectiveWeight;
+      _product.lastEffectiveWeight = _getEffectiveWeight(
```
shark0der, line 1268: As discussed, `lastEffectiveWeight` may overflow, we need a larger datatype + a check for overflows (because it may happen anyway). When an overflow is detected the type's max value instead.

contracts/modules/staking/StakingPool.sol
```diff
@@ -1309,7 +1342,9 @@ contract StakingPool is IStakingPool, ERC721 {
       products[_param.productId] = _product;
     }
 
-    require(_totalEffectiveWeight <= MAX_TOTAL_WEIGHT, "StakingPool: Total max effective weight exceeded");
+    if (_totalEffectiveWeight > MAX_TOTAL_WEIGHT) {
+      require(!targetWeightIncreased, "StakingPool: Total max effective weight exceeded");
+    }
```
shark0der, line 1347: I'd suggest swapping the conditions for `require` and `if` as it feels more natural/expressive:
```
    if (targetWeightIncreased) {
      require(_totalEffectiveWeight <= MAX_TOTAL_WEIGHT, "StakingPool: Total max effective weight exceeded");
    }
```

shark0der (approved):
lessgoooooo

Issue #386, opened 2022-09-23
Prevent cover buyers to edit covers to switch to another syndicate in disfavour of stakers


roxdanila:

roxdanila:
- [ ] Merge buyCover and editCover into 1 function
- [ ] Pass over gracePeriod when calling deallocate & allocate
- [ ] Deallocate
- [ ] Calculate refund
- [ ] Allocate

PR #480, opened 2022-11-08
Feature: merge buy & edit cover in a single function and remove refund logic from Cover

shark0der:
## Context

Partial implementation of #386.

- Given the logic overlap between `buyCover` and `editCover` the PR merges them in a single function
- Removes the refund logic from the Cover contract in order to avoid cross-pool refund (buy on staking pool X, move cover to Y hence switching the rewards from one pool to another and "rugging" the stakers)
- Moves the following items from the `CoverUtilsLib` to `Cover` contract:
  - payment handling
  - stake burn function
  - stakingPool function

## Test plan

All `buyCover` tests failed and need fixing (TBD).

## Checklist

## Review

* :rocket: = Feedback


contracts/modules/cover/Cover.sol
```diff
@@ -186,343 +185,185 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-    require(
-      isAssetSupported(_getSupportedCoverAssets(deprecatedCoverAssetsBitmap, product.coverAssets), buyCoverParams.coverAsset),
-      "Cover: Payout asset is not supported"
-    );
-    require(
-      !_isCoverAssetDeprecated(deprecatedCoverAssetsBitmap, buyCoverParams.paymentAsset),
-      "Cover: payment asset deprecated"
-    );
+    {
+      require(_products.length > params.productId, "Cover: Product not found");
```
danoctavian, line 224: What's the reason for not having all these validations at he beggining to reject the tx faster if something is wrong?

contracts/modules/cover/Cover.sol
```diff
@@ -186,343 +185,185 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
   function buyCover(
     BuyCoverParams memory params,
     PoolAllocationRequest[] memory allocationRequests
```
danoctavian, line 187: would rename this to poolAllocationRequests

The fact that there is another type called AllocationRequest makes it hard to disambiguate between what's coming from that array and what's being created inside the pool

contracts/modules/staking/StakingPool.sol
```diff
@@ -663,32 +662,42 @@ contract StakingPool is IStakingPool, ERC721 {
...
       rewardBuckets[expireAtBucket].rewardPerSecondCut += _rewardPerSecond;
 
-      // scale back from 2 to 18 decimals
-      allocatedCoverAmount *= NXM_PER_ALLOCATION_UNIT;
-
-      // premium and rewards already have 18 decimals
-      return (allocatedCoverAmount, premium, rewards);
+      // apply the global rewards ratio and the total Rewards in NXM
+      // TODO: implement me
+      // tokenController().mintStakingPoolNXMRewards(rewardsInNXM, allocationRequests[i].poolId);
```
danoctavian, line 676: what's the reason for not implementing this in this PR?

contracts/modules/cover/Cover.sol
shark0der: Moved up

contracts/modules/staking/StakingPool.sol
shark0der: I'm actually intending to move this call to the StakingPool itself. There're too many items going back and forth. It will be slightly more expensive but the amount of staking pools used in a cover will usually be small (probably most of the time just one) so calling directly from the pool might simplify things. It's entirely possible that the implementation may get unexpectedly complicated in which case I'll move it back here and implement.

contracts/modules/cover/Cover.sol
```diff
@@ -186,343 +185,185 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-    // retrieve extra required payment
-    retrievePayment(
-      premiumInPaymentAsset - refundInPaymentAsset,
-      buyCoverParams.paymentAsset,
-      buyCoverParams.commissionRatio,
-      buyCoverParams.commissionDestination
-    );
-  }
+    // ETH payment
+    if (paymentAsset == 0) {
```
danoctavian, line 338: we can add a constant here too for ETH_ASSET_ID

contracts/modules/cover/Cover.sol
```diff
@@ -186,343 +185,185 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-    if (coverAmountTrackingEnabled) {
-       totalActiveCoverInAsset[cover.coverAsset] =
-        totalActiveCoverInAsset[cover.coverAsset] - lastCoverSegment.amount + totalCoveredAmountInCoverAsset;
-    }
+      uint totalCoverAmountInNXM;
+      uint premiumInNXM;
 
-    emit CoverEdited(coverId, cover.productId, lastCoverSegmentIndex + 1, msg.sender);
-  }
+      for (uint i = 0; i < allocationRequests.length; i++) {
```
danoctavian, line 243: assuming the refund calculation is moved to the staking pool and you can only earn a refund within a particular pool, this loop looks cleaner as it is now

contracts/modules/cover/Cover.sol
```diff
@@ -186,343 +185,186 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-      * lastCoverSegment.amount
-      / PRICE_DENOMINATOR
-      * remainingPeriod
-      / MAX_COVER_PERIOD;
+      // TODO: implement using buckets
+      totalActiveCoverInAsset[params.coverAsset] += coverAmountInCoverAsset;
 
-    // Edit the last cover segment so it ends at the current block by writing to storage directly
-    _coverSegments[coverId][lastCoverSegmentIndex].period = lastCoverSegment.period - remainingPeriod;
+      emit CoverEdited(coverId, params.productId, segmentId, msg.sender, params.ipfsData);
```
kyledewy, line 304: Is it worth it to conditionally emit ipfsData here, like we do in `setProducts()`?

shark0der: Since each cover segment semantically makes sense and a new one is created on each edit - I would explicitly emit it every time in this case.

contracts/modules/cover/Cover.sol
```diff
@@ -571,34 +412,48 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-  function performStakeBurn(
+  function burnStake(
     uint coverId,
     uint segmentId,
     uint burnAmount
   ) external onlyInternal override returns (address /* owner */) {
 
-   CoverData storage cover =_coverData[coverId];
-
-   if (coverAmountTrackingEnabled) {
```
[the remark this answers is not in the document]
shark0der: Now that we surely now that we'll need it - we'll implement it using buckets.

Issue #489, opened 2022-11-10
StakingPool: depositTo & extendDeposit


roxdanila:
**depositTo**
- [x] reverts if caller is not cover contract or manager when pool is private
- [x] updates tranches including accNxmPerRewardsShare and lastAccNxmUpdate
- [x] reverts if deposit amount is 0
- [x] reverts if tranche id is not active
- [x] reverts if requested tranche expired
- [x] mints a new nft if token is 0
- [x] deposits to provided tokenId
- [x] reverts if provided tokenId is not valid
- [x] stores deposit stakeShares, rewardShares and lastAccNxmPerRewardShare
- [x] stores deposit pendingRewards if increasing existing deposit
- [x] updates pool manager rewards shares
- [x] updates tranche stake and reward shares
- [x] transfer total staked nxm to token controller contract
- [x] updates global variables activeStake, stakeSharesSupply and rewardsSharesSupply
- [x] Should emit some event - add the event in the contract as well

**extendDeposit**
- [ ] reverts if token id is 0
- [ ] reverts if new tranche ends before the initial tranche 
- [ ] reverts if new tranche is not yet available
- [ ] reverts if the new tranche already expired
- [ ] anyone can extend the deposit for a token id
- [ ] withdraws and make a new deposit if initial tranche is expired
- [ ] updates tranches including accNxmPerRewardsShare and lastAccNxmUpdate
- [ ] allows to increase the deposit
- [ ] updates the initial and new tranche stake shares and reward shares
- [ ] removes the initial tranche deposit and stores the new tranche deposit
- [ ] updates global stake and reward shares supply
- [ ] transfers increased nxm deposit to token controller
- [ ] Should emit some event - please add the event in the contract
- [ ] Should validate if the pool is private - pls implement this in the smart contract. Example:
    - Member deposit when pool is public
    - Pool switched to private
    - Member can use extendDeposit to move and increase deposit from non-expired tranche to new tranche

PR #260, opened 2022-08-08
Feature/integration tests cover

danoctavian:
Closes issues https://github.com/NexusMutual/smart-contracts/issues/294

PR #495, opened 2022-11-14
Test: Add stakingPool depositTo unit tests

patitonar:
## Context

Related to #489 

Closes https://github.com/NexusMutual/smart-contracts/issues/499
Closes https://github.com/NexusMutual/smart-contracts/issues/500

Adds new unit tests for `depositTo`. Followed the list of test cases from the issue and added new ones. 
The unit tests for `extendDeposit` will be implemented in a different PR.

Also, the following changes were applied to the `StakingPool` contract:
- Added StakeDeposited event
- Fix division by zero when calculating rewards
- Assign rewardPerSecond in allocateStake
- Fix rewards precision loss
- Fix basePremium calculation
- Improves tokenId validation

## Test plan

New tests were added in `test/unit/StakingPool/depositTo.js`

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


test/unit/StakingPool/depositTo.js
```diff
@@ -0,0 +1,425 @@
...
+      stakingPool,
+      initialStakeShares: firstDepositData.stakeShares,
+      stakeSharesIncrease: newStakeShares,
+      initialTrancheId: firstActiveTrancheId,
+      newTrancheId: firstActiveTrancheId,
+    });
+
+    // TODO: How can I make pendingRewards and lastAccNxmPerRewardShare to not be 0?
+    expect(updatedDepositData.pendingRewards).to.equal(0);
+    expect(updatedDepositData.lastAccNxmPerRewardShare).to.equal(0);
```
patitonar, line 367: @shark0der @roxdanila What are the easiest/ideal steps to accumulate rewards so these two variables are set to a value different from 0?

test/unit/StakingPool/depositTo.js
```diff
@@ -0,0 +1,425 @@
...
+    // at StakingPool.updateTranches (contracts/modules/staking/StakingPool.sol:283)
+    // at StakingPool.depositTo (contracts/modules/staking/StakingPool.sol:353)
+    await stakingPool.connect(user).depositTo([
+      {
+        amount,
+        trancheId: firstActiveTrancheId,
+        tokenId: depositNftId,
+        destination,
+      },
+    ]);
```
patitonar, line 423: @shark0der @roxdanila This call is reverting with division by zero. Found it randomly when working on tests. Is this a valid scenario/bug?

shark0der: Seems to be a valid bug that we should address. Good catch! Please add a `.skip` for now and we'll open an issue for it.

shark0der: Went ahead and implemented a fix as it seemed simple enough to do it straight away. Now it reverts with `StakingPool: Requested tranche has expired`. Please take a look!

patitonar: Awesome! Thanks!

test/unit/StakingPool/depositTo.js
shark0der: The only way to have rewards would be by allocating some capacity (ie cover buy)

roxdanila: see the integration tests for buy cover <> claims for a setup example: https://github.com/NexusMutual/smart-contracts/pull/260

patitonar: Thanks! I tried that in this test by calling `allocateStake()` (impersonating the cover contract) between two deposits but wasn't able to get values on those variables different from `0`. Probably I'm doing something wrong, I'll review the setup of the test and try different values and calls

roxdanila: cc @danoctavian

test/unit/StakingPool/depositTo.js
```diff
@@ -0,0 +1,807 @@
...
+          tokenId,
+          destination,
+        },
+      ]),
+    )
+      .to.emit(stakingPool, 'StakeDeposited')
+      .withArgs(user.address, amount, firstActiveTrancheId, depositNftId);
+  });
+
+  it.skip('reverts if provided tokenId is not valid', async function () {
```
patitonar, line 779: This test does not revert. There is a TODO in the code to validate this. Should we skip this test as it is now and addressed the fix later in a separate issue?

test/unit/StakingPool/depositTo.js
patitonar: I wasn't able to make the following variables not be 0.
```
deposit.pendingRewards
deposit.lastAccNxmPerRewardShare
```
After reviewing and debugging, it looks like the root cause is that the global variable `rewardPerSecond` is never set, which is used to calculate `accNxmPerRewardsShare` causing that variable to also be always 0.

The only time the `rewardPerSecond` variable is given a value is on `updateTranches` when expiring a bucket, which reduces its amount: `_rewardPerSecond -= rewardBuckets[_firstActiveBucketId].rewardPerSecondCut;`. 

I think it could be missing the code to increase `rewardPerSecond` with calculated rewards per second in `allocateStake()`.

Also, because of the previous thing mentioned, the two skipped tests right now revert with `Error: VM Exception while processing transaction: reverted with panic code 0x11 (Arithmetic operation underflowed or overflowed outside of an unchecked block)` in the line
`_rewardPerSecond -= rewardBuckets[_firstActiveBucketId].rewardPerSecondCut;`

Adding `rewardPerSecond += _rewardPerSecond;` in allocateStake() makes the test pass, but I'm not sure this is the correct way to fix this.

contracts/modules/staking/StakingPool.sol
```diff
@@ -675,10 +684,15 @@ contract StakingPool is IStakingPool, ERC721 {
...
-      uint _rewardPerSecond = rewards / (expireAtBucket * BUCKET_DURATION - block.timestamp);
+      // TODO UNDO the following change. Amplified the rewards given that this calculation was always 0 given that 
+      // rewards are too small compared to the denominator result
+      uint _rewardPerSecond = rewards * 10000 / (expireAtBucket * BUCKET_DURATION - block.timestamp);
 
       // 1 SLOAD + 1 SSTORE
       rewardBuckets[expireAtBucket].rewardPerSecondCut += _rewardPerSecond;
+
+      // TODO Temporary fix while debugging rewardPerSecond assignation
+      // rewardPerSecond += _rewardPerSecond;
```
patitonar, line 695: This was the temporary fix I mentioned I made to make the tests work, assigning rewards to `rewardPerSecond`

test/unit/StakingPool/depositTo.js
danoctavian: @shark0der  i checked the code as well and it seems that yes we don't have a logic that increments `rewardPerSecond` value as @patitonar says.

shark0der: @patitonar good catch!

1. Yes, the `rewardPerSecond += _rewardPerSecond;` is the correct way to fix - feel free to uncomment that code.

2.  I'm looking at the added `rewards * 10000`. I logged the premium and the rewards and I got 41 and 20 (wei) which is definitely the wrong amount for the requested `1e18` amount. Here's all I logged:
```
console.log("allocated for cover: %s", allocatedCoverAmount);
console.log("initial cap used   : %s", initialCapacityUsed);
console.log("total cap used     : %s", totalCapacity);
console.log("premium            : %s", premium);
console.log("rewards            : %s", rewards);
console.log("_rewardPerSecond   : %s", _rewardPerSecond);

Printed:
allocated for cover: 100
initial cap used   : 0
total cap used     : 200000000
premium            : 41
rewards            : 20
_rewardPerSecond   : 0
```
The bug is likely in the capacity calculations, we'll have to debug that.

3. I think there's one more item that we might want to fix. Given that we divide the total amount of rewards to be distributed to the number of seconds it's being streamed - we'll end up with some precision loss. I guess the biggest error will happen with the smallest cover over the longest period of time, which would be a cover of 0.01 nxm (smallest allocation unit) for 1 year. The cheapest price there can be would be 1% per year, which gets us to a 0.0001 (=1e14) nxm premium. `1e14 - 1e14 // (365 * 24 * 3600) * (365 * 24 * 3600) = 6256000` wei-nxm not being distributed.  That's a 0.000006% error, which is not tragic but we need to address this. I think the proper way to address this is to recalculate the `rewards` amount:
```
uint rewardStreamPeriod = expireAtBucket * BUCKET_DURATION - block.timestamp;
uint _rewardPerSecond = rewards / rewardStreamPeriod;
rewards = _rewardPerSecond * rewardStreamPeriod;
````

@patitonar would you please apply (1) and (3) in the solidity code?

File: test/unit/StakingPool/depositTo.js (hunk 3 above)
Comment (roxdanila) [reply test/unit/StakingPool/depositTo.js:779]: you can add validation in the contract for this one, looks like a quick one

**Comment by roxdanila** [comment]
nice catches so far!

File: test/unit/StakingPool/depositTo.js (hunk 1 above)
Comment (patitonar) [reply test/unit/StakingPool/depositTo.js]: Thanks! I'll apply (1) and (3) and dig more on (2) to try to find the bug

File: test/unit/StakingPool/depositTo.js [hunk 5]
```diff
@@ -0,0 +1,774 @@
...
+      },
+    ]);
+
+    const secondDepositData = await stakingPool.deposits(depositNftId, firstActiveTrancheId);
+    const secondAccNxmPerRewardsShare = await stakingPool.accNxmPerRewardsShare();
+
+    expect(secondAccNxmPerRewardsShare).to.not.equal(0);
+    expect(secondDepositData.lastAccNxmPerRewardShare).to.equal(secondAccNxmPerRewardsShare);
+    // TODO: Shouldn't pendingRewards also be updated?
+    expect(secondDepositData.pendingRewards).to.equal(0);
```
patitonar, line 365: `deposit.pendingRewards` is not being increased in the second deposit because `deposit.lastAccNxmPerRewardShare` is set at the end of the second deposit, so it is only updated on the third deposit.

Is this the expected behavior? Based on the comment, it looks like the intention is to update `pendingRewards` after the first deposit. Should we use a different variable such as `deposit.rewardsShares != 0` on this `if` statement?  
  
https://github.com/NexusMutual/smart-contracts/blob/896dbdcb08cc4a89191ef133087d1e9d72ac6517/contracts/modules/staking/StakingPool.sol#L408-L412

test/unit/StakingPool/depositTo.js
patitonar: Applied (1) here https://github.com/NexusMutual/smart-contracts/pull/495/commits/fbec993e9cd01458eeb144408a740b94ae2a4064
Applied (3) here https://github.com/NexusMutual/smart-contracts/pull/495/commits/e373490f618e5979d2173ac4ed309553d814d009

Regarding (2) I reviewed the code, and I'm not 100% but I think I found the bug, although I may be missing something or interpreting something incorrectly. I started following why the `premium` variable was so small.

It is calculated here and it is supposed to be in 18 decimals https://github.com/NexusMutual/smart-contracts/blob/896dbdcb08cc4a89191ef133087d1e9d72ac6517/contracts/modules/staking/StakingPool.sol#L656-L663

The returned value from `getPremium` is based on the result from `calculatePremiumPerYear()` https://github.com/NexusMutual/smart-contracts/blob/896dbdcb08cc4a89191ef133087d1e9d72ac6517/contracts/modules/staking/StakingPool.sol#L1529-L1538

On the tests, it was always returning the `basePremium` variable which is calculated as following https://github.com/NexusMutual/smart-contracts/blob/896dbdcb08cc4a89191ef133087d1e9d72ac6517/contracts/modules/staking/StakingPool.sol#L1541-L1551

Here `basePrice` was expected to be in 18 decimals but instead `basePrice` was always `product.targetPrice` which is a percentage expressed with 2 decimals. So I think the formula is not properly using the `basePrice` variable.

So I applied the following fix https://github.com/NexusMutual/smart-contracts/pull/495/commits/88cf5c269af0d405f5976651f790989436728ee7

```solidity
// scale coverAmount to 18 decimals and apply price percentage
uint basePremium = (coverAmount * NXM_PER_ALLOCATION_UNIT) * basePrice / TARGET_PRICE_DENOMINATOR;
```

After this change, the results were more reasonable and the tests passed. We will need to run some tests with specific numbers to validate that formula is correct when unit testing `allocateStake()` .

Thoughts on this?

patitonar: 4. Also, when I was reviewing the code I found that `PRICE_CHANGE_PER_DAY` is 18 decimals. The calculated `priceDrop` is huge compared to `product.nextPrice` and `product.targetPrice` so on the test, it always uses `product.targetPrice` as the `basePrice`. Shouldn't `PRICE_CHANGE_PER_DAY` be expressed or scaled into a similar value like nextPrice and targetPrice? (I guess this also affects `PRICE_BUMP_RATIO`) 

https://github.com/NexusMutual/smart-contracts/blob/896dbdcb08cc4a89191ef133087d1e9d72ac6517/contracts/modules/staking/StakingPool.sol#L1508-L1519

contracts/modules/staking/StakingPool.sol
```diff
@@ -279,7 +279,10 @@ contract StakingPool is IStakingPool, ERC721 {
         uint bucketStartTime = _firstActiveBucketId * BUCKET_DURATION;
         uint elapsed = bucketStartTime - _lastAccNxmUpdate;
 
-        uint newAccNxmPerRewardsShare = elapsed * _rewardPerSecond / _rewardsSharesSupply;
+        uint newAccNxmPerRewardsShare = _rewardsSharesSupply != 0
```
danoctavian, line 282: This looks fine on the surface

@shark0der  let us know if it's correct

contracts/modules/staking/StakingPool.sol
```diff
@@ -1544,11 +1557,9 @@ function setProducts(StakedProductParam[] memory params) external onlyManager {
     uint initialCapacityUsed,
     uint totalCapacity
   ) public pure returns (uint) {
-
-    // base price has 18 decimals
     // cover amount has 2 decimals (100 = 1 unit)
-    // dividing by ALLOCATION_UNITS_PER_NXM (=100) to get the right amount of decimals
-    uint basePremium = basePrice * coverAmount / ALLOCATION_UNITS_PER_NXM;
+    // scale coverAmount to 18 decimals and apply price percentage
+    uint basePremium = (coverAmount * NXM_PER_ALLOCATION_UNIT) * basePrice / TARGET_PRICE_DENOMINATOR;
```
danoctavian, line 1562: if basePrice has 18 decimals

coverAmount has 2 decimals + 16 from NXM_PER_ALLOCATION_UNIT

thats' a total of 36.

the division is then by TARGET_PRICE_DENOMINATOR = 100_00

so the final basePremium is off in terms of scaling.

test/unit/StakingPool/depositTo.js
```diff
@@ -0,0 +1,774 @@
...
+    await expect(
+      stakingPool.connect(coverSigner).depositTo([
+        {
+          amount,
+          trancheId,
+          tokenId,
+          destination,
+        },
+      ]),
+    ).to.not.be.revertedWith('StakingPool: The pool is private');
```
danoctavian, line 78: i would just make this to.not.be.reverted

let's make a successful call

test/unit/StakingPool/depositTo.js
```diff
@@ -0,0 +1,774 @@
...
+    await expect(
+      stakingPool.connect(manager).depositTo([
+        {
+          amount,
+          trancheId,
+          tokenId,
+          destination,
+        },
+      ]),
+    ).to.not.be.revertedWith('StakingPool: The pool is private');
```
danoctavian, line 89: i would just make this to.not.be.reverted

let's make a successful call

test/unit/StakingPool/depositTo.js
```diff
@@ -0,0 +1,774 @@
...
+      const totalSupply = await stakingPool.totalSupply();
+      expect(totalSupply).to.equal(1);
+    }
+
+    await stakingPool.connect(user).depositTo([
+      {
+        amount,
+        trancheId: firstActiveTrancheId,
+        tokenId,
+        destination,
```
danoctavian, line 176: let's zero out destination so it just mints to user.

we're testing the "no defined destination" situation here vs the next test.

test/unit/StakingPool/depositTo.js
```diff
@@ -0,0 +1,774 @@
...
+      period: 3600 * 24 * 30, // 30 days
+      gracePeriod: 3600 * 24 * 30,
+      globalCapacityRatio: 20000,
+      capacityReductionRatio: 0,
+      rewardRatio: 10000,
+    };
+    await stakingPool.connect(this.coverSigner).allocateStake(coverRequest);
+
+    await increaseTime(
+      20 * // days
```
danoctavian, line 343: let's use a function for this calculation

we have daysToSeconds defined in a couple of places 

can move it to.a common utils file as well and import from there

test/unit/StakingPool/depositTo.js
```diff
@@ -0,0 +1,774 @@
...
+        trancheId: firstActiveTrancheId,
+        tokenId: depositNftId,
+        destination,
+      },
+    ]);
+
+    const secondDepositData = await stakingPool.deposits(depositNftId, firstActiveTrancheId);
+    const secondAccNxmPerRewardsShare = await stakingPool.accNxmPerRewardsShare();
+
+    expect(secondAccNxmPerRewardsShare).to.not.equal(0);
```
danoctavian, line 362: we want to test it's the right value not just different from 0 we should compute an exact value here

contracts/modules/staking/StakingPool.sol
patitonar: In this other comment is the context for this change https://github.com/NexusMutual/smart-contracts/pull/495#discussion_r1024463714

test/unit/StakingPool/depositTo.js
patitonar: Updated the test here [Improve private pool unit test](https://github.com/NexusMutual/smart-contracts/pull/495/commits/6b6b6e6997dbac0f76f47f53392357be9409c785)

test/unit/StakingPool/depositTo.js
patitonar: Thanks! Found that method in the libs, updated the code to use it [Use daysToSeconds helper method](https://github.com/NexusMutual/smart-contracts/pull/495/commits/3fdabc6f56118807ad09e24e29fbf9a7c731f604)

test/unit/StakingPool/depositTo.js
patitonar: The expects were confusing, here we are validating `secondDepositData.lastAccNxmPerRewardShare` got updated. In the next line, it is checking the correct value. The intention was to validate that it got assigned the correct value (the global accNxmPerRewardsShare) and that the correct value is not 0. I didn't want to over complex the test with the specific calculation of accNxmPerRewardsShare as its value depends from the calculation in `updateTranches()` which eventually will be properly tested and we can validate the calculations there.

Updated the code to reflect that better here [Update deposit lastAccNxmPerRewardShare expects](https://github.com/NexusMutual/smart-contracts/pull/495/commits/9811ccc60543ae5521b127dc8ab7371f25808d4f)

test/unit/StakingPool/depositTo.js
patitonar: The `destination` variable is set to `AddressZero` in the setup. So all tests use the "no defined destination" except for the next test where we validate a specific destination. Do you think the default destination value and naming are confusing and should be improved?

contracts/modules/staking/StakingPool.sol
```diff
@@ -667,11 +676,15 @@ contract StakingPool is IStakingPool, ERC721 {
 
       uint rewards = premium * request.rewardRatio / REWARDS_DENOMINATOR;
       uint expireAtBucket = Math.divCeil(block.timestamp + request.period, BUCKET_DURATION);
-      uint _rewardPerSecond = rewards / (expireAtBucket * BUCKET_DURATION - block.timestamp);
+      uint rewardStreamPeriod = expireAtBucket * BUCKET_DURATION - block.timestamp;
+      uint _rewardPerSecond = rewards / rewardStreamPeriod;
+      rewards = _rewardPerSecond * rewardStreamPeriod;
```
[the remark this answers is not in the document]
kyledewy: yea seems to be a no-op. i think the second assignment of rewards can be removed.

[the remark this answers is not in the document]
shark0der: This was my suggestion. The purpose of this is exactly for avoiding precision loss. Basically if you stream 7 wei over 3 seconds, you'll be left with 1 undistributed wei. The operation calculates what's the redundant amount and corrects the rewards so we won't mint dust.

test/unit/StakingPool/depositTo.js
shark0der: Hey @patitonar, very good observations. _I think_ we ended up with these values because we initially had 18 decimals for price (initial price, next price, base price) but changed the variable sizes in structs. Your both suggestions make sense:
- https://github.com/NexusMutual/smart-contracts/commit/88cf5c269af0d405f5976651f790989436728ee7 looks correct to me
- `PRICE_CHANGE_PER_DAY` and `PRICE_BUMP_RATIO` are indeed off. I'll leave a comment with a fix suggestion in a comment below

shark0der:
```
diff --git a/contracts/modules/staking/StakingPool.sol b/contracts/modules/staking/StakingPool.sol
index 65492804..16931a84 100644
--- a/contracts/modules/staking/StakingPool.sol
+++ b/contracts/modules/staking/StakingPool.sol
@@ -118,13 +118,13 @@ contract StakingPool is IStakingPool, ERC721 {
   uint public constant INITIAL_PRICE_DENOMINATOR = 100_00;
   uint public constant TARGET_PRICE_DENOMINATOR = 100_00;
 
-  // base price bump is +0.2% for each 1% of capacity used, ie +20% for 100%
-  // 20% = 0.2
-  uint public constant PRICE_BUMP_RATIO = 0.2 ether;
+  // base price bump
+  // +0.2% for each 1% of capacity used, ie +20% for 100%
+  uint public constant PRICE_BUMP_RATIO = 20_00; // 20%
 
   // next price smoothing
-  // 0.005 ether = 0.5% out of 1e18
-  uint public constant PRICE_CHANGE_PER_DAY = 0.005 ether;
+  // 0.5% per day
+  uint public constant PRICE_CHANGE_PER_DAY = 50; // 0.5%
 
   // +2% for every 1%, ie +200% for 100%
   uint public constant SURGE_PRICE_RATIO = 2 ether;
```

shark0der: You can run `cat | git apply`, then paste the above code, followed by a `ctrl+d`

test/unit/StakingPool/depositTo.js
shark0der: @patitonar The change suggestion makes sense, very important to keep the comment as well otherwise it looks like a no-op. Feel free to apply the chane in this PR.

test/unit/StakingPool/depositTo.js
kyledewy: Using `deposit.rewardShares`  seems like a good fix to me. Can you confirm this is the expected behavior @shark0der?

test/unit/StakingPool/depositTo.js
patitonar: Thanks! Applied the fixes for `PRICE_CHANGE_PER_DAY` and `PRICE_BUMP_RATIO` here https://github.com/NexusMutual/smart-contracts/pull/495/commits/4e15371418ff847b4c30d1c0d525bbe7b70091b7

test/unit/StakingPool/depositTo.js
patitonar: Applied the improvement - including the comment- here https://github.com/NexusMutual/smart-contracts/pull/495/commits/79d213f2c88af227e35eeaa271a01f149baffdf3

contracts/modules/staking/StakingPool.sol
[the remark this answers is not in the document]
danoctavian: interesting. 

a comment explaining this above that line is really useful.

test/unit/StakingPool/depositTo.js
danoctavian: I see, i missed the fact that destination = AddressZero

should be fine.

test/unit/StakingPool/depositTo.js
patitonar: @shark0der Applied the fix here https://github.com/NexusMutual/smart-contracts/pull/495/commits/2e47b2f0d9285d5a1878ec847276a44bbc4fe9ed

shark0der (approved):
LGTM :rocket:

Issue #382, opened 2022-09-16
Refactor the contracts so they depend on a single version of MasterAware


roxdanila:
MasterAware is a contract meant to be inherited that offers easier access to data stored in the NXMaster contract (which is our on-chain contract address registry basically).

Initially we had the Iupgradable "thing" that was poorly named and had duplicate values in storage (ms and master). That was renamed to LegacyMasterAware, but nothing was changed in it.

Secondly we have MasterAware which provides basically the same functionality but has no storage duplicates and has sane modifier names.

Lastly, Dragos unilaterally developed MasterAwareV2 and used it mostly in the claim assessment module. It uses a different implementation - it stores everything in a mapping instead of separates slots, and indexes are inside an enum. This isn't bad, it's quite good actually because at some point we almost bricked TokenController because it's a proxy and we moved a variable around. Luckily we were able to perform an upgrade and fix it. Having everything in a mapping makes it safer. We still have to pull and populate the needed contracts whenever changeDependentContractAddress is called and make sure we don't miss one (currently unsure how can we make this check at compile time).

- LegacyMasterAware is currently used by Governance, ProposalCategory, MemberRoles, TokenController, ~~ClaimsReward~~, ~~QuotationData~~, ~~ClaimsData~~ - the first 4 are proxies
- MasterAware is used by ~~Gateway~~*, ~~PooledStaking~~ and MCR* - the first 2 are proxies
- MasterAwareV2 is used by Cover, Assessment, IndividualClaims, YieldTokenIncidents, CoverMigrator, AssessmentViews - all v2 contracts, nothing in prod yet, first 4 are meant to be proxies

The crossed out items are going away in v2 so we don't have to worry about them.
* Gateway is softly deprecated, as in we can't yet get rid of it as it will be hopefully used by some distributors but for no longer than a year until all covers expire and then we can trash it.
* MCR is not a proxy so we can just switch to a different implementation without any hiccups.

If we don't act, we'll end up with 3 different versions of this contract in prod which is clearly not ideal as it will be first of all confusing, secondly we'll have to be aware which contracts are using what and make sure we don't accidentally break anything.

**Solution TBD**

PR #502, opened 2022-11-17
Feature: MasterAwareV2 normalization

danoctavian:
## Context

Issue: https://github.com/NexusMutual/smart-contracts/issues/382

Full motivation for the change is explained in this doc:
https://docs.google.com/document/d/1s-L1FvBelHK-T009ydJL2R0KYRSZapgfbcq4IKqbN1E/edit?usp=sharing

Initially we had the Iupgradable abstract contract that was poorly named and had duplicate values in storage (ms and master). That was renamed to LegacyMasterAware, but nothing was changed in it.

Secondly we have MasterAware which provides basically the same functionality but has no storage duplicates and has sane modifier names.

Thirdly we have MasterAwareV2 which is the new interface that uses a mapping from an Id to a contract address to store the dependencies. This is nicer because it takes only 2 memory slots (the mapping and the master address).

We want to normalize the use of MasterAwareV2 as much as possible.

Here we are moving MCR, MemberRoles and TokenController to MasterAwareV2 interface.

It works because you can drop in replace LegacyMasterAware with MasterAwareV2 - they both take 2 memory slots AND the mapping ```  mapping(uint => address payable) internal internalContracts;``` from MasterAwareV2 is falling in the place of ```  INXMMaster public ms;```. 

This is considered safe because the mapping memory slot does not need to be cleared - it is not used by the mapping data structure.

Motivation for skipping ProposalCategory and Governance:

They are to be refactored when we are overhauling governance fully. The benefits are not there for this upgrade right now.

Hurdles: Governance.sol is fixed to 0.5.0 and needs updates to comply with 0.8.16. 

ProposalCategory generates a stack to deep error when upgraded to the 0.8.16 compiler.

## Test plan

No extra tests added. All current tests required to pass.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/modules/governance/MemberRoles.sol
```diff
@@ -223,14 +233,16 @@ contract MemberRoles is IMemberRoles, Governed, LegacyMasterAware {
...
   ) external override {
     _switchMembership(msg.sender, newAddress);
-    nxm.transferFrom(msg.sender, newAddress, nxm.balanceOf(msg.sender));
+    INXMToken _token = token();
+    _token.transferFrom(msg.sender, newAddress, _token.balanceOf(msg.sender));
 
+    ICover _cover = cover();
     // Transfer the cover NFTs to the new address, if any were given
-    cover.transferCovers(msg.sender, newAddress, coverIds);
+    cover().transferCovers(msg.sender, newAddress, coverIds);
```
kyledewy, line 241: `cover()` is called twice here. i think you meant to use the `_cover` variable instead?

contracts/interfaces/IMasterAwareV2.sol
```diff
@@ -10,7 +10,10 @@ interface IMasterAwareV2 {
     MR, // MemberRoles.sol
     MC, // MCR.sol
     CO, // Cover.sol
-    AS  // Assessment.sol
+    AS, // Assessment.sol
```
kyledewy, line 13: [TK, PS, GV] are missing comments for which contract they reference.

kyledewy (approved):
Left a couple minor comments, but otherwise LGTM :+1:

shark0der (changes requested):
Let's make an effort to get rid of `MasterAware.sol`

danoctavian:
Removed MasterAware from all except LegacyGateway.sol

As discussed with @roxdanila  we'll postpone messing with that contract until we clarify the biz requirements around what we need to support there.

Then we'll cleanup the MasterAware as well.

contracts/modules/legacy/LegacyPooledStaking.sol
```diff
@@ -44,10 +44,7 @@ contract LegacyPooledStaking is IPooledStaking, MasterAware {
 
   /* Storage variables */
 
-  bool public initialized;
-
-  INXMToken public token;
-  ITokenController public tokenController;
+  ITokenController public _unused0;
```
shark0der, line 47: `token` and `initialized` were removed - let's leave a comment of what happens with the storage layout here as it's not obvious why this is OK

contracts/modules/capital/Pool.sol
```diff
@@ -405,14 +401,29 @@ contract Pool is IPool, MasterAware, ReentrancyGuard {
...
    * @dev Implements MasterAware interface function
    */
   function changeDependentContractAddress() public {
-    nxmToken = INXMToken(master.tokenAddress());
-    tokenController = ITokenController(master.getLatestAddress("TC"));
-    mcr = IMCR(master.getLatestAddress("MC"));
+    internalContracts[uint(ID.TK)] = payable(master.tokenAddress());
+    internalContracts[uint(ID.TC)] = master.getLatestAddress("TC");
+    internalContracts[uint(ID.MC)] = master.getLatestAddress("MC");
+    internalContracts[uint(ID.MR)] = master.getLatestAddress("MR");
```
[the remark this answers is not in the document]
danoctavian: Yes. 

Any MasterAwareV2 contract is using it:

https://github.com/NexusMutual/smart-contracts/blob/3bced02d04b378ed8e70deaadd1ef33a2112fb39/contracts/abstract/MasterAwareV2.sol#L17

shark0der (approved):

PR #507, opened 2022-11-21
Fix: Audit - Staking 3.1.3 Withdrawal of NXM locked for governance voting

### LINKED ISSUE #308: Audit - Staking 3.1.3 Withdrawal of NXM locked for governance voting
[Assessment.sol#L139](https://github.com/NexusMutual/smart-contracts/blob/14a7ed6f5b6df877a2cb7eb9fd390c9cb216ccb2/contracts/modules/assessment/Assessment.sol#L139) [StakingPool.sol#L449](https://github.com/NexusMutual/smart-contracts/blob/14a7ed6f5b6df877a2cb7eb9fd390c9cb216ccb2/contracts/modules/staking/StakingPool.sol#L449)

**Description**
The stake() function in Assessment.sol does not check whether the staker's NXM is locked for voting before performing an operatorTransfer() via TokenController. As such, once a member has voted on a governance proposal, they would be able to transfer their NXM to another member by calling stake() in Assessment.sol, which will give Assessment.sol custody of the NXM, and then call unstake(), specifying the other member as the recipient of the unstaked NXM.

Two colluding members could abuse this functionality to vote twice using the same NXM, or vote using NXM obtained from a flashloan. This is likely only to affect special resolution votes, where voteWeight is directly proportional to the total NXM balance of the voter.

This issue is also present in StakingPool.sol, where a member can stake their NXM, despite having voted in governance. This is because the call in depositTo() to tokenController.depositStakedNXM() does not account for tokens locked in governance. However, this avenue of attack does not allow instant withdrawal of funds, as in the case of Assessment.sol.

**Remediation**
A check should be added before the operatorTransfer()/depositStakedNXM() functions to verify whether the member's NXM is locked in a governance vote. Similar to other places in code, this check might look something like require(block.timestamp > nxm.isLockedForMV(msg.sender));.

kyledewy:
## Context

Closes #308 

This PR addresses audit issue 3.1.3, regarding the ability to move NXM while it should be locked for voting. There are 2 attack vectors pointed out by the auditors:

1.) Vote on proposal, then stake in Assessment, then unstake to another address to vote 2nd time on proposal, and repeat.....

2.) Vote on proposal, then deposit to StakingPool, so the  manager can vote 2nd time on proposal

This was possible because `nxm.operatorTransfer()`  doesn't check that the NXM is locked for a member vote. This PR adds this check in `stakingPool.depositTo()` and in `Assessment.unstake()`.  
 
Note: the auditors suggested to put this check in `Assessment.stake()`, but this seems too restrictive, as there are legitimate situations where a member who voted on a governance proposal, should be allowed to go and vote on claims submissions after.

## Test plan
I added tests in: `unit/StakingPool/depositTo.js` and `unit/Assessment/unstake.js`

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/assessment/Assessment.sol
+++ b/contracts/modules/assessment/Assessment.sol
@@ -138,6 +138,7 @@ contract Assessment is IAssessment, MasterAwareV2 {
   /// @param amount  The amount of nxm to stake
   function stake(uint96 amount) public whenNotPaused {
     stakeOf[msg.sender].amount += amount;
+    require(block.timestamp > nxm.isLockedForMV(msg.sender), "Assessment: Nxm locked for voting");
     ITokenController(getInternalContractAddress(ID.TC))
       .operatorTransfer(msg.sender, address(this), amount);
 
--- a/contracts/modules/assessment/Assessment.sol
+++ b/contracts/modules/assessment/Assessment.sol
@@ -16,7 +16,7 @@ import "../../libraries/SafeUintCast.sol";
 /// to payouts. Mints rewards for stakers that act benevolently and allows burning fraudulent ones.
 contract Assessment is IAssessment, MasterAwareV2 {
 
-  INXMToken public immutable nxm;
+  INXMToken public immutable NXM;
 
   /* ========== STATE VARIABLES ========== */
 
@@ -43,7 +43,7 @@ contract Assessment is IAssessment, MasterAwareV2 {
   /* ========== CONSTRUCTOR ========== */
 
   constructor(address nxmAddress) {
-    nxm = INXMToken(nxmAddress);
+    NXM = INXMToken(nxmAddress);
   }
 
   function initialize () external {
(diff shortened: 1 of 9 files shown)


contracts/modules/assessment/Assessment.sol
```diff
@@ -153,6 +153,7 @@ contract Assessment is IAssessment, MasterAwareV2 {
   ///                membership during stake lockup period and thus allowing the user to withdraw
   ///                their staked amount to the new address when possible.
   function unstake(uint96 amount, address to) external whenNotPaused override {
+    require(block.timestamp > nxm.isLockedForMV(msg.sender), "Assessment: Nxm locked for voting");
```
shark0der, line 156: I think we can relax this requirement a bit. If the member has voted in governance require the `to` to be `msg.sender`. What do you think?

shark0der: Also please make the token symbol uppercase - `NXM` (it's already uppercased in staking).

contracts/modules/staking/StakingPool.sol
```diff
@@ -356,6 +356,8 @@ contract StakingPool is IStakingPool, ERC721 {
       );
     }
 
+    require(block.timestamp > nxm.isLockedForMV(msg.sender), "Staking: Senders NXM is locked for member vote");
```
shark0der, line 359: Would skip the word "Senders" to have it similar to assessment message and I'd suggest replacing "member vote" with "voting in governance"

shark0der (approved):

PR #512, opened 2022-11-22
Fix: updateTranches not expiring tranches

patitonar:
## Context

This PR fixes two issues related to expiring tranches from the `updateTranches` method.

1. Expired tranche is not being expired: In a new pool, if a tranche has a deposit and then no other activity until after the tranche end time passes, when calling `updateTranches` that tranche was not being expired. The reason is that `firstActiveBucketId` is `0` so the first active bucket and tranche were being assigned to the `currentBucketId` and `currentTrancheId`, so then the `while` condition was `false` and the tranche was not expired.
2. Division by zero error: If multiple expired tranches are being expired in a `updateTranches` call, if all stakeShares were burned before the last expiring tranche, a division by zero error was being thrown.
https://github.com/NexusMutual/smart-contracts/blob/882cc4236c912cdf6180701fbc70e5fa34a557c5/contracts/modules/staking/StakingPool.sol#L318

The following changes were applied:
1. Assign the global variables `firstActiveBucketId` and `firstActiveTrancheId` to the current bucket/tranche if they are not set. In the scenario described this will make the internal call to `updateTranches` in `depositTo` to store those variables, so next time when `updateTranches` is called, they will be correctly compared to the current tranche and enter the `while` block.
2. Validate that `_stakeSharesSupply` is not `0`, return `0` otherwise.
4. Minor fix to `depositTo` unit test setup. Moved the logic from `before` to `beforeEach` as it was not being reverted by the snapshot, affecting tests on other files.

## Test plan

New tests were added in `test/unit/StakingPool/updateTranches.js`. You can validate that previously it was not working by removing the changes applied, making the tests fail.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/staking/StakingPool.sol
+++ b/contracts/modules/staking/StakingPool.sol
@@ -215,17 +215,18 @@ contract StakingPool is IStakingPool, ERC721 {
   // updateUntilCurrentTimestamp forces rewards update until current timestamp not just until
   // bucket/tranche expiry timestamps. Must be true when changing shares or reward per second.
   function updateTranches(bool updateUntilCurrentTimestamp) public {
+
+    uint _firstActiveBucketId = firstActiveBucketId;
+    uint _firstActiveTrancheId = firstActiveTrancheId;
+
     uint currentBucketId = block.timestamp / BUCKET_DURATION;
     uint currentTrancheId = block.timestamp / TRANCHE_DURATION;
 
     // if the pool is new
-    if (firstActiveBucketId == 0) {
-      firstActiveBucketId = currentBucketId;
-      firstActiveTrancheId = currentTrancheId;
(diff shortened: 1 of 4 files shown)


patitonar:
@shark0der  Reverted the change on `updateTranches` to avoid initializing the bucket/tranche ids in a direct call if the pool is empty. As an alternative solution, I moved the initialization to `depositTo` so the ids are only initialized once when the pool receives the first deposit. 

Probably there could be some gas improvements on `updateTranches` related to the validation of  `if (_firstActiveBucketId == 0)` to avoid duplicate checks or unnecessary checks when the pool is empty or not. I'll give it a thought after confirming the current solution approach looks good

shark0der (approved):

Issue #504, opened 2022-11-18
Support fixed pricing on selected products


roxdanila:
We could have a flag that is set at listing time (which is an AB controlled process), flag means either:
- regular dynamic pricing applies and risk can be sold by any syndicate
- fixed pricing applies and risk can be sold by a designated syndicate only

Final mechanism TBD

PR #513, opened 2022-11-22
Feature: support fixed price

danoctavian:
## Context

Issue:

https://github.com/NexusMutual/smart-contracts/issues/504

Add a flag for each product that specifies whether a product has fixed pricing, plus a list of whitelisted pool ids that the respective product capacity can be allocated to.

 StakingPool.setProducts method now checks that the Pool's id is whitelisted if a product is a fixed price product.

 Whenever capacity is allocated during a buyCover flow, a useFixedPricing flag is passed to the staking pool to apply the targetPrice to that cover for pricing and bypass the normal pricing formula.

## Test plan

A buyCover unit test is written for a fixed price product.

An integration test is necessary to test buyCover using a fixed price product. 

An integration test is necessary to test StakingPool.setProducts for a product that has fixedPricing = true (test both valid and invalid pool id that should revert).

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -72,6 +72,8 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
   CoverData[] private _coverData;
   mapping(uint => mapping(uint => PoolAllocation[])) public coverSegmentAllocations;
 
+  mapping(uint => uint[]) public fixedPricingAllowedPools;
+
   // Each cover has an array of segments. A new segment is created
   // every time a cover is edited to deliniate the different cover periods.
   mapping(uint => CoverSegment[]) private _coverSegments;
@@ -257,7 +259,8 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
           params.productId,
           coverId,
           coveredAmountInNXM,
-          params.period
+          params.period,
+          product.fixedPricing
         );
 
         AllocationRequestConfig memory config = AllocationRequestConfig(
@@ -304,6 +307,25 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
 
       emit CoverEdited(coverId, params.productId, segmentId, msg.sender, params.ipfsData);
     }
+  }
+
+
+  function isValidFixedPricingPool(uint productId, uint poolId) external returns (bool) {
+
+    if (!_products[productId].fixedPricing) {
+      // no pool id restrictions
+      return true;
+    }
+    bool poolIsValid;
+    uint[] memory allowedPools = fixedPricingAllowedPools[productId];
+    for (uint i = 0; i < allowedPools.length; i++) {
+      if (allowedPools[i] == poolId) {
+        poolIsValid = true;
+        break;
+      }
+    }
+
+    return poolIsValid;
   }
 
   function retrievePayment(
@@ -527,6 +549,12 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
         product.capacityReductionRatio <= CAPACITY_REDUCTION_DENOMINATOR,
         "Cover: capacityReductionRatio > 100%"
       );
+
+      if (product.fixedPricing) {
+
+        uint productId = param.productId == type(uint256).max ? _products.length : param.productId;
+        fixedPricingAllowedPools[productId] = param.allowedPools;
+      }
 
       // New product has id == uint256.max
       if (param.productId == type(uint256).max) {
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -310,7 +310,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
   }
 
 
-  function isAllowedPool(uint productId, uint poolId) external returns (bool) {
+  function isAllowedPool(uint productId, uint poolId) external view returns (bool) {
 
     if (!_products[productId].useFixedPrice) {
       // no pool id restrictions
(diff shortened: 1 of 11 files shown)


contracts/modules/cover/Cover.sol
```diff
@@ -402,6 +416,21 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     emit CoverEdited(coverId, cover.productId, lastCoverSegmentIndex + 1, msg.sender);
   }
 
+  function checkFixedPricingAllowedPools(uint productId, PoolAllocationRequest[] memory allocationRequests) internal {
+
+    uint[] memory allowedPools = fixedPricingAllowedPools[productId];
+    for (uint i = 0; i < allocationRequests.length; i++) {
+      bool poolAllowed = false;
+      for (uint j = 0; j < allowedPools.length; i++) {
```
kyledewy, line 424: note that the second-for loop is incrementing i, instead of j

danoctavian: good find

contracts/modules/staking/StakingPool.sol
```diff
@@ -1511,15 +1514,13 @@ function setProducts(StakedProductParam[] memory params) external onlyManager {
...
-    uint initialCapacityUsed,
-    uint totalCapacity
-  ) internal returns (uint) {
+  function getPremium(PremiumParameters memory params) internal returns (uint) {
 
-    StakedProduct memory product = products[productId];
+    StakedProduct memory product = products[params.productId];
+
+    if (params.useFixedPrice) {
+      return product.targetPrice * params.period / 365 days;
```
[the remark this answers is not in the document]
danoctavian: also not sure if this is the final formula.. need to check again against the premium calculation

contracts/modules/cover/Cover.sol
```diff
@@ -402,6 +416,21 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     emit CoverEdited(coverId, cover.productId, lastCoverSegmentIndex + 1, msg.sender);
   }
 
+  function checkFixedPricingAllowedPools(uint productId, PoolAllocationRequest[] memory allocationRequests) internal {
+
+    uint[] memory allowedPools = fixedPricingAllowedPools[productId];
```
danoctavian, line 421: The reason i made this an array and not a mapping is that because when i do setProducts i want to overwrite it completely.

You can't do that with a mapping, because you can't wipe a mapping clean with 1 instruction

contracts/modules/cover/Cover.sol
```diff
@@ -402,6 +416,21 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+    for (uint i = 0; i < allocationRequests.length; i++) {
```
danoctavian, line 422: the reason i find this loop acceptable is that it's in memory and allowed pools is expected to be small 1,2 items.

shark0der:
1. I would really use a mapping instead of an array. It's more important to reduce the gas cost of the cover buy operation than to be able to override the list of pools for a product. Using a mapping that operation would be O(1). I would leave the task of making sure all allowed pools are erased to the offchain - this would indeed make the operation non-idempotent (same payload might result in a different state because not everything is overwritten), but we could have as a separate parameter (add/remove style array) that would basically make the function call behave deterministic regardless of the current state of individual items in the mapping.

2. Secondly I think we would want to avoid having an array inside the `Product` struct. As far as I know this becomes non-intuitive because whenever you assign the struct it will not copy the array (arrays cannot be assigned to the storage as a whole). If I'm not mistaken might also create issues when calling functions (but I'm really unsure here).

3. Another idea would be to use a single uint to store the staking pool ids. The staking pool id is an `uint40` so we can fit 6 of them in a slot. If we end up needing more than 6 - we probably may end up opening the product for anyone (but better confirm this assumption with Hugh). We can have a custom type to store this (see example in StakingLib). Depending on how often do we read and use it - we might want to store it in a separate mapping and not inside the product struct itself. This would also provide the option to completely override it in one go.

danoctavian:
Problems:

We have the 4 tests failing in `nexus-v2` still failing.

Integration tests are currently disabled. what should we do there?

danoctavian:
Still need to fix those testing issues before merging.

danoctavian:
Tests green now here

test/integration/setup.js
```diff
@@ -308,7 +308,9 @@ async function setup() {
         coverAssets: 0, // Use fallback
         initialPriceRatio: 100,
         capacityReductionRatio: 0,
+        fixedPricing: false,
```
roxdanila, line 311: missed rename

test/integration/setup.js
```diff
@@ -319,7 +321,9 @@ async function setup() {
         coverAssets: 0, // Use fallback
         initialPriceRatio: 100,
         capacityReductionRatio: 0,
+        useFixedPrice: false,
```
roxdanila, line 324: same here

test/unit/Cover/buyCover.js
```diff
@@ -88,6 +88,49 @@ describe('buyCover', function () {
...
+      {
+        coverId: MaxUint256,
+        owner: coverBuyer.address,
+        productId,
+        coverAsset,
+        amount,
+        period,
+        maxPremiumInAsset: expectedPremium,
+        paymentAsset: coverAsset,
+        payWitNXM: false,
```
roxdanila, line 112: `payWitNXM` doesn't exist anymore

test/unit/Cover/buyCover.js
```diff
@@ -88,6 +88,49 @@ describe('buyCover', function () {
...
   });
 
+  it('should purchase new cover with fixed price using 1 staking pool', async function () {
+    const { cover } = this;
+
+    const {
+      members: [coverBuyer],
+    } = this.accounts;
+
+    const { amount, targetPriceRatio, coverAsset, period, expectedPremium, coverId } = buyCoverFixture;
```
roxdanila, line 98: remove coverId and read it after the buyCover tx with `const coverId = await cover.coverData.length`

PR #529, opened 2022-11-28
Test: withdraw unit tests

### LINKED ISSUE #491: StakingPool unit tests: withdraw
**withdraw**
- [ ] reverts if trying to withdraw stake locked in governance
- [ ] allows to withdraw only stake
- [ ] allows to withdraw only rewards
- [ ] allows to withdraw stake and rewards from multiple tranches
- [ ] updates tranches
- [ ] updates global variables totalWithdrawnStake and totalWithdrawnRewards
- [ ] transfers nxm stake and rewards from token controller to nft owner
- [ ] allows to withdraw stake only if tranche is expired
- [ ] anyone can call to withdraw stake and rewards for a token id
- [ ] Should emit some event

paterson1:
## Context

Closes #491 

Added `Withdraw` event to `StakingPool` contract and `withdrawNXMStakeAndRewards` method to `TokenControllerMock`. Also in setup update `master` address for tokenController.

## Test plan

Added unit tests to `StakingPool/withdraw.js`.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/test/unit/StakingPool/withdraw.js
+++ b/test/unit/StakingPool/withdraw.js
@@ -2,15 +2,17 @@ const { BigNumber } = require('ethers');
 const { parseEther } = require('ethers/lib/utils');
 const { ethers, expect } = require('hardhat');
 const { daysToSeconds } = require('../../../lib/helpers');
-const { setEtherBalance, increaseTime, mineNextBlock } = require('../../utils/evm');
+const { increaseTime, mineNextBlock } = require('../../utils/evm');
 const {
   getTranches,
   TRANCHE_DURATION,
   getNewRewardShares,
   calculateStakeAndRewardsWithdrawAmounts,
+  setTime,
+  generateRewards,
 } = require('./helpers');
 
-describe.only('withdraw', function () {
+describe('withdraw', function () {
   const product0 = {
     productId: 0,
     weight: 100,
@@ -27,16 +29,12 @@ describe.only('withdraw', function () {
     ipfsDescriptionHash: 'Description Hash',
   };
 
-  // FIXME: Amount of NXM to be minted before their use as rewards on withdrawal events
-  const nxmRewards = parseEther('10');
-
   const depositToFixture = {
     ...initializeParams,
     amount: parseEther('100'),
     trancheId: 0,
     tokenId: 1,
     destination: ethers.constants.AddressZero,
-    depositNftId: 1,
   };
 
   // Rewards allocation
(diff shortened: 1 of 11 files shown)


test/unit/StakingPool/withdraw.js
```diff
@@ -0,0 +1,573 @@
...
+
+    await expect(
+      stakingPool.connect(user).withdraw([{ tokenId: depositNftId, withdrawStake, withdrawRewards, trancheIds }]),
+    )
+      .to.emit(stakingPool, 'Withdraw')
+      .withArgs(user.address, depositNftId, trancheIds[1], stakes[1], rewards[1])
+      .to.emit(stakingPool, 'Withdraw')
+      .withArgs(user.address, depositNftId, trancheIds[2], stakes[2], rewards[2])
+      .to.emit(stakingPool, 'Withdraw')
+      .withArgs(user.address, depositNftId, trancheIds[2], stakes[2], rewards[2]);
```
paterson1, line 571: This should have index 0, I'm still debugging why is not the same as the emitted args.

patitonar:
> Pushed a failed test.

While reviewing this test I noticed that the total amount of minted rewards was not being distributed to the users and manager, there was a small amount that remained locked in the contract. After debugging I found that the cause of this small amount was a precision loss in the calculation of `accNxmPerRewardsShare`. 

As it is calculated with the following formula `elapsed * _rewardPerSecond / _rewardsSharesSupply` there were a lot of decimals truncated causing the accumulated value to be less than the actual amount to distribute. The solution I thought for this was to calculate and store the accumulator with an increased precision of ONE_NXM and then scale back the value when calculating the amount of the rewards to transfer.

I applied that change in the commit [Increase accNxmPerRewardsShare precision](https://github.com/NexusMutual/smart-contracts/pull/529/commits/f930209d3ec98d6f45411d28e7aad26238a8d9e1)

Let me know your thoughts on this @shark0der

shark0der:
The `accNxmPerRewardsShare` fix looks good to me. In terms of semantics I think the previous value was:
> the accumulated nxm per **1 reward share**

where "1" was actually "1 wei"

The new value represents
> the accmuluated nxm per **1e18 reward shares**

which makes perfect sense and solves the precision loss issue.

shark0der (approved):

shark0der (changes requested):
Right now it seems to me that all tests are using a single staker (unless I missed something). Could you please add a new test where 2+ stakers would deposit and withdraw their stake and rewards? I'm not expecting to discover any issues, but just in case, you know!

roxdanila (approved):

shark0der (approved):

PR #533, opened 2022-11-29
Feature: staking pool allocate-deallocate

shark0der:
## Context

#386 

- implement allocate and deallocate functions in `requestAllocation`
- fix 2 target weight related bugs in staking pool
- fix various other bugs (for ex in TokenController, Cover)
- fix all failing unit tests (non-skipped ones)
- remove IntegrationMockStakingPool and replace with real one
- fix failing integration tests

## Test plan

All non-skipped tests were fixed to pass with the new code.

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/interfaces/ICover.sol
```diff
@@ -176,7 +186,7 @@ interface ICover {
     string calldata ipfsDescriptionHash
   ) external returns (address stakingPoolAddress);
 
-  function isAllowedPool(uint productId, uint poolId) external returns (bool);
+  function isPoolAllowed(uint productId, uint poolId) external returns (bool);
```
danoctavian, line 189: this sounds like a question, not like a isSmth type of predicate

contracts/mocks/StakingPool/SPMockCover.sol
```diff
@@ -56,18 +62,22 @@ contract SPMockCover {
...
         params.productId,
         coverId,
-        params.amount,
         params.period,
-        product.useFixedPrice
-      ),
-      AllocationRequestConfig(
         gracePeriod,
+        product.useFixedPrice,
+        // TODO: figure out if these need to be populated
```
danoctavian, line 77: given that this is a test file, do we still need this TODO here?

contracts/interfaces/ICover.sol
```diff
@@ -48,25 +87,11 @@ struct CoverData {
 struct CoverSegment {
   uint96 amount;
   uint32 start;
-  uint32 period;  // seconds
-  uint16 gracePeriodInDays;
+  uint32 period; // seconds
+  uint32 gracePeriod; // seconds
```
danoctavian, line 91: i'd add a comment how many years this can actually store 

it's like 136 or so since epoch time.

contracts/mocks/TokenControllerMock.sol
```diff
@@ -53,40 +53,46 @@ contract TokenControllerMock is MasterAwareV2 {
...
 
   function depositStakedNXM(address from, uint amount, uint poolId) external {
-    require(msg.sender == address(cover().stakingPool(poolId)), "TokenController: msg.sender not staking pool");
-
     stakingPoolNXMBalances[poolId].deposits += uint128(amount);
     token().operatorTransfer(from, amount);
   }
 
-  function withdrawNXMStakeAndRewards(address to, uint stakeToWithdraw, uint rewardsToWithdraw, uint poolId) external {}
+  function withdrawNXMStakeAndRewards(address to, uint stakeToWithdraw, uint rewardsToWithdraw, uint poolId) external {
```
danoctavian, line 80: the address to param is not used in the mock, comment to eliminate warning

contracts/modules/cover/Cover.sol
```diff
@@ -225,109 +244,124 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+
+        PoolAllocation memory previousPoolAllocation =
+          coverSegmentAllocations[allocationRequest.coverId][segmentId - 1][i];
+
+        // poolAllocationRequests must match the pools in the previous segment
+        require(previousPoolAllocation.poolId == poolAllocationRequests[i].poolId, "Cover: Unexpected pool id");
+
+        vars.previousPremiumInNXM = previousPoolAllocation.premiumInNXM;
+        vars.refund =
+          previousPoolAllocation.premiumInNXM
```
danoctavian, line 329: noting that this is a change in how refund is computed.

We used to express refund in cover Asset (Eg. DAI, ETH etc.)

now it's expressed in NXM

contracts/modules/cover/Cover.sol
```diff
@@ -619,6 +653,23 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     return (1 << coverAsset) & assetsBitMap > 0;
   }
 
+  function isPoolAllowed(uint productId, uint poolId) external view returns (bool) {
+
+    uint poolCount = allowedPools[productId].length;
+
+    if (poolCount == 0) {
+      return true;
```
danoctavian, line 661: i see the intention here but doesn't match what i know to be the spec.

If useFixedPricing, you should check what pools are enabled.

If no pool is registered no pool is allowed.

If you interpret empty list as all allowed, we might as well remove the boolean

contracts/modules/cover/CoverViewer.sol
```diff
@@ -127,74 +127,74 @@ contract CoverViewer {
...
-  ) public view returns (uint) {
-
-    ICover _cover = cover();
-    return staked *
-    _cover.globalCapacityRatio() *
-    (CAPACITY_REDUCTION_DENOMINATOR - capacityReductionRatio) /
-    GLOBAL_CAPACITY_DENOMINATOR /
-    CAPACITY_REDUCTION_DENOMINATOR;
-  }
+//  function getPoolAllocationPriceParametersForProduct(
```
danoctavian, line 130: these are still necessary for the the backend router so we need at least a TODO here to fix things and a clarification.

They are not mandatory for the audit, that is true

contracts/modules/cover/Cover.sol
```diff
@@ -470,7 +504,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
         * burnAmount / segment.amount
         * GLOBAL_CAPACITY_DENOMINATOR / globalCapacityRatio;
 
-      stakingPool(i).burnStake(cover.productId, segment.start, segment.period, burnAmountInNXM);
+      stakingPool(i).burnStake(burnAmountInNXM);
```
danoctavian, line 507: What's the intention here with burnStake?

It would still require productId i'd assume and time interval to know the tranches?

danoctavian (changes requested):
Some changes requested

Rest are questions and clarifications required

contracts/interfaces/ICover.sol
shark0der: We're using uint32 for timestamps for a while now, and not just for (short) periods of time but for actual timestamps (dates). The comment would be redundant

contracts/mocks/StakingPool/SPMockCover.sol
shark0der: Yes, the field are relevant for the cover editing. We might need to set them while unit testing, hence the todo.

contracts/modules/cover/Cover.sol
shark0der: Good point. Will raise with Hugh but it won't be fixed in this PR since it's a spec issue not a code issue.

contracts/modules/cover/Cover.sol
shark0der: Burning staked nxm affects all active tranches regardless of the cover buy time and period.

contracts/modules/cover/Cover.sol
shark0der: We've agreed to implement fixed pricing and pool allowlist as separate features. It doesn't make sense to have a product which is not allowed on any pool - if we want to disable it we have the deprecated flag for that.

contracts/modules/cover/CoverViewer.sol
shark0der: These functions need an overhaul. Due to changed signatures they started to error out and since they need reimplementation I decided to comment them out for the moment.

danoctavian (approved):
LGTM

PR #539, opened 2022-12-05
Chore: Upgrade PriceFeedOracle and Master contract

### LINKED ISSUE #467: Upgrade PriceFeedOracle and Master contracts to solidity 0.8

MilGard91:
## Context

PriceFeedOracle and Master contracts outdated

Closes #467 

Upgrading the PriceFeedOracle, Master and all relative contracts to 0.8.16.

## Test plan

N/A

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/interfaces/IPriceFeedOracle.sol
+++ b/contracts/interfaces/IPriceFeedOracle.sol
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 pragma solidity >=0.5.0;
-pragma experimental ABIEncoderV2;
 
 interface Aggregator {
   function latestAnswer() external view returns (int);
--- a/contracts/interfaces/IPriceFeedOracle.sol
+++ b/contracts/interfaces/IPriceFeedOracle.sol
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 pragma solidity >=0.5.0;
+pragma experimental ABIEncoderV2;
 
 interface Aggregator {
   function latestAnswer() external view returns (int);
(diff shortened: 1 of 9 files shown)


contracts/modules/governance/external/Proxy.sol
```diff
@@ -1,28 +1,44 @@
...
+      let size := returndatasize()
+      returndatacopy(ptr, 0, size)
+
+      switch result
+      case 0 {revert(ptr, size)}
+      default {return (ptr, size)}
+    }
+  }
+
+  receive() external payable {
```
shark0der, line 33: The code looks functionally correct. I compared this Proxy.sol from the OZ v4 we have installed and the only difference is the memory offset at which it starts, 0x40 but we can safely ignore it. The only thing I'd do here is avoid the duplicated code. Please take a look at
`node_modules/@openzeppelin/contracts-v4/proxy/Proxy.sol` and implement a `_delegate` function that would contain the assembly. The `fallback` and `receive` functions should call `_delegate`. Ignore the v4 hooks.

shark0der (changes requested):
Aside from the Proxy comment - LGTM

contracts/interfaces/IPriceFeedOracle.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 pragma solidity >=0.5.0;
+pragma experimental ABIEncoderV2;
```
shark0der, line 4: This was bugging me but just now I realized it's actually not needed. Please remove.

shark0der (approved):

PR #543, opened 2022-12-06
Test: pricing unit tests

### LINKED ISSUE #463: Pricing unit tests
Include a test that an event has been emitted at _buyCover_

kyledewy:
## Context

closes #463 

This PR adds some unit tests for product pricing in the StakingPool. 

## Test plan
Premiums and product price were tested for the extreme highs/lows for cover.amount and cover.period
For cover purchases below NXM_PER_ALLOCATION_UNIT (1e16)  the premium seems to be charged as if 1e16 nxm was sent
There is a test for an underflow while buying a new cover while a previous one has expired, which seems to be fixed by #546 

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


test/unit/StakingPool/calculatePrice.js
```diff
@@ -1,56 +1,212 @@
...
-    assert.equal(price.toString(), expectedPrice.toString());
+const { expect } = require('chai');
+const { ethers } = require('hardhat');
+const { getCurrentTrancheId, calculateBasePrice } = require('./helpers');
+const { mineNextBlock, setNextBlockTime } = require('../../utils/evm');
+const { AddressZero, MaxUint256, WeiPerEther } = ethers.constants;
+const { parseEther } = ethers.utils;
+const daysToSeconds = days => days * 24 * 60 * 60;
+
+describe('calculatePremium unit tests', function () {
```
danoctavian, line 9: just call it calculatePremium

the unit tests post fix is redundant

test/unit/StakingPool/calculatePrice.js
```diff
@@ -1,56 +1,212 @@
...
-    const price = await stakingPool.calculatePrice(amount, basePrice, activeCover, capacity);
-
-    assert.equal(price.toString(), basePrice.toString());
+  async function moveDaysForward(days) {
+    let { timestamp } = await ethers.provider.getBlock('latest');
+    timestamp += daysToSeconds(days.toString());
+    await setNextBlockTime(timestamp);
+    await mineNextBlock();
+  }
+
```
danoctavian, line 92: as discussed in DMs, let's explore some extreme cases.

Example buying all capacity in 1 go from to 100%.

Anything that has values of 0 or 100% should be explored.

danoctavian:
Looking through the code now there's 2 functions:

calculatePremiumPerYear - pure function

getPremium - stateful and side-effectful function



https://github.com/NexusMutual/smart-contracts/blob/9380443697465f9cb6feb3459f5cfcebdde862c8/contracts/modules/staking/StakingPool.sol#L1523

The only side-effectful line is where it writes out the new product state.


That can be externalized as return values (newProduct)

The only 2 values that are read are: block.timestamp and product.
 I would argue for this to be refactored as a pure function, to simplify testing.(Eg. block.timestamp, product passed in as arguments.)
 
 This greatly simplifies testing, all of it is against pure functions and is a lot nicer overall.

@shark0der  @kyledewy  let me know your thoughts.

contracts/modules/staking/StakingPool.sol
```diff
@@ -1468,26 +1468,60 @@ contract StakingPool is IStakingPool, ERC721 {
     uint totalCapacity,
     uint globalMinPrice,
     bool useFixedPrice
-  ) internal returns (uint) {
+  ) internal returns (uint premium) {
 
     StakedProduct memory product = products[productId];
     uint targetPrice = Math.max(product.targetPrice, globalMinPrice);
```
danoctavian, line 1474: i could argue for putting in this max() calculation inside the pure function as well and pass in globalMinPrice

but it's trivial, so this looks great as well

shark0der:
General note regarding `expected` vs `actual` value in tests: the first parameter to `expect` is the actual value, then at the end inside `.to.be.equal()` pass the expected value.

As an example, this code:
```
      const expectedPremium = buyCoverParamsTemplate.amount
        .mul(product.targetPrice)
        .div(INITIAL_PRICE_DENOMINATOR)
        .div(periodsInYear);
      expect(expectedPremium).to.be.equal(await cover.lastPremium());
```

should to be:
```
      const actualPremium = await cover.lastPremium();
      const expectedPremium = buyCoverParamsTemplate.amount
        .mul(product.targetPrice)
        .div(INITIAL_PRICE_DENOMINATOR)
        .div(periodsInYear);
      expect(actualPremium).to.be.equal(expectedPremium);
```

test/unit/StakingPool/calculatePrice.js
```diff
@@ -1,56 +1,468 @@
...
+const { expect } = require('chai');
+const { ethers } = require('hardhat');
+const { getCurrentTrancheId, calculateBasePrice, calculateSurgePremium, calculatePriceBump } = require('./helpers');
+const { mineNextBlock, setNextBlockTime } = require('../../utils/evm');
+const { daysToSeconds } = require('../../../lib/helpers');
+const { AddressZero, MaxUint256, WeiPerEther } = ethers.constants;
+const { parseEther } = ethers.utils;
+const { BigNumber } = ethers;
 
 describe('calculatePrice', function () {
```
danoctavian, line 10: rename file and test to match the new `calculatePremium`

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,480 @@
...
+const { ethers } = require('hardhat');
+const { getCurrentTrancheId, calculateBasePrice, calculateSurgePremium, calculatePriceBump } = require('./helpers');
+const { mineNextBlock, setNextBlockTime } = require('../../utils/evm');
+const { daysToSeconds } = require('../../../lib/helpers');
+const { AddressZero, MaxUint256, WeiPerEther } = ethers.constants;
+const { parseEther } = ethers.utils;
+const { BigNumber } = ethers;
+
+describe('calculatePremium', function () {
+  const periodInDays = 91.25;
```
shark0der, line 11: Is there any particular reason for choosing this period? A tranche has 91 days exactly if that's what the aim was.

test/unit/StakingPool/helpers.js
```diff
@@ -29,6 +29,61 @@ const setTime = async timestamp => {
   await mineNextBlock();
 };
 
+async function calculateBasePrice(stakingPool, product, priceChangePerDay) {
```
shark0der, line 32: Try making the functions as stateless as possible - i.e. avoid making them dependant on the current time or other state. Instead the needed state as arguments.
- It is not clear from the title that the function depends on the current time. If I was working on a test that uses this function I might move it within the test and not be aware that it should be called at a particular time.
- The `stakingPool` parameter is not used at all.

test/unit/StakingPool/helpers.js
```diff
@@ -29,6 +29,61 @@ const setTime = async timestamp => {
...
+  }
+
+  if (initialCapacityUsed.lte(surgeStartPoint)) {
+    return BigNumber.from(0);
+  }
+
+  return initialCapacityUsed.sub(surgeStartPoint);
+}
+
+// Note fn expects coverAmount is rounded down by NXM_PER_ALLOCATION_UNIT
```
shark0der, line 81: Comment seems wrong. `NXM_PER_ALLOCATION_UNIT` is the smallest unit of allocation the staking pool works with. Its value is 0.01 nxm. If anyone wants to allocate less than that - it is rounded UP to it (not down).

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,480 @@
...
+      buyCoverParamsTemplate.productId,
+      GLOBAL_CAPACITY_RATIO,
+      coverProductTemplate.capacityReductionRatio,
+    );
+    const expectedPremium = buyCoverParamsTemplate.amount
+      .mul(initialPrice)
+      .div(INITIAL_PRICE_DENOMINATOR)
+      .div(periodsInYear);
+    const priceBump = await calculatePriceBump(
+      buyCoverParamsTemplate.amount.div(NXM_PER_ALLOCATION_UNIT),
```
shark0der, line 184: This will lose precision.

shark0der: Probably not an issue here because it's an exact multiple of allocation unit.

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,480 @@
...
+    const calculatedPremium = await stakingPool.calculatePremium(
+      product,
+      buyCoverParams.period,
+      amount.div(NXM_PER_ALLOCATION_UNIT), // cover amount
+      0, // initialCapacityUsed
+      totalCapacity,
+      product.targetPrice, // targetPrice
+      timestamp, // current block timestamp
+    );
+    const expectedPremium = basePremium.add(surgePremium).sub(surgePremiumSkipped).div(periodsInYear);
```
shark0der, line 284: Since it's a single purchase I'd expect `surgePremiumSkipped` to be zero. Remove it from here and add a test to make sure it's zero.

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,480 @@
...
+    );
+
+    // 1st cover buy
+    {
+      await moveDaysForward(183);
+
+      const product = await stakingPool.products(productId);
+      const basePrice = await calculateBasePrice(stakingPool, product, PRICE_CHANGE_PER_DAY);
+      expect(basePrice).to.be.equal(200);
+      const premiumPerYear = await stakingPool.calculatePremiumPerYear(
```
shark0der, line 335: What you are testing here is not the premium calculation but rather the fact that the staking pool returns the result of the function correctly to the cover contract. If the `calculatePremiumPerYear` would return a crazy value this test would pass without noticing.

shark0der (changes requested):
Changes requested

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,480 @@
...
+    const priceBump = await calculatePriceBump(
+      buyCoverParamsTemplate.amount.div(NXM_PER_ALLOCATION_UNIT),
+      this.config.PRICE_BUMP_RATIO,
+      totalCapacity,
+    );
+
+    {
+      // buy cover and check premium + new price
+      const buyCoverParams = { ...buyCoverParamsTemplate, period: daysToSeconds(365) };
+      await cover.allocateCapacity(buyCoverParams, coverId, stakingPool.address);
```
danoctavian, line 125: the test here is

`calculatePremium`

we have a function now called just that which is pure.

The tests in this file shoul concern just that function.

allocateCapacity should be tested separately.

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,480 @@
...
+    const basePremium = amount.mul(basePrice).div(INITIAL_PRICE_DENOMINATOR);
+    const { surgePremium, surgePremiumSkipped } = await calculateSurgePremium(
+      stakingPool,
+      amount,
+      0 /* activeAllocations */,
+      totalCapacity,
+      this.config,
+    );
+    const { timestamp } = await ethers.provider.getBlock('latest');
+    const calculatedPremium = await stakingPool.calculatePremium(
```
danoctavian, line 275: all numerical tests should focus on calculatePremium and exhaust the test cases there.

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,480 @@
...
+      buyCoverParams.period,
+      amount.div(NXM_PER_ALLOCATION_UNIT), // cover amount
+      0, // initialCapacityUsed
+      totalCapacity,
+      product.targetPrice, // targetPrice
+      timestamp, // current block timestamp
+    );
+    const expectedPremium = basePremium.add(surgePremium).sub(surgePremiumSkipped).div(periodsInYear);
+    expect(calculatedPremium.premium).to.be.equal(expectedPremium);
+    await cover.connect(coverBuyer).allocateCapacity(buyCoverParams, coverId, stakingPool.address);
```
danoctavian, line 286: The tests shouldn't have an entry point with `allocateCapacity` even for the `requestAllocation` because they are unit tests.

This looks more like an integration test, even though it isn't.

All entry points should be through StakingPool methods

I would have another file that tests `requestAllocation`. 

But againt calculatePremium tests should *only* test function with the same name.

danoctavian (changes requested):
Left a comment above in the over-arching issue.

test/unit/StakingPool/calculatePremium.js
kyledewy: Made it a multiple of 365 days to make calculating the `premium` from `premiumForYear` easier.

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,352 @@
...
+      initialCapacityUsed,
+      totalCapacity,
+    );
+    const { surgePremium, surgePremiumSkipped } = calculateSurgePremiums(
+      coverAmountRaw,
+      initialCapacityUsed,
+      totalCapacity,
+      this.config,
+    );
+    // Note must use rounded down value to match precision loss in contracts
```
shark0der, line 103: Where exactly are we rounding down?

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,352 @@
...
+      stakingPool.calculatePremium(
+        stakedProduct,
+        period,
+        coverAmount,
+        initialCapacityUsed,
+        totalCapacity,
+        stakedProduct.targetPrice,
+        timestamp,
+      ),
+    ).to.be.revertedWithPanic(17); // divide-by-zero (0x11)
```
danoctavian, line 64: make a constant for this 17

shark0der: `0x11` is underflow/overflow

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,352 @@
...
+    await expect(
+      stakingPool.calculatePremiumPerYear(basePrice, coverAmount, initialCapacityUsed, totalCapacity),
```
danoctavian, line 66: we shouldn't test 2 functions in the same test file when dealing with unit tests.

Unit tests should normally focus on 1 function.

calculatePremiumPerYear is a different function.

If it helps to write unit tests for calculatePremiumPerYear, write them in a separate file.

I could argue it's enough to test `calculatePremium` and that would be enough.


If you want to bring an argument that calculatePremiumPerYear should be tested inidivdually as well because of complexity i could accept that.

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,352 @@
+const { expect } = require('chai');
+const { ethers } = require('hardhat');
+const { parseEther } = ethers.utils;
+const { BigNumber } = ethers;
+const { calculateSurgePremiums, calculateBasePrice, divCeil } = require('./helpers');
+const { daysToSeconds } = require('../../../lib/helpers');
+
+describe('calculatePremium', function () {
```
danoctavian, line 8: i could be missing it but i can't see the test case for:

`initialCapacityUsed < surgeStartPoint & finalCapacityUsed < surgeStartPoint`

test/unit/StakingPool/calculatePremium.js
kyledewy: The comment is confusing, so I removed it. I was referring to the fact that the `basePremium` calculation has to use `coverAmount.divCeil(NXM_PER_ALLOCATION_UNITS)` and then multiply that by `NXM_PER_ALLOCATION_UNITS` instead of using the original cover amount.

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,352 @@
...
+        coverAmount,
+        initialCapacityUsed,
+        totalCapacity,
+        stakedProduct.targetPrice,
+        timestamp,
+      ),
+    ).to.be.revertedWithPanic(18); // divide-by-zero (0x11)
+    await expect(
+      stakingPool.calculatePremiumPerYear(basePrice, coverAmount, initialCapacityUsed, totalCapacity),
+    ).to.be.revertedWithPanic(18); // divide-by-zero (0x11)
```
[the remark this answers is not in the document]
kyledewy: Sorry forgot to delete the comment. Its 18 -> 0x12

test/unit/StakingPool/calculatePremium.js
kyledewy: Ok I think it's ok to leave out premiumPerYear :ok_hand:

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,611 @@
...
+        NXM_PER_ALLOCATION_UNIT,
+        INITIAL_PRICE_DENOMINATOR,
+      );
+      const [premium, product] = await stakingPool.calculatePremium(
+        stakedProduct,
+        period,
+        coverAmount,
+        initialCapacityUsed,
+        totalCapacity,
+        basePrice,
```
danoctavian, line 60: basePrice here is passed for what the function needs as targetPrice

Was this the intention here?

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,611 @@
...
+
+    // 1st cover buy
+    {
+      const coverAmountRaw = parseEther('2400');
+      const coverAmount = divCeil(coverAmountRaw, NXM_PER_ALLOCATION_UNIT);
+      const initialCapacityUsed = BigNumber.from(0);
+      const basePrice = calculateBasePrice(timestamp, stakedProduct, PRICE_CHANGE_PER_DAY);
+      expect(basePrice).to.be.equal(200);
+      const priceBump = calculatePriceBump(coverAmount, PRICE_BUMP_RATIO, totalCapacity);
+      expect(priceBump).to.be.equal(96);
```
danoctavian, line 45: i understand why the javascript functions were introduced - to kind of show how values were derived

The issue as i think @shark0der  mentioned is that you can have the same bug in solidity and in js and gloss over

For this test it's matching the values shown by Hugh in the sheet, we can hardcode the values here as i can see you partially did with these extra assertions


From a test writing point of view, it's not exactly a test to assert the result of a javascript function that's used for testing. you are testing the testing code at that point

I can see this test written with hardcoded values from Hugh's sheet, instead of of the testing function

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,611 @@
...
+    const { INITIAL_PRICE_DENOMINATOR, NXM_PER_ALLOCATION_UNIT } = this.config;
+    // call staking pool and calculate premium
+    const stakedProduct = { ...stakedProductTemplate, nextPriceUpdateTime: timestamp };
+    const period = daysToSeconds(365);
+    const coverAmountRaw = BigNumber.from(2).pow(96);
+    const coverAmount = divCeil(coverAmountRaw, NXM_PER_ALLOCATION_UNIT);
+    const initialCapacityUsed = 0;
+    const totalCapacity = coverAmount;
+    const basePrice = calculateBasePrice(timestamp, stakedProduct, this.config.PRICE_CHANGE_PER_DAY);
+    const { premium: expectedPremium } = await stakingPool.calculatePremium(
```
danoctavian, line 305: the naming here is the other way around

expectedPremium should be a value you compute in your test code;

what comes out of `stakingPool.calculatePremium` is the `actualPremium` - what is being tested.

`actualPremium` what the function under test (solidity) says the result is
`expectedPremium` what the test code expects the function being tested to return

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,611 @@
...
+    const coverAmountRaw = parseEther('1234');
+    const coverAmount = divCeil(coverAmountRaw, NXM_PER_ALLOCATION_UNIT);
+
+    const totalCapacity = coverAmount;
+    const surgeStartPoint = totalCapacity
+      .mul(this.config.SURGE_THRESHOLD_RATIO)
+      .div(this.config.SURGE_THRESHOLD_DENOMINATOR);
+    const initialCapacityUsed = surgeStartPoint;
+    const basePrice = calculateBasePrice(timestamp, stakedProduct, this.config.PRICE_CHANGE_PER_DAY);
+    const { premium: expectedPremium } = await stakingPool.calculatePremium(
```
danoctavian, line 344: same here namings are the other way around

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,611 @@
...
+    const { INITIAL_PRICE_DENOMINATOR, NXM_PER_ALLOCATION_UNIT } = this.config;
+    const stakedProduct = { ...stakedProductTemplate, nextPriceUpdateTime: timestamp };
+    const period = daysToSeconds(365);
+    const coverAmountRaw = 1;
+    const coverAmount = divCeil(coverAmountRaw, NXM_PER_ALLOCATION_UNIT);
+
+    const totalCapacity = coverAmount.mul(100);
+    const initialCapacityUsed = 0;
+    const basePrice = calculateBasePrice(timestamp, stakedProduct, this.config.PRICE_CHANGE_PER_DAY);
+    const { premium: expectedPremium } = await stakingPool.calculatePremium(
```
danoctavian, line 438: here too

test/unit/StakingPool/calculatePremium.js
kyledewy: Oh nice catch. Should be targetPrice instead..

contracts/modules/staking/StakingPool.sol
```diff
@@ -1618,10 +1618,8 @@ contract StakingPool is IStakingPool, ERC721 {
     //              = amountOnSurge * SURGE_PRICE_RATIO * coverToCapacityRatio / 2
     //              = amountOnSurge * SURGE_PRICE_RATIO * amountOnSurge / totalCapacity / 2
 
-    uint surgePremium = amountOnSurge * SURGE_PRICE_RATIO * amountOnSurge / totalCapacity / 2;
+    uint surgePremium = amountOnSurge * amountOnSurge * NXM_PER_ALLOCATION_UNIT / totalCapacity;
```
kyledewy, line 1621: Note that I simplified the calculation here a bit. Left the JS implementation as it was before.

shark0der: The result of the function is the same because you removed  `* SURGE_PRICE_RATIO` which is basically `2 (1e18)` and the `/ 2` at the end which along with the allocation unit constants were "canceling each other out". The removal of the surge price ratio is a big no-no here as that is the pricing multiplier for the surge pricing. The `/ 2` that you removed is there because after calculating the area of the rectangle you divide by 2 to get the area of the triangle. If the surge price ratio was `3e18` instead, this would have failed. Pleaes revert these changes.

kyledewy: Yea I can understand that it makes it less readable and maintainable. Its refactored specifically because surge price ratio is `2e18`, if we are planning to change it, we would have to bring back the division by 2. It's too bad the compiler doesn't recognize the redundant calculations here though. Would it make sense to instead just change `SURGE_PRICE_RATIO` to `2e16`, keep the div by 2, and remove the final division by 100? This way if surge price ratio changes, it should be ok?

danoctavian: I agree with @shark0der  that the gains in gas are not material so i'd rather keep things clear from a mathematical point of view.

It's easier to refactor the code and update the formula and reason about it when the math matches the formula in solidity

danoctavian: surprised the compiler doesn't do things here even with optimizer on

kyledewy: Yea it's too bad. Anyways, I reverted the changes I made.

test/unit/StakingPool/requestAllocation.js
```diff
@@ -0,0 +1,488 @@
...
+  divCeil,
+  roundUpToNearestAllocationUnit,
+} = require('./helpers');
+const { mineNextBlock, setNextBlockTime } = require('../../utils/evm');
+const { daysToSeconds } = require('../../../lib/helpers');
+const { AddressZero, MaxUint256 } = ethers.constants;
+const { parseEther } = ethers.utils;
+const { BigNumber } = ethers;
+
+describe('requestAllocation', function () {
```
danoctavian, line 17: i know this ticket focuses on pricing values

requestAllocation is a fairly complex function that has a lot of effects 

We need tests for all those 

`updateStoredAllocations`

the // add new rewards section

// remove previous rewards

not sure if we already have a ticket for this

test/unit/StakingPool/calculatePremium.js
```diff
@@ -0,0 +1,624 @@
...
+  calculateBasePrice,
+  divCeil,
+  calculateFixedPricePremium,
+  calculatePriceBump,
+} = require('./helpers');
+const { daysToSeconds } = require('../../../lib/helpers');
+
+describe('calculatePremium', function () {
+  // If an arithmetic operation results in underflow or overflow outside of an unchecked { ... } block.
+  const UNDER_OR_OVERFLOW = 0x11;
```
danoctavian, line 16: i'd look to see if ethers already has this if not we can also put them in our common utils lib

contracts/modules/staking/StakingPool.sol
```diff
@@ -1598,11 +1598,13 @@ contract StakingPool is IStakingPool, ERC721 {
...
     //              = amountOnSurge * SURGE_PRICE_RATIO * amountOnSurge / totalCapacity / 2
 
-    uint surgePremium = amountOnSurge * SURGE_PRICE_RATIO * amountOnSurge / totalCapacity / 2;
-
     // amountOnSurge has two decimals
     // dividing by ALLOCATION_UNITS_PER_NXM (=100) to normalize the result
-    return surgePremium / ALLOCATION_UNITS_PER_NXM;
+    uint surgePremiumOriginal = amountOnSurge * SURGE_PRICE_RATIO * amountOnSurge / totalCapacity / 2 / ALLOCATION_UNITS_PER_NXM;
+    uint surgePremium = amountOnSurge * amountOnSurge * NXM_PER_ALLOCATION_UNIT / totalCapacity;
+    assert(surgePremiumOriginal == surgePremium);
```
danoctavian: what's the reason to have those 2 computations?

Are we going to do a cleanup here?

test/unit/StakingPool/requestAllocation.js
kyledewy: Yea agreed. I couldn't find a ticket for it yet, so I made one here: https://github.com/NexusMutual/smart-contracts/issues/582
Feel free to add anything I missed.

test/unit/StakingPool/calculatePremium.js
kyledewy: Couldn't find them in ethers. Added them all to `utils/errors.js`

contracts/modules/staking/StakingPool.sol
kyledewy: This should be cleaned up already. Is it possible you are looking at an older commit here?

danoctavian (approved):
LGTM

shark0der (approved):

PR #570, opened 2022-12-13
Feature: staking pool depositTo/withdraw/extend without arrays

shark0der:
## Context

The array parameters unnecessary complicate the logic, increasee the code size and if the functions will be mostly used for a single deposit - they'll consume more gas.
Issue: #557 

This PR simplifies this logic and makes the function act on a single deposit. There's also the new `multicall` function which adds support for performing multiple operations in a single tx if there's a need for that at any point.

## Test plan

Fixed all the tests that started to fail because of it. There's a single test that used 2 deposit in a single tx and I've used multicall for that. 

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/interfaces/IStakingPool.sol
+++ b/contracts/interfaces/IStakingPool.sol
@@ -21,20 +21,6 @@ struct AllocationRequest {
   uint globalMinPrice;
 }
 
-struct WithdrawRequest {
-  uint tokenId;
-  bool withdrawStake;
-  bool withdrawRewards;
-  uint[] trancheIds;
-}
-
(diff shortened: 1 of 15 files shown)


danoctavian:
I don't have full context as to why this change is necessary.

Is it contract size ?

the function with array is more flexible

danoctavian (changes requested):
The code looks functionally correct.

But the PR requires the description filled to explain why this is required in with the tickboxes and all.

danoctavian (approved):
lgtm

PR #573, opened 2022-12-13
Feature: Make upgradeCapitalPool copy over investmentAssets and coverAssets when pool is upgraded

### LINKED ISSUE #473: Tokenomics: Change the Pool.upgradeCapitalPool to copy over investmentAssets and coverAssets when pool is upgraded
Currently coverAssets and investmentAssets for the Pool.sol contract are constructed in the constructor() each time.
We could copy them over when the pool is upgraded.

kyledewy:
## Context

closes #473 

This PR changes the capital pool so when it is deployed, it copies over `coverAssets`, `investmentAssets` and `swapDetails` from the previous pool. It also also adds the option to add additional assets when deploying a new Pool.

## Test plan
A test was added to `unit/Pool/upgradeCapitalPool.js`. I was going to create a new unit test file `constructor.js` and add more tests, but the tests are still using truffle/web3 and I was having issues testing the reverts in the constructor. 

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/capital/Pool.sol
+++ b/contracts/modules/capital/Pool.sol
@@ -65,6 +65,8 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
 
   uint16 constant MAX_SLIPPAGE_DENOMINATOR = 10000;
 
+  address immutable public previousPool;
+
   /* events */
   event Payout(address indexed to, address indexed assetAddress, uint amount);
   event NXMSold (address indexed member, uint nxmIn, uint ethOut);
@@ -81,12 +83,12 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
     address _master,
     address _priceOracle,
     address _swapOperator,
-    address DAIAddress,
-    address stETHAddress
+    address _previousPool
   ) {
     master = INXMMaster(_master);
     priceFeedOracle = IPriceFeedOracle(_priceOracle);
     swapOperator = _swapOperator;
+    previousPool = _previousPool;
 
     // [todo] After this contract is deployed it might be worth modifying upgradeCapitalPool to
     // copy the assets on future upgrades instead of having them hardcoded in the constructor.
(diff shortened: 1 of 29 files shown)


contracts/modules/capital/Pool.sol
```diff
@@ -389,16 +420,32 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
     // Transfer investment assets.
-    uint investmentAssetsCount = investmentAssets.length;
     for (uint i = 0; i < investmentAssetsCount; i++) {
       _transferEntireAssetBalance(investmentAssets[i].assetAddress, newPoolAddress);
+      _swapDetails[i + coverAssetsCount] = swapDetails[investmentAssets[i].assetAddress];
+      _assets[i + coverAssetsCount] = investmentAssets[i];
+      isCoverAsset[i + coverAssetsCount] = false;
     }
+
+    IPool(newPoolAddress).receiveAssets(_assets, _swapDetails, isCoverAsset);
```
danoctavian, line 447: is it possible to just pass in the storage variables here or you need to construct all that stuff in memory first?

danoctavian:
At a feature spec level, this implementation makes things a little bit less flexible.

With the constructor variables we can rebuild the arrays everytime and add new things in on a need basis or deprecate etc.

Here we need to upgrade the pool and the run another governance update to add or deprecate an asset.

I'm not convinced it's ideal to have it, it does heavily reduce the probability of error so that's the big upside.

Putting this out there.

test/unit/Pool/setup.js
```diff
@@ -95,6 +102,28 @@ async function setup() {
...
+    });
+  }
+  {
+    // Set stEth asset
+    const minAmount = parseEther('24360');
+    const maxAmount = parseEther('32500');
+    const maxSlippageRatio = 0;
+    await pool.addAsset(stETH.address, decimals, minAmount, maxAmount, maxSlippageRatio, false, {
+      from: accounts.governanceContracts[0],
+    });
```
danoctavian, line 124: there was USDC going into cover assets as well

contracts/modules/capital/Pool.sol
```diff
@@ -389,16 +420,32 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
 
     // Transfer cover assets. Start from 1 (0 is ETH)
-    uint coverAssetsCount = coverAssets.length;
-    for (uint i = 1; i < coverAssetsCount; i++) {
+    uint coverAssetsCount = coverAssets.length - 1;  // substract 1 as ETH is skiped
+    uint investmentAssetsCount = investmentAssets.length;
+    uint totalAssetCount = coverAssetsCount + investmentAssetsCount;
+
+    SwapDetails[] memory _swapDetails = new SwapDetails[](totalAssetCount);
+    Asset[] memory _assets = new Asset[](totalAssetCount);
```
danoctavian, line 428: feels like passing in the 2 arrays makes things simpler.

On the other side in receiveAssets they materialize in 2 arrays again so this normalizing into 1 array feels like unnecessary complexity since it adds an extra array anyway (isCoverAsset)

test/unit/Pool/removeAsset.js
```diff
@@ -159,6 +159,7 @@ describe('removeAsset', function () {
       assert.strictEqual(minAmount.toString(), ether('24360').toString());
       assert.strictEqual(maxAmount.toString(), ether('32500').toString());
       assert.strictEqual(maxSlippageRatio.toString(), '0');
+      // TODO: does this need to be exactly this timestamp?
       assert.strictEqual(lastSwapTime.toString(), '1633425218');
```
kyledewy, line 163: The `lastSwapTime` is no longer manually set in the constructor, but instead goes through `addAsset()`, which sets lastSwapTIme to 0. I didn't make any changes for this as I wanted to ask if this exact timestamp is still required?

contracts/modules/capital/Pool.sol
```diff
@@ -202,6 +204,31 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
     return swapDetails[assetAddress];
   }
 
+  function getAssetsAndSwapDetails() external view returns (
```
kyledewy, line 207: Should we exclude abandoned assets from being copied over? I could change this function so it only returns active assets.

contracts/modules/capital/Pool.sol
```diff
@@ -389,7 +448,9 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
     require(ok, "Pool: Transfer failed");
 
     // Transfer cover assets. Start from 1 (0 is ETH)
-    uint coverAssetsCount = coverAssets.length;
+    uint coverAssetsCount = coverAssets.length;  // substract 1 as ETH is skiped
```
danoctavian, line 451: there's no substraction here the comment is not relevant imo, it's just ETH skipped at the start

danoctavian:
we can use some unit tests for the constructor itself for Pool.sol now since it's relatively complex.

danoctavian (changes requested):
Small changes, and a request for unit testing the constructor

kyledewy:
Note that this PR requires this one to be merged first: https://github.com/NexusMutual/smart-contracts/pull/584
Tests are converted to ethers & there are bug fixes for the broken tests there

shark0der:
As per discussion, the current implementation will fail to upgrade because the `previousPool.getAssetsAndSwapDetails()` call will fail. Will close this task and we'll get back to it post V2 launch.

PR #583, opened 2022-12-20
Fix: 5.1.3 Premium never sent to Pool

### LINKED ISSUE #560: 5.1.3 Premium never sent to Pool
When a new cover is bought the retrieved assets should be sent to the pool. Currently this is happening for tokens but was missed from the implementation (ETH remains stuck in the Cover contract) and it wasn't caught by tests either.

kyledewy:
## Context
closes #560 

This PR addresses a bug where the ETH premium wasn't being sent to the capital pool when buying a cover with ETH as payment asset.

## Test plan

Added checks that ETH is transferred in already existing unit tests `unit/buyCover.js`

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -414,11 +414,10 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
 
       uint remainder = msg.value - premiumWithCommission;
 
-      if (remainder > 0) {
-        // solhint-disable-next-line avoid-low-level-calls
-        (bool ok, /* data */) = address(msg.sender).call{value: remainder}("");
-        require(ok, "Cover: Returning ETH remainder to sender failed.");
-      }
+      // send premium in eth to the pool
+      // solhint-disable-next-line avoid-low-level-calls
+      (bool ok, /* data */) = address(_pool).call{value: premiumInPaymentAsset}("");
+      require(ok, "Cover: Sending ETH to pool failed.");
 
       // send commission
       if (commission > 0) {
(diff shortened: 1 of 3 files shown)


contracts/modules/cover/Cover.sol
```diff
@@ -425,7 +425,12 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
         require(ok, "Cover: Sending ETH to commission destination failed.");
       }
 
-      // TODO: send eth to pool
+      // send premium in eth to the pool
+      if (premiumInPaymentAsset > 0) {
```
shark0der, line 429: The `premiumInPaymentAsset` is already checked at line 294 so the if statement here seems redundantant. Please check and confirm.

Secondly, I'm wondering whether it's worth moving this transfer before sending the remainder to the `msg.sender`. Thinking of re-entrnacy here but I think it's very unlikely to do anything realistic here. Thoughts? cc @danoctavian

kyledewy: Yup, the if statement seems to be redundant, will remove it.. The second point seems like a good idea to me, even if the risk is low.

danoctavian: @shark0der we used the nonReentrant guard for buyCover to allow flexibility in how we structure code in `buyCover` and prioritize avoiding stack to deep and readability instead of preventing re-entrancy attacks through ordering.

I think nonReentrant guard is sufficient

danoctavian:
let's add an assertion for the premium moving to the Pool in one of the integration tests as all.

It is simplistic but it is an "integration thing" as the assets are moving there.

danoctavian (changes requested):
small change to add an extra test assertion

danoctavian (approved):
LGTM

PR #584, opened 2022-12-22
Feature: Update all capital pool unit tests to use ethers instead of truffle/web3

kyledewy:
## Context

No ticket number for this one.

I noticed the pool unit tests were still using truffle/web3 so I updated them before writing more tests in the upgrade-capital-pool PR

## Test plan
No tests were added

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/capital/Pool.sol
+++ b/contracts/modules/capital/Pool.sol
@@ -448,7 +448,7 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
     require(ok, "Pool: Transfer failed");
 
     // Transfer cover assets. Start from 1 (0 is ETH)
-    uint coverAssetsCount = coverAssets.length;  // substract 1 as ETH is skiped
+    uint coverAssetsCount = coverAssets.length;
 
     // Skip ETH
     for (uint i = 1; i < coverAssetsCount; i++) {
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1,34 +1,34 @@
 const { hex } = require('./helpers');
 
 const StakingUintParamType = {
-  MIN_STAKE: hex('MIN_STAK'),
-  MAX_EXPOSURE: hex('MAX_EXPO'),
-  MIN_UNSTAKE: hex('MIN_UNST'),
-  UNSTAKE_LOCK_TIME: hex('UNST_LKT'),
+  MIN_STAKE: hex('MIN_STAK'.padEnd(8, '\0')),
+  MAX_EXPOSURE: hex('MAX_EXPO'.padEnd(8, '\0')),
+  MIN_UNSTAKE: hex('MIN_UNST'.padEnd(8, '\0')),
+  UNSTAKE_LOCK_TIME: hex('UNST_LKT'.padEnd(8, '\0')),
 };
 
 const PoolUintParamType = {
-  minPoolEth: hex('MIN_ETH'),
+  minPoolEth: hex('MIN_ETH'.padEnd(8, '\0')),
 };
 
 const PoolAddressParamType = {
-  swapOperator: hex('SWP_OP'),
-  priceFeedOracle: hex('PRC_FEED'),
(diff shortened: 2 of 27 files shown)


lib/constants.js
```diff
@@ -1,34 +1,34 @@
...
 
 const PoolUintParamType = {
-  minPoolEth: hex('MIN_ETH'),
+  minPoolEth: hex('MIN_ETH'.padEnd(8, '\0')),
 };
 
 const PoolAddressParamType = {
-  swapOperator: hex('SWP_OP'),
-  priceFeedOracle: hex('PRC_FEED'),
+  swapOperator: hex('SWP_OP'.padEnd(8, '\0')),
```
kyledewy, line 15: This was reverting without the 0 padding. I wasn't sure if the other ones were ok, so I added it there also.

contracts/modules/capital/Pool.sol
```diff
@@ -389,7 +448,9 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
     require(ok, "Pool: Transfer failed");
 
     // Transfer cover assets. Start from 1 (0 is ETH)
-    uint coverAssetsCount = coverAssets.length;
+    uint coverAssetsCount = coverAssets.length;  // substract 1 as ETH is skiped
```
danoctavian, line 451: typo skiped -> skipped

Also you are starting with an index of 1 below so it seems to me you shouldn't substract for ETH to be skipped. and you actually don't so the comment is misleading

contracts/modules/capital/Pool.sol
```diff
@@ -77,44 +78,45 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
     _;
   }
 
+
   constructor (
```
danoctavian, line 82: the constructor is quite complex now.

It deserves its own unit test at this point

danoctavian (changes requested):
small change requests, mainly to add a dedicated constructor unit test

contracts/modules/capital/Pool.sol
kyledewy: Oh nice catch! Yup that comment didn't get deleted from the old code. Will fix that today.

contracts/modules/capital/Pool.sol
kyledewy: Yea I agree on this point. I wanted to update to ethers before writing the tests though, so maybe I will add them to the PR this is merged into?

test/unit/StakingPool/depositTo.js
```diff
@@ -474,7 +474,7 @@ describe('depositTo', function () {
 
     const { firstActiveTrancheId, maxTranche } = await getTranches(daysToSeconds(0), DEFAULT_GRACE_PERIOD);
 
-    const tranches = Array(maxTranche - firstActiveTrancheId + 1)
+    const tranches = Array(maxTranche - firstActiveTrancheId)
```
kyledewy, line 477: Not sure why this broke during this PR, but the StakingPool was reverting on this test when using 8 tranches. The js variable `firstActiveTrancheId` isn't the current firstActiveTrancheId in the contract, but is the first one available for this deposit, relative to `period` and `gracePeriod`. It was iterating past the max active tranches for this reason.

contracts/modules/capital/Pool.sol
danoctavian: oh yeah sure where it make sense

test/utils/errors.js
```diff
@@ -0,0 +1,45 @@
+// Used for generic compiler inserted panics.
+const COMPILER = 0x00;
```
danoctavian, line 2: assuming these were not defined anywhere in ethers

great work on having them here

danoctavian (approved):
LGTM good cleanup

some minor comment

contracts/modules/capital/Pool.sol
danoctavian: let's do this in the other PR right

PR #599, opened 2023-01-05
Feature: Implement bucket system for active cover expiry

### LINKED ISSUE #387: Implement a buckets system for active cover amount tracking instead of per-cover
We should keep track of the total cover for MCR purposes. Cover amounts will be grouped by cover asset. Each asset will need its buckets that will track the amounts to expire in the future. Every time we roll from one bucket to another we need to apply the expiration from the new bucket - that's done when buying/editing a cover. The applied expirations are only done for the cover asset of the cover that's being currently bought/edited (don't loop through all assets on cover buy). When reading the current active cover amount we'll need to check if the last processed bucket is the current one, and there wasn't any cover bought in the current bucket - we'll need to apply the expirations in-memory (read amount, apply expirations on the fly, and return the value without updating the storage) - the reason for this being that the function is view. Expose a function to process them without buying a cover.

kyledewy:
## Context
Closes #387 

This PR adds support for active cover tracking. Everytime cover is bought, the amount to expire is stored in the bucket after the cover expires. Everytime `burnStake()` and `buyCover` are called, all buckets before and up to `currentBucketId` are expired, reducing `activeCoverAmountInAsset`.
When a cover is edited, the previous segment is instantly expired, and a new one is created at the bucket after the end of the new cover segment...
When a claim is made `burnStake()` is called, and this could be referencing an already expired segment or one in the future. If it is expired, the burn amount is already included in the expired amount, so it only uses `expiredAmount`. If the segment hasn't expired, active cover is reduced by `(expiredAmount + burnAmount)`.

## Test plan
Tests are added at: `test/unit/Cover/totalActiveCoverInAsset.js`

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/modules/cover/Cover.sol
```diff
@@ -91,7 +92,10 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
   // Eg. coverAssetsFallback = 3 (in binary 11) means assets at index 0 and 1 are supported.
   uint32 public coverAssetsFallback;
 
-  // TODO: implement using buckets
+  // The last time activeCoverExpirationBuckets was updated
+  uint public lastBucketUpdateId;
+  // assetId => bucketId => coverExpiring
+  mapping(uint24 => mapping(uint => uint)) public activeCoverExpirationBuckets;
```
shark0der, line 98: Use `uint` for the key, it should be slightly cheaper

contracts/modules/cover/Cover.sol
```diff
@@ -290,8 +301,15 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
       )
     );
 
-    // TODO: implement using buckets
-    totalActiveCoverInAsset[params.coverAsset] += coverAmountInCoverAsset;
+    // Update totalActiveCover
+    uint currentBucketId = block.timestamp / BUCKET_SIZE;
+    // TODO: delete expired buckets?
+    expiredCoverAmount += getExpiredCoverAmount(params.coverAsset, lastBucketUpdateId, currentBucketId);
```
shark0der, line 307: When `lastBucketUpdateId == currentBucketId`, the `getExpiredCoverAmount` function will return the amount for the current bucket. This seems to subtract the current bucket at every cover buy. Needs a check here to make sure you do only once. Ideally I would inline the while loop here and avoid generating a public function because (1) we already have one that is autogenerated and (2) contract size got to 24k, terribly close to the limit.

contracts/modules/cover/Cover.sol
```diff
@@ -290,8 +301,15 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+    lastBucketUpdateId = currentBucketId;
+    totalActiveCoverInAsset[params.coverAsset] = totalActiveCoverInAsset[params.coverAsset] + coverAmountInCoverAsset - expiredCoverAmount;
+    // Update expirationBuckets
+    uint bucketAtExpiry = Math.divCeil(block.timestamp + 1 + params.period, BUCKET_SIZE);
```
shark0der, line 311: Why the `+1`?

contracts/modules/cover/Cover.sol
```diff
@@ -91,7 +92,10 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
   // Eg. coverAssetsFallback = 3 (in binary 11) means assets at index 0 and 1 are supported.
   uint32 public coverAssetsFallback;
 
-  // TODO: implement using buckets
+  // The last time activeCoverExpirationBuckets was updated
+  uint public lastBucketUpdateId;
```
shark0der, line 96: We need to track the bucket ids for each asset separately, otherwise you'll only expire cover amounts on one asset and skip updates on others. This can be packed nicely together with `totalActiveCoverInAsset` that you have below. Creating a struct and limiting its size to a single slot should do it.

shark0der (commented):
Direction definitely looks good! Left a couple of comments with some changes that are currently needed!

contracts/modules/cover/Cover.sol
kyledewy: Oops thats unnecessary. Removed it.

contracts/modules/cover/Cover.sol
```diff
@@ -505,9 +555,28 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+      uint lastUpdateId = _activeCover.lastBucketUpdateId;
+      uint currentBucketId = block.timestamp / BUCKET_SIZE;
+      _activeCover.lastBucketUpdateId = currentBucketId.toUint64();
+      uint expiredCoverAmount = expireActiveCoverBuckets(cover.coverAsset, lastUpdateId, currentBucketId);
+      // Reduce total active cover
+      uint burnSegmentBucketId = Math.divCeil((segment.start + segment.period), BUCKET_SIZE);
+      uint amountToBurn = (expiredCoverAmount + burnAmount);
+
+      // TODO: how to handle claims that come after cover segment has expired
+      _activeCover.totalActiveCoverInAsset = amountToBurn > _activeCover.totalActiveCoverInAsset
```
kyledewy, line 571: This was a quick fix for the claims "double" expiry problem. Wasn't sure of a better way to handle this scenario.

contracts/modules/cover/Cover.sol
```diff
@@ -109,6 +111,12 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     globalCapacityRatio = 20000; // x2
     globalRewardsRatio = 5000; // 50%
     coverAssetsFallback = 3; // 0x11 - DAI and ETH
+    uint64 bucketIdStart = (block.timestamp / BUCKET_SIZE).toUint64();
+    uint assetId;
+    while (assetId <= coverAssetsFallback) {
+      activeCover[assetId].lastBucketUpdateId = bucketIdStart;
+      ++assetId;
+    }
```
shark0der, line 119: Each `assetId` should have a single bit `1` and others `0` (ex `0b010`). Current loop will go through all numbers instead of just looping through valid asset ids.

contracts/modules/cover/Cover.sol
```diff
@@ -206,8 +220,23 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+    // Update totalActiveCover
+    {
+      ActiveCover storage _activeCover = activeCover[params.coverAsset];
+      uint lastUpdateId = _activeCover.lastBucketUpdateId;
+      uint currentBucketId = block.timestamp / BUCKET_SIZE;
+      _activeCover.lastBucketUpdateId = currentBucketId.toUint64();
+      // Get expired cover amount
+      expiredCoverAmount += expireActiveCoverBuckets(params.coverAsset, lastUpdateId, currentBucketId);
+      // Update total active cover in storage
+      _activeCover.totalActiveCoverInAsset = (_activeCover.totalActiveCoverInAsset + coverAmountInCoverAsset - expiredCoverAmount).toUint192();
```
shark0der, line 233: Style: line exceeds 120 chars

roxdanila:
Main point from me is that there are a bunch of skipped unit tests related to active cover amount tracking. Would be good to get them fixed and re-enabled with this PR as well.

Screen Shot 2023-01-11 at 6 36 09 PM
Screen Shot 2023-01-11 at 6 36 32 PM

Also these ones are likely not relevant anymore:
Screen Shot 2023-01-11 at 6 36 18 PM

shark0der (changes requested):
As per the discussion:
- Do not delete the buckets and make the function view
- Fix the accounting when burning (too much is subtracted)

shark0der:
Blocked by [#624](https://github.com/NexusMutual/smart-contracts/issues/624)

Issue #553, opened 2022-12-13
CoverTrancheAllocation must be set


roxdanila:

roxdanila:
duplicate of https://github.com/NexusMutual/smart-contracts/issues/591

Issue #552, opened 2022-12-13
Staking: Generate allocation id and return it to Cover


roxdanila:
When the Cover contract calls the staking pool to allocate the capacity, the staking pool will store the cover allocations using the cover id key (`coverTrancheAllocations[coverId]` mapping). This will lead to issues if the cover buyer requests allocation from the same pool multiple times for the same cover (repeated pool id in the array). To prevent this we need the staking pools to use an allocation id that will be generated by the staking pool itself for each allocation request and will return it to the cover contract. The allocation id should be reused when editing the cover (so stays the same when allocation is edited). If the user repeats the pool inside the allocation requests array - each item will have a separate allocation id.

PR #603, opened 2023-01-06
Feature: Generate allocation ID

MilGard91:
## Context

Closing issues: #552, #553

Adding allocation id that is generated in all new allocations, that can be later used in editing those allocations.

## Test plan

Two new tests were added for this change but also modified all the tests that have `buyCover`, since the signature of the function has changed.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/interfaces/ICover.sol
+++ b/contracts/interfaces/ICover.sol
@@ -37,6 +37,7 @@ enum CoverUintParams {
 struct PoolAllocationRequest {
   uint40 poolId;
   uint coverAmountInAsset;
+  uint allocationId;
 }
 
 struct RequestAllocationVariables {
@@ -79,6 +80,7 @@ struct PoolAllocation {
   uint40 poolId;
   uint96 coverAmountInNXM;
   uint96 premiumInNXM;
+  uint allocationId;
 }
 
 struct CoverData {
--- a/contracts/interfaces/ICover.sol
+++ b/contracts/interfaces/ICover.sol
@@ -37,7 +37,6 @@ enum CoverUintParams {
 struct PoolAllocationRequest {
   uint40 poolId;
   uint coverAmountInAsset;
-  uint allocationId;
 }
 
 struct RequestAllocationVariables {
--- a/contracts/interfaces/ICover.sol
+++ b/contracts/interfaces/ICover.sol
@@ -79,7 +79,7 @@ struct PoolAllocation {
   uint40 poolId;
   uint96 coverAmountInNXM;
   uint96 premiumInNXM;
-  uint allocationId;
+  uint24 allocationId;
 }
 
 struct CoverData {
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -126,7 +126,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     uint segmentId;
 
     AllocationRequest memory allocationRequest;
-
+    uint maxUint = type(uint).max;
     {
       require(_products.length > params.productId, "Cover: Product not found");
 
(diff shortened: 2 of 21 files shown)


contracts/modules/cover/Cover.sol
```diff
@@ -126,7 +126,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     uint segmentId;
 
     AllocationRequest memory allocationRequest;
-
+    uint maxUint = type(uint).max;
```
shark0der, line 129: This needlessly occupies a stack slot as we don't actually need this variable. Remove the variable and inline the `type(uint).max` expression.

contracts/interfaces/ICover.sol
```diff
@@ -79,6 +80,7 @@ struct PoolAllocation {
   uint40 poolId;
   uint96 coverAmountInNXM;
   uint96 premiumInNXM;
+  uint allocationId;
```
shark0der, line 83: Is this used anywhere?

contracts/modules/cover/Cover.sol
```diff
@@ -260,33 +261,39 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-
-      // omit deallocated pools from the segment
-      if (coverAmountInNXM != 0) {
-        coverSegmentAllocations[allocationRequest.coverId][segmentId].push(
-          PoolAllocation(
-            poolAllocationRequests[i].poolId,
-            coverAmountInNXM.toUint96(),
-            premiumInNXM.toUint96()
-          )
+      {
```
shark0der, line 264: The added scope is redundant here as the block is the latest in the for loop block and it won't help to avoid a stack too deep error. Remove the added scope.

contracts/interfaces/ICover.sol
```diff
@@ -37,6 +37,7 @@ enum CoverUintParams {
 struct PoolAllocationRequest {
   uint40 poolId;
   uint coverAmountInAsset;
+  uint allocationId;
```
shark0der, line 40: `PoolAllocationRequest` is user input. If the pool is new - it should be max uint, if the pool is not new - it should be the previous allocation id (which we already know). As a consequence it doesn't seem like we need any user input here. Remove this field.

contracts/modules/cover/Cover.sol
```diff
@@ -260,33 +261,39 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+        // converting asset amount to nxm and rounding up to the nearest NXM_PER_ALLOCATION_UNIT
+        uint coverAmountInNXM = Math.roundUp(
+          Math.divCeil(poolAllocationRequests[i].coverAmountInAsset * ONE_NXM, nxmPriceInCoverAsset),
+          NXM_PER_ALLOCATION_UNIT
+        );
+        if (poolAllocationRequests[i].allocationId == type(uint).max) {
+          allocationRequest.allocationId = type(uint).max;
+        } else {
+          allocationRequest.allocationId = poolAllocationRequests[i].allocationId;
+        }
```
shark0der, line 274:
```
if (source == X) {
  target = X;
} else {
  target = source;
}
```
wouldn't target always get assigned the source value? :grin: 

Instead of trusting user input, this should be assigned based on whether there's a previous allocation id or not. This check is done at line 249 (`if (vars.previousPoolAllocationsLength > i)`).

contracts/modules/staking/StakingPool.sol
```diff
@@ -615,13 +616,16 @@ contract StakingPool is IStakingPool {
       );
 
       // no need to charge any premium
-      return 0;
+      return (0, request.allocationId);
```
shark0der, line 619: Since the amount is zero and there's no allocation anymore - return `(0, 0)`

contracts/modules/staking/StakingPool.sol
```diff
@@ -615,13 +616,16 @@ contract StakingPool is IStakingPool {
...
     }
-
+    uint coverAllocationAmount;
```
shark0der, line 621: Insert a blank line after a scope/block's closing `}`

shark0der (changes requested):
Almost there! :raised_hands:

shark0der (approved):

Issue #586, opened 2022-12-23
Bug in manager reward fee calculation


roxdanila:
Anatol, [Dec 21, 2022 at 7:52:37 PM]:
i think i've spotted a "bug" in the manager's reward shares calculation

if let's say the pool fee = 5 (%)

if the user gets 100 reward shares, the manager will get 5 reward shares

5 / 105 is not 5%

this part was tested but the tests were done to match the solidity

the formula should probably be:

uint newFeeRewardShares = newRewardsShares * poolFee / (POOL_FEE_DENOMINATOR - poolFee);

so for new 100 reward shares the manager would get
100 * 5 / (100 - 5) = 5.263157894736842

and to check the math: 
5.263157894736842 / 105 = 0.05012531328320802 %

need a sanity check please

i'm not entirely sure this is correct

the initial implementation technically is not wrong, but it's misleading

because you set it to be 5% but it won't be 5%

it'll be something different

it mints 5% on top, instead of manager's fee to be 5%

Issue #545, opened 2022-12-06
Extract NFT out of the staking contract


shark0der:
- [ ] new staking NFT contract
- [ ] pool manager ownership 
- [ ] tokenId 0 shouldn't be treated as a special case

roxdanila:
Also addressed: 
- Audit 5.1.1 Locking of funds of addresses with TokenController allowances

Issue #608, opened 2023-01-09
Fork test: Migrate PooledStaking


danoctavian:
Complete the following stages in the fork test migration-v2.js file.

- process all PooledStaking pending actions
- migrate top stakers to new v2 staking pools

roxdanila:
@danoctavian re `migrate top stakers to new v2 staking pools`:

We'll have a hardcoded list of members that will migrate their stake to create a new staking pool and move over their existing product allocations and current prices as initial state.

List of stakers to be mkigrated to staking pool managers:

- [ ] 0x1337def1fc06783d4b03cb8c1bf3ebf7d0593fc4
- [ ] 0x963df0066ff8345922df88eebeb1095be4e4e12e
- [ ] 0x87b2a7559d85f4653f13e6546a14189cd5455d45
- [ ] 0x46de0c6f149be3885f28e54bb4d302cb2c505bc2

Everybody else should have their stakes unlocked and will have the option to withdraw it anytime, and if they decide, they can then delegate to any existing staking pool manually.

This will involve some changes to the PoolStaking.sol current logic FYI.

roxdanila:
Last TODO: push all rewards before processing and disable functionality to accumulate more

PR #613, opened 2023-01-09
Feature: Staking nft as standalone contract

shark0der:
## Context

This PR addresses the following issues:
- #545 
- #563 
- #551 
- #561 
- #586 

- Created the `StakingNFT` contract based on Solmate's `ERC721` along with an interface for it which extends OZ's `IERC721`. The contract also supports [ERC-165: Standard Interface Detection](https://eips.ethereum.org/EIPS/eip-165). Just like `CoverNFT` - `StakingNFT` is not an internal contract and is not upgradable.
- Given there's a single NFT for all pools, manager tracking is no longer done using an NFT and each staking pool is tracking it individually via the `manager` state variable. As a consequence, the manager's fee is tracked as a deposit with id `MAX_UINT` which is also the flag for a new deposit (passed via the `tokenId` parameter). This also allows us to remove the check if the token owner has voted in goverance (needs to be implemented when switching the manager).
- The initial deposit was removed. The rationale for that is that given how rarely this operation is done, it's really not worth implementing a separate code path, and during audit a DoS vector was found that was abusing it.
- Created the `StakingPoolFactory` contract which is used to deploy an instance of the `MinimalBeaconProxy` using `CREATE2` - this gives us the opportunity to calculate the addresses of the deployed staking pools deterministically. Just like the NFT contracts, the factory is not an internal contract and is not upgradable - which is desired in this case because it needs to deploy the exactly same bytecode and it's crucial to avoid any accidental changes to it because that would change the init code hash. Both `StakingPoolFactory` and `MinimalBeaconProxy` have an exact solidity pragma version and have the optimizations forcefully turned on via `hardhat.config.js/solidity.js`. The corresponding functionality was removed from `CoverUtilsLib`.
- `MinimalBeaconProxy` constructor was modified to remove the constructor argument and it calls back the factory to fetch the beacon address instead. This makes it easier to have deterministic addresses on all chains (dev/test/main-nets) and avoid the dependence on the beacon address which would change the init code hash (normally the constructor arguments are appended at the end and technically are part of the init code).
- Created the `StakingPoolLibrary` library which provides a pure function to calculate the address of a staking pool and requires the staking pool factory address and the pool id as inputs. There's also the `patch-staking-pool-library.js` script which compiles the contracts, calculates the init code hash and updates the library code. There's also a test for this but it will be included in a subsequent PR which fixes all tests.
- The state variables in the `Cover` contract were moved at the top. Rationale is that if someone adds a constant but forgets to add the `constant` specifier - it will become a storage variable which will mess up all slots. With the constants at the bottom such mistakes can still happen but there's a much lower chance to mess everything up as it adds a new slot instead of shifting the existing ones so at least we have a chance to reverse the change and fix the storage layout. Ideally this should be enforced everywhere.
- `Cover` contract now requires the staking NFT and staking pool factory as constructor parameters which are stored as immutable variables.
- Cover migration functions were moved from `Cover`/`CoverUtilsLib` to the already existing and conveniently named `CoverMigrator` contract. Instead, `Cover` has now a `addLegacyCover` `onlyInternal` function which is called from the `CoverMigrator`.
- Staking nfts now can also be transfered when switching the membership
- `TokenController` no longer calls the `Cover` contract to get the staking pool ids but uses the library's pure function instead.

TBD (will open separate issues):
- Transfer staking pool ownership, consider implementing [EIP-173: Contract Ownership Standard](https://eips.ethereum.org/EIPS/eip-173). Check if voted in governance when the pool ownership is transfered.
- `LegacyPooledStaking` deposit migration
- There are some others issues which I've marked with TODOs in the code

## Test plan

All tests are fixed but will come in a separate PR in order to make it easier to review the code.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/modules/token/TokenController.sol
```diff
@@ -32,11 +30,21 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
+    address quotationDataAddress,
+    address claimsRewardAddress,
+    address stakingPoolFactoryAddress
+  ) {
     quotationData = IQuotationData(quotationDataAddress);
     claimsReward = claimsRewardAddress;
+    stakingPoolFactory = stakingPoolFactoryAddress;
   }
 
+  // TODO: probably better to move this to master
```
roxdanila, line 47: one time use, better leave a comment

scripts/patch-staking-pool-library.js
```diff
@@ -0,0 +1,48 @@
+const fs = require('fs');
```
roxdanila, line 1: leave a comment at the top to explain why it's needed pls

contracts/libraries/StakingPoolLibrary.sol
```diff
@@ -0,0 +1,25 @@
...
+library StakingPoolLibrary {
+
+  function getAddress(address factory, uint poolId) internal pure returns (address) {
+
+    bytes32 hash = keccak256(
+      abi.encodePacked(
+        hex'ff',
+        factory,
+        poolId, // salt
+        hex'203b477dc328f1ceb7187b20e5b1b0f0bc871114ada7e9020c9ac112bbfb6920' // init code hash
```
danoctavian, line 17: Can it be clarified in the comment where this code is coming from?

Looks magical at first glance

contracts/modules/cover/Cover.sol
```diff
@@ -460,39 +427,38 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-    emit StakingPoolCreated(
-      stakingPoolAddress,
-      stakingPoolCount,
-      manager,
-      stakingPoolImplementation
-    );
-
-    stakingPoolCount++;
-
+    // TODO: consider returning the pool id as well
```
roxdanila, line 461: please return it

contracts/modules/legacy/LegacyGateway.sol
```diff
@@ -51,7 +52,8 @@ contract LegacyGateway is IGateway, MasterAware {
     incidents = ILegacyIncidents(master.getLatestAddress("IC"));
     pool = IPool(master.getLatestAddress("P1"));
     memberRoles = IMemberRoles(master.getLatestAddress("MR"));
-    cover = ICover(master.getLatestAddress("CO"));
+    // TODO: check contract code
```
danoctavian, line 55: what's this TODO about?

Is there something left to do here?

contracts/modules/legacy/LegacyPooledStaking.sol
```diff
@@ -1474,10 +1474,12 @@ contract LegacyPooledStaking is IPooledStaking, MasterAwareV2 {
       initialPoolFee,
       maxPoolFee,
       params,
-      deposit,
-      trancheIdInEffect,
       migrationData.ipfsDescriptionHash
     );
+
+    // TODO: create the first deposit
```
danoctavian, line 1480: another TODO left here.

Do we have a ticket to address this? Let's link it if so so we don't leave things hanging

roxdanila: This function needs a bit of a rewrite as part of https://github.com/NexusMutual/smart-contracts/issues/608 - the TODO can be treated then.

contracts/modules/legacy/LegacyGateway.sol
roxdanila: just make sure that CoverMigrator has indeed the "CL" code - can you confirm @danoctavian ?

contracts/modules/staking/StakingNFT.sol
```diff
@@ -0,0 +1,187 @@
...
+    return
+      interfaceId == 0x01ffc9a7 || // ERC165 Interface ID for ERC165
+      interfaceId == 0x80ac58cd || // ERC165 Interface ID for ERC721
+      interfaceId == 0x5b5e139f;   // ERC165 Interface ID for ERC721Metadata
+  }
+
+  // ERC721
+
+  function tokenURI(uint id) public view virtual returns (string memory) {
+    // TODO: implement token uri
```
danoctavian, line 96: sounds like this should be addressed part of this PR

unless there's some product level decision we're waiting for, in which case let's link a ticket

contracts/modules/legacy/LegacyGateway.sol
danoctavian: @roxdanila  confirming CL is CoverMigrator in the upgrade:

https://github.com/NexusMutual/smart-contracts/blob/9ab7a560500c441cbd47e0c624eff17bcb657db7/test/fork/migration-v2.js#L410

Reason being as we know the backwards compatibility with armor and any other users of the legacy Claims code path for submitClaim.

contracts/modules/staking/StakingPoolFactory.sol
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  function create(address _beacon) external returns (uint poolId, address stakingPoolAddress) {
+
+    require(
+      msg.sender == operator,
+      "StakingPoolFactory: Only operator can create staking pools"
+    );
+
+    beacon = _beacon;
```
patitonar, line 31: Does it make sense to empty `beacon` after the proxy is deployed for the gas refund?

contracts/modules/staking/StakingNFT.sol
shark0der: https://github.com/NexusMutual/smart-contracts/pull/613#discussion_r1065624351

contracts/modules/staking/StakingPoolFactory.sol
shark0der: It's a similar situation to the reentrancy locks - OZ decided to use values `1` and `2` instead of zero - non-zero. The reason is because the next time the slot will be dirty and it will be cheaper to write.

contracts/modules/cover/CoverNFT.sol
```diff
@@ -7,6 +7,8 @@ import "solmate/src/tokens/ERC721.sol";
 
 contract CoverNFT is ERC721 {
 
+  // TODO: consider adding totalSupply()
```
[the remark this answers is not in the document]
shark0der: #619

contracts/modules/staking/StakingPool.sol
```diff
@@ -176,42 +176,23 @@ contract StakingPool is IStakingPool, ERC721 {
...
-    address from,
-    address to,
-    uint[] calldata tokenIds
-  ) external onlyCoverContract {
-    uint length = tokenIds.length;
-    for (uint i = 0; i < length; i++) {
-      transferFrom(from, to, tokenIds[i]);
-    }
+  function isApprovedOrOwner(address spender, uint tokenId) internal view returns (bool) {
+    // TODO: will revert for inexistent token
```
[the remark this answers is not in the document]
shark0der: Seems like this behaviour is as per spec.

contracts/libraries/StakingPoolLibrary.sol
shark0der: Comment added.

contracts/modules/cover/CoverMigrator.sol
```diff
@@ -1,36 +1,156 @@
...
+    uint amount;
+
+    {
+      (
+      /*uint coverId*/,
+      address coverOwner,
+      address legacyProductId,
+      bytes4 currencyCode,
+      /*uint sumAssured*/,
+      /*uint premiumNXM*/
```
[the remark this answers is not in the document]
shark0der: Done

contracts/modules/cover/Cover.sol
```diff
@@ -443,15 +361,64 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+    uint coverAsset,
+    uint amount,
+    uint start,
+    uint period,
+    address newOwner
+  ) external onlyInternal returns (uint coverId) {
+
+    ProductType memory productType = _productTypes[_products[productId].productType];
+
+    // TODO: consider using v1 grace period
```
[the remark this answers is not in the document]
shark0der: Updated.

contracts/modules/legacy/LegacyGateway.sol
```diff
@@ -16,6 +16,8 @@ import "../../interfaces/IQuotation.sol";
 import "../../interfaces/IQuotationData.sol";
 import "../../interfaces/ITokenController.sol";
 import "../../interfaces/ICover.sol";
+// TODO: decide if we want to use an interface
```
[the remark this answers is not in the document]
shark0der: Removed.

contracts/modules/legacy/LegacyGateway.sol
```diff
@@ -136,7 +138,7 @@ contract LegacyGateway is IGateway, MasterAware {
   function submitClaim(uint coverId, bytes calldata data) external override returns (uint) {
     // [todo] Maybe we could use data to specify other addresses and only use tx.origin if empty,
```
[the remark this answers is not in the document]
shark0der: Removed.

contracts/modules/staking/StakingNFT.sol
```diff
@@ -0,0 +1,187 @@
...
+    require(msg.sender == operator, "NOT_OPERATOR");
+    require(newOperator != address(0), "INVALID_OPERATOR");
+    operator = newOperator;
+  }
+
+  // minting and supply
+
+  function mint(uint poolId, address to) public returns (uint id) {
+
+    // TODO: make sure msg.sender is a staking pool
```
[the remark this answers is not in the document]
shark0der: Done.

contracts/modules/token/TokenController.sol
shark0der: Updated.

scripts/patch-staking-pool-library.js
shark0der: Comment added.

contracts/modules/cover/Cover.sol
shark0der: Updated.

contracts/modules/staking/StakingPoolFactory.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+pragma solidity =0.8.16;
+
+import "../../interfaces/IStakingPoolFactory.sol";
+import "./MinimalBeaconProxy.sol";
+
+contract StakingPoolFactory is IStakingPoolFactory {
+
+  address public operator;
```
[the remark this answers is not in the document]
shark0der: Added.

roxdanila (approved):

patitonar (approved):

Issue #566, opened 2022-12-13
CoverSegment struct: add the current globalCapacityFactor to it and store it when buying / editing a cover


roxdanila:
When burning, we need to calculate what amount to burn by dividing it to the global capacity factor. If the global capacity factor changed since the cover was bought - we will need the one that was actual at the cover buy. Add a new field in the CoverSegment struct and store the global capacity factor at cover buy. When burning use the saved one instead of the global.

PR #621, opened 2023-01-11
Fix: Add the current globalCapacityFactor to store

MilGard91:
## Context

closing #566 

Adding `globalCapacityFactor` to segments of the Cover to be used when burning stake.

## Test plan

Test expanded with test with change of globalCapacityFactor before burn stake

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -466,7 +466,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
   function burnStake(
     uint coverId,
     uint segmentId,
-    uint burnAmount
+    uint payoutAmountInAsset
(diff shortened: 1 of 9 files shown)


contracts/modules/cover/Cover.sol
```diff
@@ -390,7 +391,8 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
         start.toUint32(),
         period.toUint32(),
         productType.gracePeriod,
-        0 // global rewards ratio
+        0, // global rewards ratio
+        globalCapacityRatio
```
shark0der, line 395: Technically we won't burn any staking pool so the amount here is irelevant, but since we're (hypothetically) divide by it, can you set this to `1`, please?

shark0der (approved):

Issue #609, opened 2023-01-09
Fork test: assert storage slots on newly deployed contracts to check for storage corruption


danoctavian:
Check storage slots versus the previous version with assertions of the contract for :

NXMaster, GV, CR, TC, MCR, MR, CO, PS, P1, GW, CoverMigrator

To make sure the storage slots that need to be preserved are preserved correctly.

PR #640, opened 2023-01-19
Migration: V2 migration fork test storage slots

MilGard91:
## Context

Closing: #609

Extracting Storage Layout from artifacts and using them to check the values on contracts using `getStorageAt` function before and after upgrade.

## Test plan

Fork test expanded with storage tests.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/token/TokenController.sol
+++ b/contracts/modules/token/TokenController.sol
@@ -21,8 +21,8 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
 
   address public _unused_token;
   address public _unused_pooledStaking;
-  address public _unused_minCALockTime;
-  address public _unused_claimSubmissionGracePeriod;
+  uint public _unused_minCALockTime;
+  uint public _unused_claimSubmissionGracePeriod;
 
   // coverId => CoverInfo
   mapping(uint => CoverInfo) public override coverInfo;
--- a/test/fork/migration-v2.js
+++ b/test/fork/migration-v2.js
@@ -175,6 +175,7 @@ describe('V2 upgrade', function () {
     this.claimsData = await factory('CD');
     this.pooledStaking = await factory('PS');
     this.gateway = await factory('GW');
+    this.priceFeedOracle = await factory('PRICEORACLE');
 
     // Pool value related info
     this.poolValueBefore = await this.pool.getPoolValueInEth();
@@ -189,6 +190,22 @@ describe('V2 upgrade', function () {
 
     this.enzymeShares = await ethers.getContractAt('ERC20Mock', ENZYMEV4_VAULT_PROXY_ADDRESS);
     this.enzymeSharesBalanceBefore = await this.enzymeShares.balanceOf(this.pool.address);
+
+    // non proxy contract data
+    this.contractData = {
+      mcr: {
+        before: {},
+        after: {},
+      },
+      pool: {
+        before: {},
+        after: {},
+      },
+      priceFeedOracle: {
+        before: {},
+        after: {},
+      },
+    };
   });
 
   // Setup needed to test that `claimPayoutAddress` storage is cleaned up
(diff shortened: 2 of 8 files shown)


test/fork/migration-v2.js
```diff
@@ -163,8 +178,34 @@ describe('v2 migration', function () {
...
+      },
+    };
+  });
+
+  it('compare storage layout of proxy upgradable contracts', function () {
+    compareStorageLayout(v1StorageLayout.NXMaster, v2StorageLayout.NXMaster);
+    compareStorageLayout(v1StorageLayout.Governance, v2StorageLayout.Governance);
+    compareStorageLayout(v1StorageLayout.MemberRoles, v2StorageLayout.MemberRoles);
+    compareStorageLayout(v1StorageLayout.PooledStaking, v2StorageLayout.LegacyPooledStaking);
+    compareStorageLayout(v1StorageLayout.Gateway, v2StorageLayout.LegacyGateway);
```
danoctavian, line 208: Any reason to skip TokenController here?

danoctavian: It needs to be in

test/fork/migration-v2.js
```diff
@@ -555,6 +638,101 @@ describe('v2 migration', function () {
...
+
+    Object.entries(this.contractData.priceFeedOracle.before.assetsForEth).forEach(([asset, value]) => {
+      expect(
+        this.contractData.priceFeedOracle.after.assetsForEth[asset],
+        `AssertionError: values of assetsEthRate in ${asset} don't match\n`,
+      ).to.be.equal(value);
+    });
+  });
+
   it.skip('deploy staking pool', async function () {
```
danoctavian, line 736: let's assert one more thing here post upgrades:

for each contract assert it's dependencies in MasterAwareV2.internalContracts are set correctly.

`internalContracts` is now made public so you can just read them.

Each contract has a subset of deps so we could whitelist those for each, an approach similar to all the generalized assertions done til now will work nicely.

If anyone thinks this should be a different Pr, i'm ok with that, but the complexity is not that high

we

shark0der (changes requested):
Moved the storage tests to their own folder. Needs a better implementation for the test itself, group the state variables by slot and check every offset instead of looking them up by name.

test/storage/index.js
```diff
@@ -0,0 +1,116 @@
+const { config, run } = require('hardhat');
+const { expect } = require('chai');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+const extractStorageLayout = require(path.join(config.paths.root, 'scripts/extract-storage-layout'));
+
+describe('STORAGE LAYOUT TESTS', function () {
```
danoctavian, line 9: good work on separating this in its own test.

The evaluation does not need dynamic on-chain data.

test/storage/index.js
```diff
@@ -0,0 +1,116 @@
+const { config, run } = require('hardhat');
```
danoctavian, line 1: now what i'd suggest is creating another file here sitting next to index.js to follow the pattern for tests so far.

I know it's a single file but in case we break things down and to keep it consistent we want the index.js file to be just that, the index of all the tests such as here:

https://github.com/NexusMutual/smart-contracts/blob/862539ae93e7c00d5ccb9069f5afa0c5fcd4704e/test/unit/Assessment/index.js#L4

danoctavian, line 1: Furthermore, can we add this run to the build to run automatically next to unit and integration tests?

As opposed to the migration test this doesn't need a network connection to a fork node so it should run easily.

test/index.js
```diff
@@ -1,4 +1,5 @@
 describe('Nexus Mutual', function () {
   require('./unit');
   require('./integration');
+  require('./storage');
```
danoctavian, line 4: great making it part of the build

danoctavian:
@MilGard91  still needs a lint fix

shark0der:
To be merged only after `fix/migration-v2-stakingpool-pricing-asserts` is merged #691

test/fork/migration-v2.js
shark0der: Probably better to implement this in the basic functionality tests.

shark0der (approved):

contracts/modules/token/TokenController.sol
```diff
@@ -21,8 +21,8 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
 
   address public _unused_token;
   address public _unused_pooledStaking;
-  address public _unused_minCALockTime;
-  address public _unused_claimSubmissionGracePeriod;
+  uint public _unused_minCALockTime;
```
danoctavian, line 24: the fact that it was an address before was strange originally in v1 it used to be uint

test/fork/migration-v2.js
```diff
@@ -590,6 +648,101 @@ describe('V2 upgrade', function () {
...
+    this.contractData.mcr.after.desiredMCR = await this.mcr.desiredMCR();
+    this.contractData.mcr.after.lastUpdateTime = await this.mcr.lastUpdateTime();
+    this.contractData.mcr.after.previousMCR = await this.mcr.previousMCR();
+
+    Object.entries(this.contractData.mcr.before).forEach(([key, value]) => {
+      expect(this.contractData.mcr.after[key], `AssertionError: values of ${key} don't match\n`).to.be.equal(value);
+    });
+
+    // POOL
+    const assetsData = await Promise.all(
```
danoctavian, line 668: can we make an assert here for the length or what assets we expect there to be?


The code is generalized so it would make it obvious everything that should there is in there

test/layout/slots.js
```diff
@@ -0,0 +1,152 @@
...
+      // t_mapping(t_uint256,t_struct(CoverInfo)_storage)
+      .replace(/\)\d+/, ')')
+      // 2. replace contract types with t_address
+      .replace(/t_contract\([^)]+\)/, 't_address')
+      // 2. ignore payable specifier for addresses
+      .replace(/t_address_payable/, 't_address')
+  );
+};
+
+describe('Storage layout', function () {
```
danoctavian, line 23: have we verifed the quality of this test by intentionally breaking layout to see the differences?

PR #663, opened 2023-01-27
Feature: v1 partial claims

shark0der:
## Context

[#654](https://github.com/NexusMutual/smart-contracts/issues/654)

- Implements partial claims on top of V1. Once the claim is accepted - the cover is no longer usable.
- Removes the status 12 (payout pending) by reverting when the payout fails in order to remove the retrying logic

## Test plan

Added new tests. Fixed the failing status 12/13/14 tests.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/claims/Claims.sol
+++ b/contracts/modules/claims/Claims.sol
@@ -74,15 +74,14 @@ contract Claims is IClaims, LegacyMasterAware {
     incidents = IIncidents(ms.getLatestAddress("IC"));
   }
 
-  /**
-   * @dev Submits a claim for a given cover note.
-   * Adds claim to queue incase of emergency pause else directly submits the claim.
-   * @param coverId Cover Id.
-   */
+  /// @dev Submits a claim for a given cover
+  /// @param coverId Cover Id.
   function submitClaim(uint coverId) external {
     _submitClaim(coverId, msg.sender, 0);
   }
 
+  /// @param coverId               the cover id
+  /// @param requestedPayoutAmount requested payout amount, or 0 if the claim is for the full covered amount
   function submitPartialClaim(uint coverId, uint requestedPayoutAmount) external {
     _submitClaim(coverId, msg.sender, requestedPayoutAmount);
   }
@@ -109,6 +108,7 @@ contract Claims is IClaims, LegacyMasterAware {
     uint coveredAmount = qd.getCoverSumAssured(coverId);
     require(requestedPayoutAmount <= coveredAmount, "Claims: Requested payout amount is greater than covered amount");
 
+    // if requested amount = 0, the claim is for the entire covered amount
     tc.markCoverClaimOpenWithRequestedAmount(coverId, requestedPayoutAmount);
     qd.changeCoverStatusNo(coverId, uint8(IQuotationData.CoverStatus.ClaimSubmitted));
 
--- a/contracts/modules/claims/Claims.sol
+++ b/contracts/modules/claims/Claims.sol
@@ -109,7 +109,7 @@ contract Claims is IClaims, LegacyMasterAware {
     uint coveredAmount = qd.getCoverSumAssured(coverId);
     require(requestedPayoutAmount <= coveredAmount, "Claims: Requested payout amount is greater than covered amount");
 
-    tc.markCoverClaimOpen(coverId, requestedPayoutAmount);
+    tc.markCoverClaimOpenWithRequestedAmount(coverId, requestedPayoutAmount);
     qd.changeCoverStatusNo(coverId, uint8(IQuotationData.CoverStatus.ClaimSubmitted));
 
     uint claimId = cd.actualClaimLength();
--- a/contracts/modules/claims/Claims.sol
+++ b/contracts/modules/claims/Claims.sol
@@ -80,14 +80,18 @@ contract Claims is IClaims, LegacyMasterAware {
    * @param coverId Cover Id.
    */
   function submitClaim(uint coverId) external {
-    _submitClaim(coverId, msg.sender);
+    _submitClaim(coverId, msg.sender, 0);
+  }
+
+  function submitPartialClaim(uint coverId, uint requestedPayoutAmount) external {
+    _submitClaim(coverId, msg.sender, requestedPayoutAmount);
   }
 
   function submitClaimForMember(uint coverId, address member) external onlyInternal {
-    _submitClaim(coverId, member);
+    _submitClaim(coverId, member, 0);
   }
 
-  function _submitClaim(uint coverId, address member) internal {
+  function _submitClaim(uint coverId, address member, uint requestedPayoutAmount) internal {
 
     require(!ms.isPause(), "Claims: System is paused");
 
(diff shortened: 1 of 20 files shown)


contracts/modules/claims/Claims.sol
```diff
@@ -80,14 +80,18 @@ contract Claims is IClaims, LegacyMasterAware {
    * @param coverId Cover Id.
    */
   function submitClaim(uint coverId) external {
-    _submitClaim(coverId, msg.sender);
+    _submitClaim(coverId, msg.sender, 0);
```
roxdanila, line 83: let's pass this as full cover amount here, as you're also returning it in `TC.coverInfo()` and it gets a bit confusing imo

contracts/modules/claims/ClaimsReward.sol
```diff
@@ -89,15 +89,16 @@ contract ClaimsReward is IClaimsReward, LegacyMasterAware {
...
 
-      if (payoutSucceeded) {
-        c1.setClaimStatus(claimId, 14);
-      } else {
-        c1.setClaimStatus(claimId, 12);
-      }
+      attemptClaimPayout(coverid);
+
+      // 14 = payout succeeded
+      c1.setClaimStatus(claimId, 14);
```
roxdanila, line 101: so this change assumes the payout always succeeds and you bypass status 12 (Claim Accepted Payout Pending) all together. I think this is fine for the purpose of FTX claims. 

side question: can `attemptClaimPayout` not be successful for any other reason than not enough tokens in the pool?

contracts/modules/claims/Claims.sol
danoctavian: I would expect the default is to claim full amount, unless we make 0 signify full amount, because claiming 0 means nothing

contracts/modules/claims/ClaimsReward.sol
danoctavian: Was there another expected wave of claims that we expect before v2 launch?

test/integration/Claims/dai-payouts.js
```diff
@@ -236,9 +235,9 @@ describe('DAI cover claim payouts', function () {
     assert(actualPayout.eq(expectedPayout), 'should have transfered the cover amount');
   });
 
-  it('[A1, status: 0, 7, 12, 13] CA accept, closed with closeClaim(), claim payout fails with status 12 and goes to status 13 after 60 retries', async function () {
+  it('[A1, status: 0, 7, -, 14] CA accept, closed with closeClaim(), claim payout fails, then succeeds', async function () {
```
danoctavian, line 238: what does the - mean between 7 and 14?

contracts/modules/claims/ClaimsReward.sol
shark0der: > can attemptClaimPayout not be successful for any other reason than not enough tokens in the pool?
yes, if:
- target address rejects eth transfers (contracts without receive/fallback)
- token to be transfered has blacklisted the pool or the recipient

I do not expect it to generally fail with reasons other than insufficients funds.

contracts/modules/claims/Claims.sol
shark0der: zero was a convention to indicate full amount. will change to full value.

test/integration/Claims/dai-payouts.js
shark0der: Claims can no longer have status 12 as the tx will fail. I've used `-` with the meaning that the status didn't change.

contracts/modules/claims/Claims.sol
danoctavian: I think either way is fine as long as it's documented since 0 doesn't make sense for a claim

test/integration/Claims/eth-payouts.js
```diff
@@ -87,6 +86,61 @@ describe('send claim payout to the payout address', function () {
...
+
+    const { cd, cl, qd, cr, tc } = this.contracts;
+    const coverAmount = 5;
+    const requestedAmount = 3;
+    const cover = { ...coverTemplate, amount: coverAmount };
+
+    await buyCover({ ...this.contracts, cover, coverHolder });
+    const [coverId] = await qd.getAllCoversOfUser(coverHolder);
+
+    await expectRevert(
```
danoctavian, line 99: would expect this as a separate test case normally but it's acceptable since this code is sunset

danoctavian:
Overall looks correct we should run the fork with the basic functionality tests to verify that  as well

contracts/modules/claims/Claims.sol
shark0der: Ended up just documenting it because the logic for setting it to a full amount implied changing the `Incidents` contract or adding a new dependant contract (`QuotationData`) to the `TokenController` contract, plus creating a new mock for `QuotationData` for `TokenController` tests, so it's not worth it.

danoctavian: sounds good

test/fork/utils.js
```diff
@@ -34,7 +34,7 @@ async function submitGovernanceProposal (categoryId, actionData, members, gv) {
   }
 
   await time.increase(604800);
-  const closeTx = await gv.closeProposal(id, { from });
+  const closeTx = await gv.closeProposal(id, { from, gas: 12e6 });
```
danoctavian, line 37: does it fail without this? curiosity, i haven't used this so far

test/fork/one-time-partial-claims.js
```diff
@@ -0,0 +1,406 @@
...
+      'TokenController: Max claim count exceeded',
+    );
+
+    await expectRevert(
+      claims.submitClaim(coverId, { from: coverOwner }),
+      'TokenController: Max claim count exceeded',
+    );
+  });
+
+  it('processes pending actions', async function () {
```
danoctavian, line 217: do we want to assert anything here in relation to burns?

test/fork/one-time-partial-claims.js
```diff
@@ -0,0 +1,406 @@
...
+    const { pooledStaking } = this;
+    while (await pooledStaking.hasPendingActions()) {
+      await pooledStaking.processPendingActions(100);
+    }
+  });
+
+  it('submits full claim for cover 7907', async function () {
```
danoctavian, line 224: is there a reason  linked to reality why one is partial and one is full for these 2 particular covers?

Or just to test the 2 cases on 2 arbitrary covers that are valid

test/fork/one-time-partial-claims.js
```diff
@@ -0,0 +1,406 @@
...
+    const { claims, quotationData, tokenController } = this;
+
+    const coverId = 7907; // FTX
```
danoctavian, line 227: in retrospect a parametrized function for claim submission would have made these 2 test cases easier to read

test/fork/one-time-partial-claims.js
```diff
@@ -0,0 +1,406 @@
...
+    // fast forward 72h
+    await time.increase(72 * 3600);
+
+    const coverAmount = await quotationData.getCoverSumAssured(coverId);
+    const coverOwner = await quotationData.getCoverMemberAddress(coverId);
+    const claimAmount = coverAmount;
+    const payoutAmount = ether(claimAmount.toString());
+
+    // eth
+    const coverOwnerETHBalanceBefore = toBN(await web3.eth.getBalance(coverOwner));
```
danoctavian, line 276: assuming this is just for printing but no ETH should go out

danoctavian (approved):
LGTM


fork test looks good, just some questions posted, happy to talk in DMs about them

roxdanila (approved):

PR #666, opened 2023-01-30
Test: Add unit tests for StakingNFT and StakingPoolFactory

### LINKED ISSUE #552: Staking: Generate allocation id and return it to Cover

kyledewy:
## Context

Closes #552 

Added unit tests for `StakingNFT` and `StakingPoolFactory` contracts

## Test plan
Added tests to:
`test/unit/StakingPoolFactory/stakingPoolFactory.js`
`test/unit/StakingNFT/stakingNFT.js`

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/test/unit/StakingNFT/stakingNFT.js
+++ b/test/unit/StakingNFT/stakingNFT.js
@@ -153,7 +153,7 @@ describe('StakingNFT', function () {
     ).to.be.revertedWith('NOT_ERC721_RECEIVER');
   });
 
-  it('should support erc721 interface', async function () {
+  it('should support erc721 and ERC165 interfaces', async function () {
     const { stakingNFT } = this;
     // 0x80ac58cd // ERC165 Interface ID for ERC721
     expect(await stakingNFT.supportsInterface('0x80ac58cd')).to.be.equal(true);
--- a/test/unit/StakingNFT/stakingNFT.js
+++ b/test/unit/StakingNFT/stakingNFT.js
@@ -143,14 +143,33 @@ describe('StakingNFT', function () {
     expect(await stakingNFT.ownerOf(0)).to.be.equal(nonOwner.address);
   });
 
-  // TODO: why is safeTransferFrom() not found?
-  it.skip('should fail to safeTransfer to a contract that does not implement onERC721Received', async function () {
+  it('should fail to safeTransfer with bytes to a contract without onERC721Received function', async function () {
     const { stakingNFT, cover } = this;
     const [owner] = this.accounts.members;
     await stakingNFT.connect(this.stakingPoolSigner).mint(this.poolId, owner.address);
+    // Reverts without reason if the contract does not implement onERC721Received
     await expect(
-      stakingNFT.connect(owner).safeTransferFrom(owner.address, cover.address, BigNumber.from(0), toBytes('cafe')),
-    ).to.be.revertedWith('NOT_ERC721_RECEIVER');
+      stakingNFT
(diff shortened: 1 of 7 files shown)


test/unit/StakingNFT/stakingNFT.js
```diff
@@ -53,9 +66,101 @@ describe('StakingNFT', function () {
...
+
+  // TODO: why is safeTransferFrom() not found?
+  it.skip('should fail to safeTransfer to a contract that does not implement onERC721Received', async function () {
+    const { stakingNFT, cover } = this;
+    const [owner] = this.accounts.members;
+    await stakingNFT.connect(this.stakingPoolSigner).mint(this.poolId, owner.address);
+    await expect(
+      stakingNFT.connect(owner).safeTransferFrom(owner.address, cover.address, BigNumber.from(0), toBytes('cafe')),
+    ).to.be.revertedWith('NOT_ERC721_RECEIVER');
+  });
```
patitonar, line 154: `safeTransferFrom()` is not found because there are two implementations of it. It is needed to provide the full signature of the method to use. In this case:
```js
await expect(
  stakingNFT
    .connect(owner)
    ['safeTransferFrom(address,address,uint256,bytes)'](
      owner.address,
      cover.address,
      BigNumber.from(0),
      toBytes('cafe'),
    ),
).to.be.revertedWith('NOT_ERC721_RECEIVER');
```

I tried it, and it works. There is one issue with the expected revert message  `NOT_ERC721_RECEIVER`. It is reverting without a reason. 

The problem is that the implementation of solmate does the following:
```solidity
require(
  to.code.length == 0 ||
  ERC721TokenReceiver(to).onERC721Received(msg.sender, from, id, data)
    == ERC721TokenReceiver.onERC721Received.selector,
  "UNSAFE_RECIPIENT"
);
```

As the contract does not implement `onERC721Received()`, the call `ERC721TokenReceiver(to).onERC721Received(msg.sender, from, id, data)` reverts without a reason before comparing the returned value. That revert message will only be thrown if a contract implements the method but does not return the correct value.

This is handled differently in [OpenZeppelin implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/91e8d0ba3c3beb8a1db31310d8599664e48639ef/contracts/token/ERC721/ERC721.sol#L406-L416).

I'm not sure if it is worth updating the contract implementation as the result is the same. As an alternative, we can just add a comment in the tests explaining why they are reverting in that way. Thoughts? @kyledewy @shark0der

patitonar (approved):
Looks good! Added one minor comment

test/unit/StakingNFT/stakingNFT.js
kyledewy: Oh nice! I totally forgot you could call functions like that.
I would be fine to leave it a silent revert, since it's the same result.

roxdanila (approved):

PR #704, opened 2023-02-08
Chore: Cleanup PS migration code and fix bugs

roxdanila:
## Context

Cleanup PS migration code

- renames for clarity
- remove unused struct
- move struct and constants at the beginning of the file
- indentation fixes
- bug fix where we were creating the staking pools by setting the manager address to be the staker address (they are different for Armor)
- added logic to prevent dust due to division errors

## Test plan

Ran fork test

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/legacy/LegacyPooledStaking.sol
+++ b/contracts/modules/legacy/LegacyPooledStaking.sol
@@ -11,17 +11,30 @@ import "../../interfaces/IProductsV1.sol";
 import "../../interfaces/IStakingPool.sol";
 
 contract LegacyPooledStaking is IPooledStaking, MasterAwareV2 {
+  /* Constants */
+
+  address constant ARMOR_STAKER = 0x1337DEF1FC06783D4b03CB8C1Bf3EBf7D0593FC4;
+  address constant ARMOR_MANAGER = 0xFa760444A229e78A50Ca9b3779f4ce4CcE10E170;
+  address constant HUGH = 0x87B2a7559d85f4653f13E6546A14189cd5455d45;
+  address constant NM_FOUNDATION = 0x963Df0066ff8345922dF88eebeb1095BE4e4e12E;
+  uint constant MAX_ACTIVE_TRANCHES = 8;
+
+  ICover public immutable cover;
+  IProductsV1 public immutable productsV1;
+
   /* Structs */
 
-  struct MigrationData {
+  struct StakingPoolMigrationData {
     address stakerAddress;
-    string ipfsDescriptionHash;
     address managerAddress;
+    string ipfsDescriptionHash;
     bool isPrivatePool;
     uint initialPoolFee;
     uint maxPoolFee;
-
+    uint deposit;
+    uint[MAX_ACTIVE_TRANCHES] trancheStakeRatio;
   }
+
   /* Events */
 
   // deposits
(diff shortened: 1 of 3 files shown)


contracts/modules/legacy/LegacyPooledStaking.sol
```diff
@@ -1197,208 +1198,196 @@ contract LegacyPooledStaking is IPooledStaking, MasterAwareV2 {
...
+    uint firstTrancheId = block.timestamp / 91 days + 1;
+
+    for (uint i = 0; i < MAX_ACTIVE_TRANCHES; i++) {
+      uint trancheDeposit = migrationData.deposit * migrationData.trancheStakeRatio[i] / 100;
+      if (trancheDeposit == 0) {
+        continue;
+      }
+
+      totalStakeRatio += migrationData.trancheStakeRatio[i];
+      // Prevent dust due to precision loss and deposit everything remaining in the last tranche
```
roxdanila, line 1276: open to feedback here

contracts/modules/legacy/LegacyPooledStaking.sol
```diff
@@ -1197,208 +1198,196 @@ contract LegacyPooledStaking is IPooledStaking, MasterAwareV2 {
...
+    stakers[stakerAddress].deposit = 0;
 
     INXMToken nxm = token();
     uint nxmBalanceBefore = nxm.balanceOf(address(this));
 
-    // ratios have no decimal points. eg 5 is 5%
-    uint[TRANCHE_COUNT] memory stakerTrancheRatios;
-
-    (ProductInitializationParams[] memory params, uint deposit) = getStakerConfig(stakerAddress);
+    ProductInitializationParams[] memory productInitParams = getProductInitParams(
```
roxdanila, line 1312: I took out the deposit from the output of this function on purpose, because it was/is a getter, but it mutated state.

danoctavian: The naming was not reflective of the mutation agreed i think what you did is better.

The usage of it is in only 1 place, it's not public but it does make things easier to reason about to have a pure view function. Thus we can even make it public in case it helps with checks/assertions.

contracts/modules/legacy/LegacyPooledStaking.sol
danoctavian: i think it's fine, the code runs only once we don't need to optimize it too much

Allocating all to last tranche that's non-zero works

contracts/modules/legacy/LegacyPooledStaking.sol
```diff
@@ -1197,208 +1198,196 @@ contract LegacyPooledStaking is IPooledStaking, MasterAwareV2 {
...
       migrateToPool(
         StakingPoolMigrationData(
+          NM_FOUNDATION, // stakerAddress
+          NM_FOUNDATION, // managerAddress
+          '', // ipfsDescriptionHash --- TODO fill in
+          true, // isPrivatePool
           0, // initialPoolFee
-          20, // maxPoolFee
-          stakerTrancheRatios,
+          99, // maxPoolFee
```
danoctavian, line 1374: Is this 99 still a dummy value for now?

contracts/modules/legacy/LegacyPooledStaking.sol
```diff
@@ -1207,213 +1208,196 @@ contract LegacyPooledStaking is IPooledStaking, MasterAwareV2 {
...
+          ARMOR_STAKER, // stakerAddress
+          ARMOR_MANAGER, // managerAddress
+          '', // ipfsDescriptionHash --- TODO fill in
           false, // isPrivatePool
-          '', // ipfsDescriptionHash
-          ARMOR_MANAGER // managerAddress
+          15, // initialPoolFee
+          25, // maxPoolFee
+          armorAAALowRiskPoolDeposit, // deposit
+          [uint256(20), 25, 25, 15, 10, 0, 0, 0] // stake on each tranche, as % out of the deposit
```
danoctavian, line 1353: this array is the same as the one below for Armor, i could argue for using the same constant as it already was

contracts/modules/legacy/LegacyPooledStaking.sol
roxdanila: made it public

contracts/modules/legacy/LegacyPooledStaking.sol
roxdanila: nope. it should be the max we can set this value to. we can't set it to 100, as we end up with division by 0 issues

contracts/modules/legacy/LegacyPooledStaking.sol
roxdanila: I am trying to avoid black boxes and scrolling back and forth. it doesn't make a difference afterall

contracts/modules/legacy/LegacyPooledStaking.sol
danoctavian: cool, will need to change the expected in the migration-v2.js then, it's set at 20

danoctavian (approved):
LGTM once build passes

PR #733, opened 2023-02-14
Fix: start covers at block timestamp

shark0der:
## Context

Cover edits where underflowing because of a mismatch between cover start in Cover contract (`block.timestamp + 1`) and cover start / rewards streaming start in StakingPool (`block.timestamp`).

This PR makes the cover start at `block.timestamp` and instead explicitly prevents claim submission in the same block.

## Test plan

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/assessment/IndividualClaims.sol
+++ b/contracts/modules/assessment/IndividualClaims.sol
@@ -291,7 +291,7 @@ contract IndividualClaims is IIndividualClaims, MasterAwareV2 {
         "Invalid claim method for this product type"
       );
       require(requestedAmount <= segment.amount, "Covered amount exceeded");
-      require(segment.start <= block.timestamp, "Cover starts in the future");
+      require(block.timestamp > segment.start, "Cannot buy cover and submit claim in the same block");
       require(
         uint(segment.start) + uint(segment.period) + uint(segment.gracePeriod) > block.timestamp,
         "Cover is outside the grace period"
--- a/test/unit/Cover/editCover.js
+++ b/test/unit/Cover/editCover.js
@@ -1478,7 +1478,7 @@ describe('editCover', function () {
       .mul(period)
       .div(priceDenominator)
       .div(3600 * 24 * 365)
-      .mul(2); // for the 2 pools
+      .mul(2); // for 2 pools
 
     const extraPremium = expectedEditPremium.sub(expectedRefund);
 
--- a/test/unit/Cover/editCover.js
+++ b/test/unit/Cover/editCover.js
@@ -35,11 +35,16 @@ describe('editCover', function () {
 
     const { expectedPremium, segment, coverId: expectedCoverId } = await buyCoverOnOnePool.call(this, coverBuyFixture);
 
+    const passedPeriod = BigNumber.from(10);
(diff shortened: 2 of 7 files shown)


kyledewy (approved):

kyledewy:
Looks good aside from the broken tests.

contracts/modules/assessment/IndividualClaims.sol
```diff
@@ -291,7 +291,7 @@ contract IndividualClaims is IIndividualClaims, MasterAwareV2 {
         "Invalid claim method for this product type"
       );
       require(requestedAmount <= segment.amount, "Covered amount exceeded");
-      require(segment.start <= block.timestamp, "Cover starts in the future");
+      require(block.timestamp > segment.start, "Cannot buy cover and submit claim in the same block");
```
danoctavian, line 294: feel like it reads better as segment.start < block.timestamp

test/unit/Cover/editCover.js
```diff
@@ -39,7 +39,7 @@ describe('editCover', function () {
 
     const expectedRefund = segment.amount
       .mul(targetPriceRatio)
-      .mul(segment.period)
+      .mul(segment.period - 1)
```
danoctavian, line 42: As far as i understand it:

We use previousExpiration here to compute the refund
https://github.com/NexusMutual/smart-contracts/blob/8cc2511924d23f1d20563ef71063411ecf656db6/contracts/modules/cover/Cover.sol#L334


previousExpiration is computed here as lastSegment.start + lastSegment.period. 

https://github.com/NexusMutual/smart-contracts/blob/8cc2511924d23f1d20563ef71063411ecf656db6/contracts/modules/cover/Cover.sol#L212

With the new change to the solidity code, the start moment is 1 second in the past so the expire is 1 second in the past versus how it used to be.


Therefore in the formula

```        vars.refund =
          previousPoolAllocation.premiumInNXM
          * (allocationRequest.previousExpiration - block.timestamp) // remaining period
          / (allocationRequest.previousExpiration - allocationRequest.previousStart); // previous period```
          
          The numerator is smaller by 1.
          
          But the denominator is the same because both allocationRequest.previousExpiration  and allocationRequest.previousStart are the same.
          
          
          Therefore this -1 is right.

File: contracts/modules/assessment/IndividualClaims.sol (hunk 1 above)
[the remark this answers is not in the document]
Comment (shark0der) [reply contracts/modules/assessment/IndividualClaims.sol:294]: I read it `block.timestamp > segment.start` as "current time is greater than the segment start time" ie "we're in the future"

File: test/unit/Cover/editCover.js [hunk 3]
```diff
@@ -1302,7 +1488,7 @@ describe('editCover', function () {
         coverAsset,
         amount: increasedPoolAmount.mul(2),
         period,
-        maxPremiumInAsset: extraPremium.add(10),
+        maxPremiumInAsset: extraPremium.add(1),
```
danoctavian, line 1491: there's 1 more here.. fixing

test/unit/Cover/editCover.js
danoctavian: Disregard this comment above.

roxdanila (approved):

Issue #418, opened 2022-10-11
Cover: Fix implementation of `getCover()`


roxdanila:
It should only return information regarding the cover, not the segments (i.e. returning the start date of the first segment and end date of the last segment is misleading)

PR #744, opened 2023-02-16
Reimplement getCovers in CoverViewer

roxdanila:
## Context
Closes https://github.com/NexusMutual/smart-contracts/issues/418

Clean up the implementation of getCovers used in the UI:
- restructured the data
- removed product / product type related info that already exists in state
- removed commented code
- removed unused functions

## Test plan
None yet

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/cover/CoverViewer.sol
+++ b/contracts/modules/cover/CoverViewer.sol
@@ -2,33 +2,26 @@
 
 pragma solidity ^0.8.9;
 
-import "solmate/src/tokens/ERC721.sol";
-import "@openzeppelin/contracts-v4/token/ERC20/IERC20.sol";
-import "@openzeppelin/contracts-v4/token/ERC20/utils/SafeERC20.sol";
-import "@openzeppelin/contracts-v4/proxy/beacon/UpgradeableBeacon.sol";
-
+import "../../abstract/Multicall.sol";
 import "../../interfaces/ICover.sol";
-import "../../interfaces/IPool.sol";
-import "../../interfaces/IERC20Detailed.sol";
 import "../../interfaces/INXMMaster.sol";
 
-contract CoverViewer {
-  struct CoverView {
-    uint24 productId;
-    uint16 productType;
-    address yieldTokenAddress;
-    uint96 amountPaidOut;
-    uint amountRemaining;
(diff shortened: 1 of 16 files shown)


contracts/modules/cover/CoverViewer.sol
```diff
@@ -2,200 +2,88 @@
...
+      }
+      covers[i].segments = segments;
     }
-    return coverViews;
+    return covers;
   }
 
-  function getCoverSegments(uint coverId) external view returns (CoverSegment[] memory segments) {
-
+  function getCoverSegments(uint coverId) external view returns (Segment[] memory) {
```
danoctavian, line 67: would be nice if we can unify the code here with the one in `getCovers` at least partially.

Could use a _getCoverSegments that's internal

contracts/modules/cover/CoverViewer.sol
```diff
@@ -2,200 +2,88 @@
...
-    uint96 amountPaidOut;
-    uint amountRemaining;
-    uint coverStart;
-    uint coverEnd;
-    uint8 coverAsset;
-    string coverAssetSymbol;
-    uint8 claimMethod;
-    uint32 gracePeriod;
+
+  struct Segment {
```
danoctavian, line 10: The ICover.CoverSegment also contains:

  uint24 globalRewardsRatio;
  uint24 globalCapacityRatio;
  
  I don't se an immediate use case for it now, something to consider adding here

danoctavian (approved):
LGTM

The code unification between getCovers and get getCoverSegments is a nice-to-have.

contracts/modules/cover/CoverViewer.sol
shark0der: I don't think it's worth it.

contracts/modules/cover/CoverViewer.sol
shark0der: Those are for internal use, not really needed outside of the contract. Since it's the CoverViewer we're talking about here we can add any time as it doesn't require a gov vote to upgrade.

shark0der (approved):

Issue #645, opened 2023-01-22
TokenController accounting issue - excess of 13k NXM in TC + excess in governance rewards


danoctavian:
Migration test verification for balance of TC doesn't check out as expected.

What should happen:

V1
Token Controller
- claim assessment stakes - will be sent to their owners
- cover notes - stay in TC

Claim Rewards
- governance rewards - will be moved to TC
- claim assessment rewards - will be sent to their owners

———————

V2
Token Controller
- cover notes
- governance rewards

Balance expectations at the end:

balance(CR) = 0 
balance(TC) = cover notes + gov rewards

What actually happens:

balance(CR) = 0 
balance(TC) = cover notes + gov rewards + ~7000 NXM

https://github.com/NexusMutual/smart-contracts/blob/nexus-v2/test/fork/migration-v2.js#L632

  // TODO: this does NOT pass. Find out where the extra 7k tokens is from.
    // The outputs of the above log:
    // {
    //   tcNxmBalance: '21186831578421870058919',
    //   rewardsSum: '870391213513961173071',
    //   coverNotesSum: '13324809641365910004774'
    // }
    // expect(tcNxmBalance).to.be.equal(rewardsSum.add(coverNotesSum));

Furthermore, there is a smaller discrepancy in the gov rewards tokens with an extra 236 NXM.

```expect(governanceRewardsMigrated).to.be.equal(this.governanceRewardsSum);```

Yields:

  -1106654884061072517264
  +870391213513961173071
  
  
  Extra tokens:
  
  236.26367054711136 NXM in Governance rewards

PR #750, opened 2023-02-20
Chore: TC accounting

roxdanila:
## Context
Closes https://github.com/NexusMutual/smart-contracts/issues/645

Refactor of how we do accounting of tokens for migration, with the most notable changes being:
- don't automatically send the stakes to claim assessors
- script to fetch all tokens locked in TC
- assertions that GVV + CLA rewards = balance(CR) before migration
- assertions that TC locked tokens  = balance(TC) before migration

## Test plan
Ran the fork test

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


scripts/v2-migration/get-governance-rewards.js
```diff
@@ -0,0 +1,121 @@
...
+  const voteLogs = await provider.getLogs(filter);
+  const addresses = voteLogs
+    .map(log => {
+      const data = governance.interface.parseLog(log);
+      const { from } = data.args;
+      return from.toLowerCase();
+    })
+    .filter(onlyUnique);
+
+  // Computed using Dune: https://dune.com/queries/2012946
```
danoctavian, line 52: would it be possible to do this query here dinamically?

Are we considering hardcoding here ok because delegations are disabled?

test/fork/migration-v2.js
```diff
@@ -616,36 +595,7 @@ describe('V2 upgrade', function () {
...
-      tcNxmBalance: tcNxmBalance.toString(),
-    });
-
-    const totalToProcess = claimAssessors.length;
-    console.log(`Processing withdrawClaimAssessmentTokens for ${totalToProcess} claim assesors`);
-    let amountProcessed = 0;
-    while (claimAssessors.length > 0) {
-      const batchSize = 100;
-      const batch = claimAssessors.splice(0, batchSize);
-      await this.tokenController.withdrawClaimAssessmentTokens(batch);
```
danoctavian, line 637: Are we disabling this to avoid gas costs or there are other reasons?

roxdanila: I don't see good reasons to do it. Any manual action is prone to errors, and we need to maintain the data structures and logic around locked tokens anyway due to the CN deposits that are stored similarly. I'm trying to minimize error surface first and foremost.

scripts/v2-migration/get-governance-rewards.js
roxdanila: Delegations are disabled, so this list won't change

test/fork/migration-v2.js
roxdanila: FYI I just found out there were 2 duplicates in the list of generated addresses...

scripts/v2-migration/get-governance-rewards.js
roxdanila: Also, we're using this list just for assertions in the test, we're not doing any magic txs

danoctavian: cool that's fine then

danoctavian:
Still needs an update for the lint

scripts/v2-migration/get-tc-locked-amount.js
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }
+    console.log(
+      `Found ${
+        Object.keys(memberLockedNXM).length
+      } members with locked NXM in TC; processed a batch of 200/${memberCount}`,
+    );
+  }
+
+  console.log('Writing output to output/tc-locked-amount.json...');
```
danoctavian, line 71: when logging it's better to use the OUTPUT_FILE constant in case value changes for output file name

test/fork/migration-v2.js
```diff
@@ -584,6 +585,46 @@ describe('V2 upgrade', function () {
...
+      daiBalanceDiff: formatEther(daiBalanceAfter.sub(this.daiBalanceBefore)),
+      stEthBalanceBefore: formatEther(this.stEthBalanceBefore),
+      stEthBalanceAfter: formatEther(stEthBalanceAfter),
+      stEthBalanceDiff: formatEther(stEthBalanceAfter.sub(this.stEthBalanceBefore)),
+      enzymeSharesBalanceBefore: formatEther(this.enzymeSharesBalanceBefore),
+      enzymeSharesBalanceAfter: formatEther(enzymeSharesBalanceAfter),
+      enzymeSharesBalanceDiff: formatEther(enzymeSharesBalanceAfter.sub(this.enzymeSharesBalanceBefore)),
+    });
+
+    // Why 2 wei difference?
```
danoctavian, line 615: there's still no explanation for this right?

test/fork/migration-v2.js
```diff
@@ -655,55 +667,100 @@ describe('V2 upgrade', function () {
...
+    // Withdraw CN token for all members
+    for (const lock of CN_LOCKED_AMOUNT_OUTPUT) {
+      const memberBalanceBefore = await this.nxm.balanceOf(lock.member);
+      await this.tokenController.withdrawCoverNote(lock.member, lock.coverIds, lock.lockReasonIndexes);
+      const memberBalanceAfter = await this.nxm.balanceOf(lock.member);
+      expect(memberBalanceAfter.sub(memberBalanceBefore)).to.be.equal(lock.amount);
 
-    const rewardsSum = this.governanceRewardsSum;
+      const lockReasonsCount = (await this.tokenController.getLockReasons(lock.member)).length;
+      expect(lockReasonsCount).lte(1);
```
danoctavian, line 731: what is thee reason for asserting this ?

lockReasonCount <= 1

test/fork/migration-v2.js
```diff
@@ -584,6 +585,46 @@ describe('V2 upgrade', function () {
...
+      stEthBalanceDiff: formatEther(stEthBalanceAfter.sub(this.stEthBalanceBefore)),
+      enzymeSharesBalanceBefore: formatEther(this.enzymeSharesBalanceBefore),
+      enzymeSharesBalanceAfter: formatEther(enzymeSharesBalanceAfter),
+      enzymeSharesBalanceDiff: formatEther(enzymeSharesBalanceAfter.sub(this.enzymeSharesBalanceBefore)),
+    });
+
+    // Why 2 wei difference?
+    expect(poolValueDiff.abs(), 'Pool value in ETH should be the same').lessThanOrEqual(BigNumber.from(2));
+    expect(stEthBalanceAfter.sub(this.stEthBalanceBefore).abs(), 'stETH balance should be the same').lessThanOrEqual(
+      BigNumber.from(2),
```
danoctavian, line 618: is there a 2 wei difference here?

if there is then it causes the overall pool value 2 wei diff

now as to why this happens, it could be some strangeness in the stETH internal math, i do remember they had off-by-1 wei errors when depositing ETH. 

you'd expect this to be exactly equal bc its a transfer but i wouldnt sweat too hard over it

test/fork/migration-v2.js
```diff
@@ -173,82 +184,72 @@ describe('V2 upgrade', function () {
...
+      claRewardsSum: formatEther(this.claRewardsSum),
+      totalRewardsInCR: formatEther(this.governanceRewardsSum.add(this.claRewardsSum)),
+      crBalance: formatEther(crBalance),
+      extraAmount: formatEther(crBalance.sub(this.governanceRewardsSum.add(this.claRewardsSum))),
+    });
 
-    this.governanceRewardsSum = governanceRewardsSum;
+    // Currently there are still 1.655901756826619689 NXM extra when comparing
+    // the sum of governance rewards and claim assessment rewards with the balance of CR
+    // expect(crBalance).to.be.equal(governanceRewardsSum.add(claRewardsSum));
```
danoctavian, line 235: we might want to enable this with a 2 nxm error tolerance which is high, but in case something does break it and make it worse at least we notice something broke

danoctavian:
relatively minor observations and suggestions, overall LGTM

test/fork/migration-v2.js
roxdanila: When we withdraw CLA stakes, we don't delete the reasons, just mark the entry as "claimed"

test/fork/migration-v2.js
roxdanila: the stETH shares difference varies between 1 and 2 wei - unsure why tbh, but didn't have time to dig into their transfer function. could be related.

danoctavian: yes that may be it, also some computation against the ETH backing amount, if you deposit 1 ETH it may give you 0.99999.. stETH.

we can live with this i think.

PR #753, opened 2023-02-21
Feature: totalBalanceOf and delegations

### LINKED ISSUE #558: Count pool manager NXM balance (including delegations) in `totalBalanceOf` function
- switch membership should be aware if the user is a manager
- prevent withdraw membership if the user is a manager
- transfer staking pool ownership to new user for each pool

shark0der:
## Context

Fixes #558: voting delegation and staking pool manager tracking.

Include a summary of the change.

## Test plan

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/modules/token/TokenController.sol
```diff
@@ -391,24 +413,154 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
+    _assignStakingPoolManager(poolId, manager);
+  }
+
+  /// Creates a ownership transfer offer for a staking pool
+  /// The offer can be accepted by the proposed manager before the deadline expires
+  ///
+  /// @param poolId           id of the staking pool
+  /// @param proposedManager  address of the proposed manager
+  /// @param deadline         timestamp after which the offer expires
+  function createStakingPoolOwnershipOffer(
```
danoctavian, line 502: What's the reason for creating this offer -> accept flow instead of just transferring the ownership based on the current manager's decision?

Unclear from comments and original ticket

contracts/modules/legacy/LegacyPooledStaking.sol
```diff
@@ -1422,20 +1422,20 @@ contract LegacyPooledStaking is IPooledStaking, MasterAwareV2 {
...
-      migrationData.stakerAddress,
+    (uint stakingPoolId, address stakingPoolAddress) = cover.createStakingPool(
       migrationData.isPrivatePool,
       migrationData.initialPoolFee,
       migrationData.maxPoolFee,
       params,
       migrationData.ipfsDescriptionHash
     );
 
+    tokenController().assignStakingPoolManager(stakingPoolId, migrationData.stakerAddress);
```
danoctavian, line 1437: migrationData has a parameter called migrationData.managerAddress

The fact that we don't use it here properly is a big that is fixed by the PR currently in progress.


```tokenController().assignStakingPoolManager(stakingPoolId, migrationData.managerAddress);```

File: contracts/modules/token/TokenController.sol [hunk 3]
```diff
@@ -189,36 +199,48 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
-    (
-      uint assessmentStake,
-      /*uint104 rewardsWithdrawableFromIndex*/,
-      /*uint16 fraudCount*/
-    ) = assessment().stakeOf(_of);
+    if (includeManagedStakingPools) {
+      uint managedStakingPoolCount = managerStakingPools[_of].length;
+      for (uint i = 0; i < managedStakingPoolCount; i++) {
+        uint poolId = managerStakingPools[_of][i];
+        amount += stakingPoolNXMBalances[poolId].deposits;
```
danoctavian, line 239: how about the .rewards why not add those too here?

contracts/modules/token/TokenController.sol
shark0der: Prevents denial of service attacks against members as you could create and transfer ownership of multiple staking pools to a particular address which would prevent them from voting in governance as the function wouldn't fit in a block. Fix is also applied when creating a staking pool (cannot assign a random address as a manager - msg.sender is used).

contracts/modules/legacy/LegacyPooledStaking.sol
shark0der: Was not aware that `stakerAddress` and `managerAddress` could be different so I just removed one of them. Do you want me to fix this here or you want to fix it in your PR?

contracts/modules/token/TokenController.sol
shark0der: Technically claims and governance rewards were never accounted for here, btw, main reason being gas costs. PooledStaking rewards were - but not in a delegation context, just your rewards. From the discussion it seems to be preferred to not delegate the rewards - so we'll leave that one out. Accounting for your own staking rewards in `totalBalanceOf` would be costly, so we're leaving that one out as well - it will be consistent across the system not to include rewards in the `totalBalanceOf`.

contracts/modules/legacy/LegacyPooledStaking.sol
danoctavian: it's a simple fix of just replacing it with managerAddress, i'd do it here as well 

code will get merged after should be the same val

contracts/modules/token/TokenController.sol
danoctavian: Understood, here it would be cheap to add obviously but for consistency, it would need to always use rewards to be delegated, that makes sense.

contracts/modules/token/TokenController.sol
danoctavian: Can we get some value range for what is an acceptable max for pool count, gas-wise, give we have those 2 nested loops here and the voting itself?

Doesn't need to be enforced but it would give us an idea.

shark0der: - There aren't any nested loops, it's just a single for loop
- Inside the loop we're reading to storage variables:
```
uint poolId = managerStakingPools[_of][i];
amount += stakingPoolNXMBalances[poolId].deposits;
```
This would place us in the range of hundreds (if not thousands) of staking pools - not a real concern given the staking pool ownership transfer has to be accepted.

contracts/modules/legacy/LegacyPooledStaking.sol
shark0der: Done

contracts/modules/token/TokenController.sol
danoctavian: Awesome hundreds/thousands sounds great. Plenty of room for normal users to not be concerned about this value.

Only people pushing the system limits will actually worry about it.

contracts/modules/governance/Governance.sol
```diff
@@ -80,7 +78,6 @@ contract Governance is IGovernance, LegacyMasterAware {
 
   bool internal actionParamsInitialised;
   uint internal actionWaitingTime;
-  IMinimalCover internal cover;
```
danoctavian, line 83: this was a newly added storage slot i presume and now we're rolling it back.

shark0der: It didn't made it to prod so there should be no issue. Let me know if you notice otherwise

contracts/modules/governance/MemberRoles.sol
```diff
@@ -207,24 +207,33 @@ contract MemberRoles is IMemberRoles, Governed, MasterAwareV2 {
...
+
+    require(isMember(msg.sender), "MemberRoles: Caller is not a member");
+
+    require(
+      block.timestamp > token.isLockedForMV(msg.sender),
+      "MemberRoles: Locked for voting in governance"
+    );
+
+    require(
+      tokenController().isStakingPoolManager(msg.sender) == false,
```
danoctavian, line 222: say you have a staking pool, you want to withdraw your address from membership.

The flow is to prepare another address, send an offer for staking pool ownership and wait for it to be accepted.

In an ideal case i'm thinking i'd want to be able to just exit the system if my staking pools have no NXM stake, without having to go though this ownership transfer flow (which involves another member or me with another member address to accept a potentially empty staking pool).

This is somewhat of an edge case since its termination, i wanted to walk through it to see if we are all on-board with how it goes.

contracts/modules/governance/Governance.sol
danoctavian: yes lgtm

contracts/modules/governance/MemberRoles.sol
roxdanila: valid point. unsure if there is an easy way to do so, but regardless i suggest creating an issue for this and treating it after the launch

danoctavian: Agreed let's make the issue.

roxdanila (approved):

PR #770, opened 2023-02-28
Reduce cover size & update max cover period to 365 days

roxdanila:
## Context

Latest PRs increased Cover size past limit again.

- make more constants private
- remove IC unused dependency
- made max cover period 365 days instead of 364 (with @shark0der 's blessing)

## Test plan
Ran existing tests

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -9,7 +9,6 @@ import "@openzeppelin/contracts-v4/token/ERC20/utils/SafeERC20.sol";
 import "../../abstract/MasterAwareV2.sol";
 import "../../interfaces/ICover.sol";
 import "../../interfaces/ICoverNFT.sol";
-import "../../interfaces/IIndividualClaims.sol";
 import "../../interfaces/IPool.sol";
 import "../../interfaces/IStakingNFT.sol";
 import "../../interfaces/IStakingPool.sol";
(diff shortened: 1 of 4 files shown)


contracts/modules/cover/Cover.sol
```diff
@@ -63,9 +62,9 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
   uint private constant GLOBAL_CAPACITY_DENOMINATOR = 10_000;
   uint private constant REWARD_DENOMINATOR = 10_000;
 
-  uint public constant MAX_COVER_PERIOD = 364 days;
+  uint private constant MAX_COVER_PERIOD = 365 days;
```
danoctavian, line 65: why is there a value increase here?

Doesn't this affect the bucket size computation with 91 * 4 = 364?

shark0der: Most of the time we'll be mid-tranche (not bucket). This means that we'll use 4 tranches for a 1-year cover anyway so the 364 days max only helps in very few cases.

danoctavian: sorry meant to say tranche, ok understood

danoctavian (approved):
LGTM

PR #776, opened 2023-03-01
Replace ALLOCATION_UNITS_PER_NXM function with mock in buyV2Covers.js

Gideonnn:
Really small fix for the `buyV2Covers.js` script. Since the `ALLOCATION_UNITS_PER_NXM` function of the `Cover` contract is private now, it is mocked.


mixplore (approved):
Thx for fixing this 🙌

PR #805, opened 2023-03-10
Fix: Revert if trying to initialize new pool with forbidden or non-existing products

### LINKED ISSUE #803: StakingProducts: When adding initial staking products to a newly created staking pool, we should check for each product that it exists & is allowed for the pool

kyledewy:
## Context

Fixes #803 - When adding initial staking products to a newly created staking pool, we should check for each product that it exists & is allowed for the pool

This PR checks that each product is allowed to be initialized for newly created staking pools in `Cover.sol`. The check is made for migrations and for new pool creations. 

Changes:
-  reverts with `PoolNotAllowedForThisProduct(uint productId)` if `isPoolAllowed()` returns false
    - (this should happen when the product doesn't exist, product is deprecated or the poolId isn't found in the allow list)
- changed isPoolAllowed to also return false when the product doesn't exist

## Test plan
I added a unit and integration test in `createStakingPool.js` and a fork test at `test/fork/staked-product-authorization.js`

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


test/integration/Cover/createStakingPool.js
```diff
@@ -121,4 +122,107 @@ describe('createStakingPool', function () {
...
+    await expect(
+      cover.connect(manager).createStakingPool(
+        false, // isPrivatePool,
+        DEFAULT_POOL_FEE, // initialPoolFee
+        DEFAULT_POOL_FEE, // maxPoolFee,
+        [nonExistingProduct], // products
+        '', // ipfsDescriptionHash
+      ),
+    ).to.be.revertedWithPanic(errors.INVALID_ARRAY_ACCESS);
+    // TODO: do we want an explicit error for the non-existing product?
```
kyledewy, line 142: Note that there is no explicit existence check, but it currently tries to index the products array at an invalid index and reverts.

roxdanila:
@kyledewy converted back to draft until you add context, proposed changes, etc.

contracts/modules/cover/Cover.sol
```diff
@@ -781,12 +800,13 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     return uint(activeCover[assetId].totalActiveCoverInAsset);
   }
 
-  function isPoolAllowed(uint productId, uint poolId) external view returns (bool) {
+  // Returns true if the pool exists and is allowed to have the product
+  function isPoolAllowed(uint productId, uint poolId) public view returns (bool) {
```
kyledewy, line 804: Do we want to check for deprecated products here? Cover purchases will be blocked for the product, so not sure if it's worth the extra gas.

contracts/modules/cover/Cover.sol
```diff
@@ -501,6 +501,10 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     string calldata ipfsDescriptionHash
   ) external whenNotPaused returns (uint /*poolId*/, address /*stakingPoolAddress*/) {
 
+    uint numProducts = productInitParams.length;
+
+    (uint poolId, address stakingPoolAddress) = stakingPoolFactory.create(address(this));
```
danoctavian, line 506: was this moved here just to have access to the poolId number?

Could also be predicted because it's always +1

You want to run validation first before doing the actual updates (gas reasons, security reasons), so i'd say let's see if we can keep this after validation

contracts/modules/cover/Cover.sol
```diff
@@ -781,17 +800,22 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+  function isPoolAllowed(uint productId, uint poolId) public view returns (bool) {
 
     uint poolCount = allowedPools[productId].length;
 
+    // If no pools are specified, nothing is blacklisted
     if (poolCount == 0) {
-      return true;
+      // If the product is deprecated, it's not allowed
+      // If the product doesn't exist, it's not allowed
+      return  productId < _products.length && !_products[productId].isDeprecated;
```
danoctavian, line 812: still thinking this can be a separate predicate

It can be an internal function so it gets inlined

danoctavian (changes requested):
Happy to argue over the obervations made but to me some changes are necessary

test/integration/Cover/createStakingPool.js
```diff
@@ -1,14 +1,15 @@
...
-const { AddressZero } = ethers.constants;
+const { AddressZero, One } = ethers.constants;
 const { parseEther } = ethers.utils;
 
 function calculateTrancheId(currentTime, period, gracePeriod) {
   return Math.floor((currentTime + period + gracePeriod) / (91 * 24 * 3600));
 }
 
-const DEFAULT_POOL_FEE = '5';
+const DEFAULT_POOL_FEE = One.mul(5);
```
danoctavian, line 12: BigNumber.from() isn't the same thing?

Its a bit odd to write it like this

contracts/modules/cover/Cover.sol
```diff
@@ -672,6 +690,14 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
     return _productTypes.length;
   }
 
+  function isProductDeprecated(uint productId) internal view returns (bool) {
```
danoctavian, line 693: here i think i wasn't super clear in my original comment i meant to say if we take this check out of isPoolAllowed then it makes sense to make it an internal function

as it's still inside of there and used in 1 place i'd just inline

_products[productId].isDeprecated

because it's only used in 1 place 

My bad if i wasn't clear enough.

contracts/modules/cover/Cover.sol
```diff
@@ -781,20 +807,26 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
+  // Returns true if the product exists and the pool is authorized to have the product
+  function isPoolAllowed(uint productId, uint poolId) public view returns (bool) {
 
-    uint poolCount = allowedPools[productId].length;
-
-    if (poolCount == 0) {
-      return true;
-    }
+    // If product exists and is not deprecated, check if the pool is allowed
+    if (doesProductExist(productId) && !isProductDeprecated(productId)) {
```
danoctavian, line 814: I think we should just inline

return _products[productId].isDeprecated

Initially i thought you're removing this out of isPoolAllowed then it doesn't make sen se to be a separate function mainly cause it's only used here.

My bad if i didn't explain this clearly - if you choose to keep those checks here which i think is fine lets inline the code since those internal functions are not used somewhere else

contracts/modules/cover/Cover.sol
```diff
@@ -781,20 +807,26 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-    }
 
+      for (uint i = 0; i < poolCount; i++) {
+        if (allowedPools[productId][i] == poolId) {
+          // If allowed pools are specified, the product must already be initialized
+          return true;
+        }
+      }
+    }
+    // product doesn't exist or is deprecated
```
danoctavian, line 829: let's fix the comment so it matches what happens

basically the pool is not supported

shark0der:
Here's a couple of comments:

- weight adjustment (particularly weight reduction) for a deprecated product reverts, it should not.

- stake migration from v1 pooled staking to a new v2 pool will not be used anymore, that logic should be nuked from `Cover.createStakingPool` entirely.

- checking if the pool is allowed for a specific product when we're creating the pool is pointless: the pool is about to be created, it does not exist yet, hence we never marked the product as being allowed to be added to that pool as anyone will be able to backrun our whitelist tx and create a pool with that id, in other words - ownership of that pool id is not guaranteed it yet so we don't whitelist unless the pool is already created. the only check that makes sense here is to check that the product is not bound to any pools at all (= allowed on any pool) using `allowedPools[productId].length == 0`. technically, however, this check is done early inside `isPoolAllowed` but it would be much cheaper to only check that the product is allowed on all pools because we avoid some storage reads (pool count for example) so it makes it cheaper overall.

- when editing a product, product existence is checked twice (inside both `isPoolAllowed` and `getPriceAndCapacityRatios`)
  - the product existence check should not happen inside `isPoolAllowed` but in the `createStakingPool` function (directly inside the for loop that calls `isPoolAllowed` seems to be a good place)
  - if the product existence check remains inside `isPoolAllowed` (in case the deduplication is solved differently) the function name should reflect it. probably to just better pull it out of this function.

kyledewy:
Changes sound great to me. Will get started on them.
Another thing I realized is that we could change `isPoolAllowed(poolId, productId)` to  take a list of productIds to avoid repeatedly making calls to the cover contract during  `StakingProducts.setProducts()` function. Not sure if it's worth it, but it could also be combined into the `getPriceAndCapacityRatios()` call. 
 Let me know if you see any problems with that approach.

contracts/modules/cover/Cover.sol
```diff
@@ -781,20 +799,29 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard {
...
-
-    uint poolCount = allowedPools[productId].length;
+  // Returns true if the product exists and the pool is authorized to have the product
+  function isPoolAllowed(uint productId, uint poolId) public view returns (bool) {
 
-    if (poolCount == 0) {
-      return true;
-    }
+    // If product exists and is not deprecated, check if the pool is allowed
+    if (productId < _products.length && !_products[productId].isDeprecated){
```
danoctavian, line 806: as suggested by @shark0der 

this logic of product existence should be out of this function pulled in the outer calling function

```the product existence check should not happen inside isPoolAllowed but in the createStakingPool function (directly inside the for loop that calls isPoolAllowed seems to be a good place)```

or the function renamed

kyledewy:
I made the changes suggested by @shark0der , along with the optimization I suggested above. The only thing that should be missing is the deprecated check in setProducts yet. Want to write some tests for it as well.

shark0der (approved):

PR #824, opened 2023-03-16
Update bumped price when setting the target price on an existing product

shark0der:
## Context

Complete context on slack.

Update bumped price when settting the target price for an existing product. 

## Test plan

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/test/unit/StakingProducts/setProducts.js
+++ b/test/unit/StakingProducts/setProducts.js
@@ -161,8 +161,8 @@ describe('setProducts unit tests', function () {
     expect(bumpedPriceBefore).to.be.equal(_initialPrices[0]);
     expect(bumpedPriceUpdateTimeBefore).to.be.equal(initialTimestamp);
 
-    await increaseTime(daysToSeconds(2)); // 1% bump
-    const bumpPercentage = BigNumber.from(100); // 1%
+    await increaseTime(daysToSeconds(2)); // 1% drop
+    const priceDrop = BigNumber.from(100); // 1%
 
     // increase targetPrice
     const productEditParams = [{ ...products[0], targetPrice: 2000 }];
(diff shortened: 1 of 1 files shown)


contracts/modules/staking/StakingProducts.sol
```diff
@@ -162,12 +162,32 @@ contract StakingProducts is IStakingProducts, MasterAwareV2, Multicall {
...
           revert TargetPriceTooHigh();
         }
+
         if (_param.targetPrice < globalMinPriceRatio) {
           revert TargetPriceBelowMin();
         }
+
+        // if this is an existing product, when the target price is updated we need to calculate the
+        // current base price using the old target price and update the bumped price to that value
+        // uses the same logic as calculatePremium()
```
danoctavian, line 176: we may want to make a pure internal function if it's the exact same logic to avoid duplicatoj

contracts/modules/staking/StakingProducts.sol
```diff
@@ -162,12 +162,32 @@ contract StakingProducts is IStakingProducts, MasterAwareV2, Multicall {
...
+        if (_product.bumpedPriceUpdateTime != 0) {
```
kyledewy, line 177: Note: I think we need to make a var `isNewProduct` since `product.bumpedPriceUpdateTime` is already set to timestamp above.

test/integration/setup.js
```diff
@@ -511,7 +511,7 @@ async function setup() {
     productId: 0,
     weight: 100,
     initialPrice: 1000,
-    targetPrice: 1000,
+    targetPrice: 100,
```
kyledewy, line 514: This resolved a broken test in `integration/Pool/token-price.js` 
An assertion was off at line 266 `    expect(spotTokenPricePostMCRPosting).to.be.lt(spotTokenPricePreMCRPosting);`
I'm not exactly sure why it effected this. I was only adjusting it because I noticed the initial prices (cover products) were lower than the target price.

test/unit/StakingProducts/setProducts.js
```diff
@@ -141,6 +142,41 @@ describe('setProducts unit tests', function () {
...
+      await stakingProducts.getProduct(poolId, newProductTemplate.productId);
+
+    expect(bumpedPriceBefore).to.be.equal(_initialPrices[0]);
+    expect(bumpedPriceUpdateTimeBefore).to.be.equal(initialTimestamp);
+
+    await increaseTime(daysToSeconds(2)); // 1% bump
+    const bumpPercentage = BigNumber.from(100); // 1%
+
+    // increase targetPrice
+    const productEditParams = [{ ...products[0], targetPrice: 2000 }];
```
danoctavian, line 168: do we have an equivalent test case for a target price decrease?

That's what actually happened in prod

danoctavian:
Are we waiting for something extra or can we move this out of draft stage? 

Seems like extra tests would benefit

test/unit/StakingProducts/setProducts.js
```diff
@@ -141,6 +142,41 @@ describe('setProducts unit tests', function () {
...
+    const { timestamp: initialTimestamp } = await ethers.provider.getBlock('latest');
+
+    const { bumpedPrice: bumpedPriceBefore, bumpedPriceUpdateTime: bumpedPriceUpdateTimeBefore } =
+      await stakingProducts.getProduct(poolId, newProductTemplate.productId);
+
+    expect(bumpedPriceBefore).to.be.equal(_initialPrices[0]);
+    expect(bumpedPriceUpdateTimeBefore).to.be.equal(initialTimestamp);
+
+    await increaseTime(daysToSeconds(2)); // 1% bump
+    const bumpPercentage = BigNumber.from(100); // 1%
```
danoctavian, line 165: can we make this a computed variable based on time passed in the test?

According to

PRICE_CHANGE_PER_DAY * timeSinceLastUpdate / 1 days

test/unit/StakingProducts/setProducts.js
```diff
@@ -1,6 +1,7 @@
 const { expect } = require('chai');
```
danoctavian, line 1: we should have a test here that best captures the error in prod, where the actual price was sitting at the initial targer price until the manager updated the target which caused it to drop and assert around that

danoctavian (changes requested):
it's not out of Draft stage yet but my change requests are around testing code

contracts/modules/staking/StakingProducts.sol
```diff
@@ -162,12 +164,33 @@ contract StakingProducts is IStakingProducts, MasterAwareV2, Multicall {
...
         }
+
         if (_param.targetPrice < globalMinPriceRatio) {
           revert TargetPriceBelowMin();
         }
+
+        // if this is an existing product, when the target price is updated we need to calculate the
+        // current base price using the old target price and update the bumped price to that value
+        // uses the same logic as calculatePremium()
+        if (!isNewProduct) {
```
danoctavian, line 179: how about we only do this if the targetPrice is different?

Seems like with current code path you can have setTargetPrice = true and put the same targetPrice

shark0der: In that case you just don't need `setTargetPrice = true`, no?

test/unit/StakingProducts/setProducts.js
shark0der: Test added

danoctavian (approved):
LGTM

kyledewy (approved):

PR #830, opened 2023-03-22
Chore/compilation fail with warnings

MilGard91:
## Context

Issue: #605 

Added `hardhat-ignore-warnings` which converts all warning to errors and added set -e to the compilation job to fail ci if the compilation fails

## Test plan

Change the size of any contract and compilation should fail

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,9 @@ jobs:
             - node_modules
       - run:
           name: Compile contracts
-          command: npm run compile
+          command: |
+            set -e 
+            npm run compile
       - save_cache:
           key: nexus-mutual-contracts-{{ .Environment.CIRCLE_SHA1 }}
           paths:
--- a/hardhat.config.js/tasks.js
+++ b/hardhat.config.js/tasks.js
@@ -43,3 +43,11 @@ task(TASK_COMPILE)
 
     await runSuper();
   });
+
+task('coverage').setAction(async function (args, hre, runSuper) {
+  hre.config.warnings = {
+    ...hre.config.warnings,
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -19,7 +19,9 @@ jobs:
             - node_modules
       - run:
           name: Compile contracts
-          command: npm run compile
+          command: |
+            set -e
```
danoctavian, line 23: what does this new part achieve here as opposed to the raw command

hardhat.config.js/tasks.js
```diff
@@ -43,3 +43,11 @@ task(TASK_COMPILE)
 
     await runSuper();
   });
+
+task('coverage').setAction(async function (args, hre, runSuper) {
```
danoctavian, line 47: i saw the build failing so it looks like it's working .


My only concern here:

Why is this part of the coverage stage and not another stage?

Compilation fails normally in checkout and install right?

MilGard91: You can try changing the contract with something you know that is a warning (size of the contract) and try running compile. Now it will be an error and it will fail.

for the coverage it needs to have warning instead of the error because of all the mocks and tests. It doesn’t even run the tests if I leave them as error, so only for coverage we will use warnings.

danoctavian: ok understood so basically you are preventing the breakage of this stage so it only fails at compile

danoctavian (approved):
lgtm

roxdanila (approved):

PR #831, opened 2023-03-27
Fix: effectiveWeight should be calculated out of raw capacity, not weighted capacity

danoctavian:
## Context

Issue: https://github.com/NexusMutual/smart-contracts/issues/829

The bug in StakingProducts._getEffectiveWeight affects the computation:

```    uint actualWeight = Math.min(totalAllocation * WEIGHT_DENOMINATOR / totalCapacity, type(uint16).max);```

The `totalCapacity` parameter coming out of `stakingPool.getTrancheCapacities` is multiplied and scaled down in accordance to the targetWeight. However, this is not the way to compute the `actualWeight` which should be calculated with respect to the raw capacity, without involving that product's target weight.

The proposed change is to add a `productTargetWeight` parameter to StakingPool.getTrancheCapacities in order to allow obtaining raw capacity unaffected by productTargetWeight by setting it to 100 in this case. All other usage cases just read the actual value for the the product target weight and pass it in.

## Test plan

Wrote a set of integration tests specifically for `StakingProducts.recalculateEffectiveWeightsForAllProducts`.

Testing of the whole upgrade and of the fixing of current production broken values is done on the fork test.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### CODE REVIEW

File: contracts/modules/staking/StakingPool.sol [hunk 1]
```diff
@@ -849,11 +849,11 @@ contract StakingPool is IStakingPool, Multicall {
   }
 
   function getTrancheCapacities(
-    uint productId,
     uint firstTrancheId,
     uint trancheCount,
     uint capacityRatio,
-    uint reductionRatio
+    uint reductionRatio,
+    uint productTargetWeight
```
danoctavian, line 856: reason for doing this:

Couldn't find a way to reliably reverse the calculation in StakingProducts in the case that productTargetWeight is set to 0 (information loss).

We could still expose the same function (exact same interface) without this parameter and have an internal _ getTrancheCapacities that has the parameter and is used by all other ones.

I didn't see getTrancheCapacities used anywhere yet though so i modified it directly.

test/integration/StakingProducts/recalculateEffectiveWeightsForAllProducts.js
```diff
@@ -0,0 +1,297 @@
...
+          targetWeight: targetWeightAfterCoverBuy,
+          productId,
+        },
+      ]);
+      firstProductExpectedActiveWeight = expectedActiveWeight;
+    }
+
+    let secondProductExpectedActiveWeight;
+    {
+      // Cover inputs for cover purchase used to increase activeWeight
```
danoctavian, line 236: could refactor this block into a function, i duplicated it for now

MilGard91 (commented):
Apart for the ids, lgtm

test/fork/recalculate-effective-weights-bug-fix.js
```diff
@@ -0,0 +1,102 @@
...
+    this.config = await getConfig.call(this);
+  });
+
+  it('should recalculate effective weight for all products in all pools', async function () {
+    const { stakingPoolFactory, stakingProducts } = this;
+
+    const poolCount = await stakingPoolFactory.stakingPoolCount();
+
+    for (let i = 0; i <= poolCount; i++) {
+      await stakingProducts.recalculateEffectiveWeightsForAllProducts(i);
```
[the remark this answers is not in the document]
danoctavian: thanks good point

MilGard91 (approved):

test/fork/recalculate-effective-weights-bug-fix.js
```diff
@@ -0,0 +1,128 @@
+const { ethers } = require('hardhat');
```
danoctavian, line 1: i made a separate file from @kyledewy  to just test this particular upgrade but reused his verification logic.

Happy to adjust if you think there should just be 1.

roxdanila (changes requested):
I think we can find a cleaner solution, this is a bit messy imo

contracts/modules/staking/StakingProducts.sol
```diff
@@ -328,15 +336,10 @@ contract StakingProducts is IStakingProducts, MasterAwareV2, Multicall {
...
-    uint[] memory trancheCapacities = stakingPool.getTrancheCapacities(
-      productId,
-      block.timestamp / TRANCHE_DURATION, // first active tranche id
-      MAX_ACTIVE_TRANCHES,
-      globalCapacityRatio,
-      capacityReductionRatio
-    );
-
-    uint totalCapacity = Math.sum(trancheCapacities);
+    uint activeStake = stakingPool.getActiveStake();
```
danoctavian, line 339: fix implemented as @shark0der  proposed.

shark0der (approved):

roxdanila (approved):
🚢

PR #857, opened 2023-05-03
Fix processExpirations by advancing to next bucket and depositing to next tranche

shark0der:
## Context

Process expirations bucket related tests were failing due to bucket mismatch. The tests assumed the bucket where the rewards expire is the next one, however the cover period did not fit in the current bucket so it was expired in the `currentBucket + 2` instead.

This PR advances the time to the beginning of the next bucket (unconditionally). This however may lead to a different problem where the tranche expires sooner than the bucket which would fail the cover buy (reward generation), hence it changes the deposit to happen on a further tranche.

## Test plan

No new tests required.

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/test/unit/StakingPool/processExpirations.js
+++ b/test/unit/StakingPool/processExpirations.js
@@ -1,7 +1,7 @@
 const { ethers } = require('hardhat');
 const { expect } = require('chai');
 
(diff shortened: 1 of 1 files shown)


danoctavian (approved):
good fix, LGTM

Issue #451, opened 2022-10-27
Add support for ETH in SwapOperator.recoverAsset


danoctavian:
Currently SwapOperator.recoverAsset only supports ERC20 tokens.

Add support for recovering ETH as well.

PR #860, opened 2023-05-04
Fix: SwapOperator denominator bug + enzyme full withdrawal

danoctavian:
## Context

Issues

https://github.com/NexusMutual/smart-contracts/issues/861

https://github.com/NexusMutual/smart-contracts/issues/451

Test how withdrawal of Enzyme would work and fix any bugs related to that.

Also add recovery of ETH from swap operator.

Fix the bug in SwapOperator where the denominator for the slippage calculation is 1e18 instead of 10000 (the correct value) for the 2 Enzyme calls.

Notes:

A withdrawal in any amount requires:

* changing Enzyme vault shares asset details min  to 0  and max to a value that's lower than the current shares amount (eg. 1) respectively.
* Upgrade SwapOperator to fix the denominator bug.
* Get Enzyme to enable the new SwapOperator as a Shares Receiver.
* call swapEnzymeVaultShareForETH with the current amount of shares (15348001519741036985819).

Expecting to receive this amount of ETH back : ```13171239851868541139524``` in wei or roughly ```13171.23985186854``` ETH.

## Test plan

Write a fork test that extracts all ETH from enzyme and clears out all the shares.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/capital/SwapOperator.sol
+++ b/contracts/modules/capital/SwapOperator.sol
@@ -509,12 +509,23 @@ contract SwapOperator {
     // Validate there's no current cow swap order going on
     require(!orderInProgress(), "SwapOp: an order is already in place");
 
+    IPool pool = _pool();
+
+    if (assetAddress == ETH) {
+      uint ethBalance = address(this).balance;
+      require(ethBalance > 0, "SwapOp: Balance = 0");
+
+      // We assume ETH is always supported so we directly transfer it back to the Pool
+      (bool sent, ) = payable(address(pool)).call{value: ethBalance}("");
+      require(sent, "SwapOp: Failed to send Ether to pool");
+
+      return;
+    }
+
     IERC20 asset = IERC20(assetAddress);
 
     uint balance = asset.balanceOf(address(this));
     require(balance > 0, "SwapOp: Balance = 0");
-
-    IPool pool = _pool();
 
     SwapDetails memory swapDetails = pool.getAssetSwapDetails(assetAddress);
 
--- a/contracts/modules/capital/SwapOperator.sol
+++ b/contracts/modules/capital/SwapOperator.sol
@@ -513,7 +513,7 @@ contract SwapOperator {
 
     if (assetAddress == ETH) {
       uint ethBalance = address(this).balance;
-      require(ethBalance > 0, "SwapOp: Balance = 0");
+      require(ethBalance > 0, "SwapOp: ETH balance to recover is 0");
 
       // We assume ETH is always supported so we directly transfer it back to the Pool
       (bool sent, ) = payable(address(pool)).call{value: ethBalance}("");
--- a/contracts/modules/capital/SwapOperator.sol
+++ b/contracts/modules/capital/SwapOperator.sol
@@ -392,7 +392,7 @@ contract SwapOperator {
       (, uint netShareValue) = enzymeFundValueCalculatorRouter.calcNetShareValue(enzymeV4VaultProxyAddress);
 
       uint avgAmountOut = amountIn * 1e18 / netShareValue;
-      uint maxSlippageAmount = avgAmountOut * swapDetails.maxSlippageRatio / 1e18;
+      uint maxSlippageAmount = avgAmountOut * swapDetails.maxSlippageRatio / MAX_SLIPPAGE_DENOMINATOR;
       uint minOutOnMaxSlippage = avgAmountOut - maxSlippageAmount;
 
       require(amountOutMin >= minOutOnMaxSlippage, "SwapOp: amountOutMin < minOutOnMaxSlippage");
@@ -458,7 +458,7 @@ contract SwapOperator {
 
       // avgAmountOut in ETH
       uint avgAmountOut = amountIn * netShareValue / (10 ** fromToken.decimals());
-      uint maxSlippageAmount = avgAmountOut * swapDetails.maxSlippageRatio / 1e18;
+      uint maxSlippageAmount = avgAmountOut * swapDetails.maxSlippageRatio / MAX_SLIPPAGE_DENOMINATOR;
       uint minOutOnMaxSlippage = avgAmountOut - maxSlippageAmount;
 
       // slippage check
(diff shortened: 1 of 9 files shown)


test/fork/withdraw-enzyme-shares.js
```diff
@@ -0,0 +1,146 @@
...
+      '0',
+      '1',
+      '10000' // max slippage ratio
+    );
+
+    await enableAsEnzymeReceiver(this.swapOperator.address);
+
+    const swapAmount = shareAmount; //  parseEther('100');
+    const minOut = shareAmount.mul(84).div(100);
+    await this.swapOperator.connect(swapController).swapEnzymeVaultShareForETH(swapAmount, minOut);
```
danoctavian, line 144: good news this works as soon as we fix the bug.

You get a ratio of the ETH. 85% or so.

contracts/modules/capital/SwapOperator.sol
```diff
@@ -509,13 +509,24 @@ contract SwapOperator {
     // Validate there's no current cow swap order going on
     require(!orderInProgress(), "SwapOp: an order is already in place");
 
+    IPool pool = _pool();
+
+    if (assetAddress == ETH) {
```
danoctavian, line 514: figured we add this here as part of the ugprade so we're prepared for everything

test/fork/withdraw-enzyme-shares.js
roxdanila: you have to use the oracle here instead of this hardcoded value. this is basically the exchange rate

danoctavian: Tried it and added code that uses that but enzyme returns something under the oracle price (in the 20-25 ether range less).

contracts/modules/capital/SwapOperator.sol
```diff
@@ -392,7 +392,7 @@ contract SwapOperator {
       (, uint netShareValue) = enzymeFundValueCalculatorRouter.calcNetShareValue(enzymeV4VaultProxyAddress);
 
       uint avgAmountOut = amountIn * 1e18 / netShareValue;
-      uint maxSlippageAmount = avgAmountOut * swapDetails.maxSlippageRatio / 1e18;
+      uint maxSlippageAmount = avgAmountOut * swapDetails.maxSlippageRatio / MAX_SLIPPAGE_DENOMINATOR;
```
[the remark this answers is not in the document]
roxdanila: so if I read this correctly, the bug makes the slippage much smaller than intended. this hasn't been an issue when we deposited to Enzyme, as the exchange rate was 1:1 - is that right?

contracts/modules/capital/SwapOperator.sol
```diff
@@ -458,7 +458,7 @@ contract SwapOperator {
 
       // avgAmountOut in ETH
       uint avgAmountOut = amountIn * netShareValue / (10 ** fromToken.decimals());
-      uint maxSlippageAmount = avgAmountOut * swapDetails.maxSlippageRatio / 1e18;
+      uint maxSlippageAmount = avgAmountOut * swapDetails.maxSlippageRatio / MAX_SLIPPAGE_DENOMINATOR;
```
roxdanila, line 461: how come we are hitting this issue - isn't the slippage to withdraw 0 anyway?

roxdanila: I agree the issue must be fixed, just trying to understand the behaviour

test/fork/withdraw-enzyme-shares.js
```diff
@@ -0,0 +1,124 @@
...
+      ? new ethers.providers.JsonRpcProvider(network.config.url)
+      : ethers.provider;
+  return provider.getSigner(address);
+};
+
+describe('prevent switch or withdraw membership when tokens are locked', function () {
+  before(async function () {
+    // Initialize evm helper
+    await evm.connect(ethers.provider);
+    await getSigner('0x1eE3ECa7aEF17D1e74eD7C447CcBA61aC76aDbA9');
```
roxdanila, line 30: is this a random address? or the swap controller?

contracts/modules/capital/SwapOperator.sol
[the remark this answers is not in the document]
danoctavian: Yeah Exactly the exchange rate was just fine until the loss event, and at deposit time it was 1:1 so this bug was never hit.

test/fork/withdraw-enzyme-shares.js
danoctavian: It is pretty much random, checked with @shark0der  and it was probably testing the getSigner function and i just copy pasted it along with the whole before block

removed from everywhere now

contracts/modules/capital/SwapOperator.sol
shark0der: The slippage bug is valid, however what you are describing here is not a consequence of the slippage (first of all because we don't have any slippage in fork tests unless we push a new price to chainlink), but rather a consequence of the enzyme withdrawal fees.

The provided fix is correct, but we should also fetch the withdrawal fee from enzyme and apply it on top of `avgAmountOut` and then applying the slippage on top of the result. With the withdrawal fee applied, the fork test should pass with a 0% slippage.

contracts/modules/capital/SwapOperator.sol
shark0der: Addition looks good to me. Would consider adding a method to unconditionally close an order (say we call it `forceCloseOrder`) for the scenario when `cowSettlement.filledAmount(...)`, `cowSettlement.setPreSignature(...)` or other functions that are called in `closeOrder` are reverting. The only conditions I would leave in place in `forceCloseOrder` would be `validateUID(...)` and `order.validTo` check. Could totally be tackled in a separate issue.

contracts/modules/capital/SwapOperator.sol
danoctavian: Following the indications from Enzyme's team, i learned that to get the correct net share value we can basically trigger a small redemption (2 wei in this case) which triggers net share value recomputation.

I trigger that here:
https://github.com/NexusMutual/smart-contracts/pull/860/files#diff-eca555b76e500a76402b3318a3186118d210bf7a1157b2302b8b0e9e3b144ae4R462

That value is equal to gross share value after and they are both reflective of the fee applied (no fee application is needed.).


Once this happens i then execute the swap for the full amount.

shark0der: Following yesterday's chat discussion we settled for the oracle + slippage so I'm resolving this conversation.

shark0der (approved):

PR #868, opened 2023-05-15
Feature: Expand Cover contract with `expireCover`

MilGard91:
## Context

When cover expires is no way to refresh the capacity before the bucket expires

Issue: #863 

Added `expireCover` that will expire cover  even though the bucket didn't expire, so we can use the capcaity.

## Test plan

Added tests for `expireCover` function in integration tests.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -281,17 +281,28 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
   }
 
   function expireCover(uint coverId) external {
+
     uint segmentId = _coverSegments[coverId].length - 1;
     CoverSegment memory lastSegment = coverSegmentWithRemainingAmount(coverId, segmentId);
     CoverData memory cover = _coverData[coverId];
     uint expiration = lastSegment.start + lastSegment.period;
 
-    // require last segment to be expired
     if (expiration > block.timestamp) {
       revert CoverNotYetExpired(coverId);
     }
 
-    for (uint allocationIndex = 0; allocationIndex < coverSegmentAllocations[coverId][segmentId].length; allocationIndex++) {
+    uint currentBucketId = block.timestamp / BUCKET_SIZE;
+    uint bucketAtExpiry = Math.divCeil(expiration, BUCKET_SIZE);
+
+    if (currentBucketId >= bucketAtExpiry) {
+      revert CoverAlreadyExpired(coverId);
+    }
+
+    for (
+      uint allocationIndex = 0;
+      allocationIndex < coverSegmentAllocations[coverId][segmentId].length;
+      allocationIndex++
+    ) {
       PoolAllocation memory allocation =  coverSegmentAllocations[coverId][segmentId][allocationIndex];
       AllocationRequest memory allocationRequest;
       // editing just the needed props for deallocation
@@ -307,8 +318,8 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
       );
 
     }
+
     // remove cover amount from from expiration buckets
-    uint bucketAtExpiry = Math.divCeil(expiration, BUCKET_SIZE);
     activeCoverExpirationBuckets[cover.coverAsset][bucketAtExpiry] -= lastSegment.amount;
   }
 
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -140,9 +140,9 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
 
       Product memory product = _products[params.productId];
 
-       if (product.isDeprecated) {
-         revert ProductDeprecated();
-       }
+      if (product.isDeprecated) {
+        revert ProductDoesntExistOrIsDeprecated();
+      }
 
       if (!isCoverAssetSupported(params.coverAsset, product.coverAssets)) {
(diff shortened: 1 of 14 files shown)


contracts/modules/cover/Cover.sol
```diff
@@ -286,6 +279,40 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
     emit CoverEdited(coverId, params.productId, segmentId, msg.sender, params.ipfsData);
   }
 
+  function expireCover(uint[] calldata coverIds) external onlyMember {
+    for (uint i=0; i < coverIds.length; i++) {
```
roxdanila, line 283: one potential contract size optimization here would be to take as input only 1 cover id, instead of looping through this array. it's not ideal, but if we can keep the Multicall in, it's not that bad

contracts/modules/cover/Cover.sol
```diff
@@ -286,6 +279,40 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+      uint segmentId = _coverSegments[coverId].length - 1;
+      CoverSegment memory lastSegment = coverSegmentWithRemainingAmount(coverId, segmentId);
+      CoverData memory cover = _coverData[coverId];
+      uint expiration = lastSegment.start + lastSegment.period;
+
+      // require last segment to be expired
+      if (expiration > block.timestamp) {
+        revert CoverNotYetExpired(coverId);
+      }
+      for(uint j = 0; j < coverSegmentAllocations[coverId][segmentId].length; j++) {
```
roxdanila, line 294: always leave a space before `(`

contracts/modules/cover/Cover.sol
roxdanila, line 283: always leave a space beforer and after `=`

contracts/modules/cover/Cover.sol
```diff
@@ -133,35 +133,29 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
       }
 
       Product memory product = _products[params.productId];
 
-      if (product.isDeprecated) {
-        revert ProductDeprecated();
-      }
+       if (product.isDeprecated) {
+         revert ProductDeprecated();
+       }
```
shark0der, line 145: What exactly do you think you're doing here? :sweat_smile:

contracts/modules/cover/Cover.sol
```diff
@@ -286,6 +280,37 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+      stakingPool(allocation.poolId).requestAllocation(
+        0, // amount
+        0, // previous premium
+        allocationRequest
+      );
+
+    }
+    // remove cover amount from from expiration buckets
+    uint bucketAtExpiry = Math.divCeil(expiration, BUCKET_SIZE);
+    activeCoverExpirationBuckets[cover.coverAsset][bucketAtExpiry] -= lastSegment.amount;
```
shark0der, line 312: We also need to check if the bucket was not processed already otherwise we'd be subtracting twice.

test/integration/Cover/expireCover.js
```diff
@@ -0,0 +1,170 @@
...
+  it('should revert when cover is not expired', async function () {
+    const { cover } = this.contracts;
+    const [coverBuyer] = this.accounts.members;
+    const { amount } = buyCoverFixture;
+    const coverBuyerAddress = await coverBuyer.getAddress();
+
+    await cover
+      .connect(coverBuyer)
+      .buyCover(
+        { ...buyCoverFixture, commissionDestination: coverBuyerAddress, owner: coverBuyerAddress },
```
shark0der, line 74: Why do you need to specify the commission destination and owner and not use the default from the template?

test/integration/Cover/expireCover.js
```diff
@@ -0,0 +1,170 @@
...
+
+    const coverId = await cover.coverDataCount();
+
+    await expect(cover.connect(coverBuyer).expireCover(coverId)).to.be.revertedWithCustomError(
+      cover,
+      `CoverNotYetExpired`,
+    );
+  });
+
+  it('should expire an cover', async function () {
```
shark0der, line 87: `a cover`

contracts/modules/cover/Cover.sol
```diff
@@ -133,35 +133,29 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
-        GLOBAL_MIN_PRICE_RATIO
-      );
+      allocationRequest.productId = params.productId;
+      allocationRequest.coverId = coverId;
+      allocationRequest.period = params.period;
+      allocationRequest.gracePeriod = _productTypes[product.productType].gracePeriod;
+      allocationRequest.globalCapacityRatio = GLOBAL_CAPACITY_RATIO;
+      allocationRequest.capacityReductionRatio = product.capacityReductionRatio;
+      allocationRequest.rewardRatio = GLOBAL_REWARDS_RATIO;
+      allocationRequest.globalMinPrice = GLOBAL_MIN_PRICE_RATIO;
```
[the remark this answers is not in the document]
MilGard91: This reduced the size of the contract

contracts/modules/cover/Cover.sol
MilGard91: They were commented out, and when I brought them back this happened :/

shark0der (approved):

PR #875, opened 2023-06-13
Fix: allocation carry over

shark0der:
## Context

Issue #872 

Carry over allocations from previous tranches.

## Test plan

Added a test that fails with the current release-candidate and passes with the fix.

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/staking/StakingPool.sol
+++ b/contracts/modules/staking/StakingPool.sol
@@ -924,31 +924,61 @@ contract StakingPool is IStakingPool, Multicall {
 
       uint[] memory trancheCapacities = getTrancheCapacities(
         request.productId,
-        firstTrancheIdToUse,
-        MAX_ACTIVE_TRANCHES - startIndex, // count
+        _firstActiveTrancheId,
+        MAX_ACTIVE_TRANCHES, // count
         request.globalCapacityRatio,
         request.capacityReductionRatio
       );
 
       uint remainingAmount = coverAllocationAmount;
+      uint carryOver;
       uint packedCoverAllocations;
+
+      for (uint i = 0; i < startIndex; i++) {
+
+        uint allocated = trancheAllocations[i];
+        uint capacity = trancheCapacities[i];
+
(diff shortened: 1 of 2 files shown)


contracts/modules/staking/StakingPool.sol
```diff
@@ -922,31 +922,61 @@ contract StakingPool is IStakingPool, Multicall {
 
       uint[] memory trancheCapacities = getTrancheCapacities(
         request.productId,
-        firstTrancheIdToUse,
-        MAX_ACTIVE_TRANCHES - startIndex, // count
+        _firstActiveTrancheId,
```
danoctavian, line 925: what's the logic behind starting with _firstActiveTrancheId vs firstTrancheIdToUse

shark0der: - first active tranche id - exactly what it says, from all active tranches it has the lowest id
- first active to use - from all active ones, the first one that expires after the cover grace period, basically the first one we can use for the current cover

test/unit/StakingPool/requestAllocation.js
```diff
@@ -1530,4 +1530,117 @@ describe('requestAllocation', function () {
...
+        secondCoverAmount.add(1), // slightly over the limit
+        previousPremium,
+        unfullfillableRequest,
+      ),
+    ).to.be.revertedWithCustomError(stakingPool, 'InsufficientCapacity');
+
+    stakingPool.connect(this.coverSigner).requestAllocation(
+      secondCoverAmount, // exact available amount
+      previousPremium,
+      unfullfillableRequest,
```
danoctavian, line 1643: this one is actually fulfillable right?

it doesn't revert

roxdanila (approved):

test/unit/StakingPool/requestAllocation.js
shark0der: Adjusted the variable names.

PR #913, opened 2023-08-04
Add typings

Gideonnn:
Added `typescript` and `tsup` dependencies to export abis and addresses with type information. This can now be ran with `npm run build`.

Now `@nexusmutual/sdk` can re-export this data with included types. This package is now compatible with both commonjs and esm.


deployments/package.json
```diff
@@ -1,16 +1,12 @@
...
-    "./abis": "./dist/abis.js",
-    "./abis/*": "./dist/abis/*.js",
-    "./addresses": "./dist/addresses.js"
-  },
+  "typings": "./dist/index.d.ts",
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
   "files": [
-    "dist",
-    "README.md"
```
Gideonnn, line 13: Removed `README.md` because it is [included by default](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files).

deployments/package.json
```diff
@@ -28,5 +24,12 @@
...
   },
-  "homepage": "https://github.com/NexusMutual/smart-contracts#readme"
+  "homepage": "https://github.com/NexusMutual/smart-contracts#readme",
+  "scripts": {
+    "build": "node pre-publish.js"
+  },
+  "devDependencies": {
+    "tsup": "^7.2.0",
+    "typescript": "^5.1.6"
+  }
```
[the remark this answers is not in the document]
shark0der: The pre-publish script fetches the deps correctly from the parent folder without any magic

deployments/package.json
```diff
@@ -1,16 +1,12 @@
 {
   "name": "@nexusmutual/deployments",
-  "version": "2.1.1",
+  "version": "2.2.0",
```
shark0der, line 3: We keep this in sync with the main package json since the deployment package reflects the smart contracts version so bumping this independently is undersirable.

deployments/src/abis.ts
```diff
@@ -0,0 +1,103 @@
+import Assessment from '../generated/abis/Assessment.json';
```
shark0der, line 1: Is it possible to generate this file on the fly?

deployments/package.json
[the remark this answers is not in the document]
Gideonnn: Yeah I think this is a good idea. This wasn't my favorite part of the pr either

deployments/package.json
Gideonnn: Oh, I didn't know. I did publish it already. Let's do a bump soon and sync up

deployments/src/abis.ts
Gideonnn: Not as easy as we do with other files. Since this is a `.ts` file, it tells typescript/rollup which modules are imported. This info will be used to determine what to treeshake away and what to bundle. So we would need to generate this before running typescript, which is a little too much complexity for what it's worth imo.

What we can do to reduce the maintenance burden is to make a script that generates and overwrites this file for us. This way we can have it in source control, but we can re-generate to check/update anytime we like 🤔

deployments/package-lock.json
shark0der, line 1: Remove this file

deployments/src/abis.ts
shark0der: > What we can do to reduce the maintenance burden is to make a script that generates and overwrites this file for us

That's exactly what I meant! `pre-publish.js` seems a good place for this.

deployments/pre-publish.js
```diff
@@ -122,6 +93,21 @@ const main = async () => {
...
+  console.log('Building source');
+  await build({
+    entry: ['deployments/src/index.ts'],
+    outDir: 'deployments/dist',
+    format: ['cjs', 'esm'],
+    splitting: false,
+    sourcemap: true,
+    clean: true,
+    dts: true,
+    publicDir: 'generated', // copy generated files to dist
```
shark0der, line 105: Is `generated` a temporary folder?

Gideonnn: Yes, it is used to store generated files that are used during compilation. So we build files (like `abis.ts` now) manually and then feed it to `tsup` for compilation. That way we can import these files before they exist like we do [here](https://github.com/NexusMutual/smart-contracts/pull/913/files#diff-e5ae286dd52131f050edbe32b9fcab68d602f076da1a755e8e2a71778cb66124R2).

deployments/src/abis.ts
Gideonnn: Good idea, added!

Issue #886, opened 2023-06-27
Cover Edit: Increase amount


roxdanila:
Should include unit tests

PR #916, opened 2023-08-25
draft: Edit cover increase amount and increase period

danoctavian:
## Context

This pull request is tackling both increasing amount and increasing period through 1 function.

https://github.com/NexusMutual/smart-contracts/issues/886

https://github.com/NexusMutual/smart-contracts/issues/887

Additionally this includes integration tests:

https://github.com/NexusMutual/smart-contracts/issues/890

For the functionality.

The implementation follows the spec as described here for cover edits:

https://www.notion.so/nxmcommunity/Cover-edit-27187e358a9c45b7a2ce721f52bf3a39#12b50533c7f644f392d44f21b3e9069d

The code change involves separating the edit functionality into its own separate function

```requestEditAllocations```

The creation of a new cover is now handled in

```requestNewAllocations```

The ```StakingPool.sol``` contract requires changes to handle the new pricing logic for both extending amount and extending cover.

The pricing of increasing cover amount and period is applied in that order for an edit.

The totalPremium is first evaluated for an increase in amount, using the current remaining period for the segment and is used to compute the extra amount premium to be paid as such:

```premium to be paid = Max((coverAmountInNXMNew - coverAmountOldRepricedInNXM), 0) / coverAmountInNXMNew * totalPremium```

The totalPremium is then evaluated a second time for increasing the period, using both the increased amount and increased period and used to compute the extra amount premium to be paid as such:

```premium to be paid = extraPeriod / newPeriod * totalPremium```

## Test plan

Unit tests for the edit path of cover are handled as part of ```test/unit/Cover/editCover.js```.

Integration tests are handled as part of ```test/integration/Cover/buyCover.js```.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/interfaces/IStakingPool.sol
+++ b/contracts/interfaces/IStakingPool.sol
@@ -18,6 +18,7 @@ struct AllocationRequest {
   uint capacityReductionRatio;
   uint rewardRatio;
   uint globalMinPrice;
+  uint32 extraPeriod;
 }
 
 struct StakedProductParam {
@@ -62,6 +63,12 @@ interface IStakingPool {
     uint128 rewardsShares;
   }
 
+  struct RequestAllocationVariables {
+    uint coverAllocationAmount;
+    uint initialCapacityUsed;
+    uint totalCapacity;
+  }
+
   function initialize(
     bool isPrivatePool,
     uint initialPoolFee,
@@ -74,7 +81,7 @@ interface IStakingPool {
 
   function requestAllocation(
     uint amount,
-    uint previousPremium,
+    uint previousAllocationAmountRepriced,
     AllocationRequest calldata request
   ) external returns (uint premium, uint allocationId);
 
--- a/contracts/mocks/Cover/CoverMockStakingPool.sol
+++ b/contracts/mocks/Cover/CoverMockStakingPool.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.8.18;
 
 import "../../interfaces/IStakingPool.sol";
 import "../../libraries/Math.sol";
+import "hardhat/console.sol";
 
 contract CoverMockStakingPool is IStakingPool {
 
@@ -65,17 +66,26 @@ contract CoverMockStakingPool is IStakingPool {
 
   function requestAllocation(
     uint amount,
-    uint /*previousPremium*/,
+    uint previousAllocationAmountRepriced,
     AllocationRequest calldata request
-  ) external override returns (uint premium, uint allocationId) {
+  ) external override returns (uint extraPremium, uint allocationId) {
 
     usedCapacity[request.productId] += amount;
 
-    premium = request.useFixedPrice
+    uint premium = request.useFixedPrice
       ? calculateFixedPricePremium(amount, request.period, mockPrices[request.productId])
       : calculatePremium(mockPrices[request.productId], amount, request.period);
 
-    return (premium, allocationId);
+    if (amount == 0) {
+      return (0, 0);
+    }
+
+    uint remainingPeriod = request.period - request.extraPeriod;
+    uint extraAmount = amount > previousAllocationAmountRepriced ? amount - previousAllocationAmountRepriced : 0;
+
+    extraPremium = premium * extraAmount * remainingPeriod / request.period / amount +
+      premium * request.extraPeriod / request.period;
+    return (extraPremium, allocationId);
   }
 
   function calculateFixedPricePremium(
--- a/contracts/modules/cover/Cover.sol
+++ b/contracts/modules/cover/Cover.sol
@@ -21,6 +21,8 @@ import "../../libraries/SafeUintCast.sol";
 import "../../libraries/StakingPoolLibrary.sol";
 import "../../interfaces/IStakingProducts.sol";
 import "../../interfaces/ICoverProducts.sol";
+import "hardhat/console.sol";
+import "../../libraries/Math.sol";
 
 contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Multicall {
   using SafeERC20 for IERC20;
@@ -66,6 +68,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
   uint private constant CAPACITY_REDUCTION_DENOMINATOR = 10000;
   uint private constant GLOBAL_CAPACITY_DENOMINATOR = 10_000;
   uint private constant REWARD_DENOMINATOR = 10_000;
+  uint private constant COVER_ALLOCATION_RATIO_DENOMINATOR = 10_000;
 
   uint private constant MAX_COVER_PERIOD = 365 days;
   uint private constant MIN_COVER_PERIOD = 28 days;
@@ -164,6 +167,8 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
       allocationRequest.globalMinPrice = GLOBAL_MIN_PRICE_RATIO;
     }
 
+    uint nxmPriceInCoverAsset = pool().getTokenPriceInAsset(params.coverAsset);
+
     uint previousSegmentAmount;
 
     if (params.coverId == 0) {
@@ -173,9 +178,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
       _coverData[coverId] = CoverData(params.productId, params.coverAsset, 0 /* amountPaidOut */);
 
     } else {
-      revert EditNotSupported();
 
-      /*
       // existing cover
       coverId = params.coverId;
 
@@ -212,17 +215,17 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
       uint bucketAtExpiry = Math.divCeil(lastSegment.start + lastSegment.period, BUCKET_SIZE);
       activeCoverExpirationBuckets[params.coverAsset][bucketAtExpiry] -= lastSegment.amount;
       previousSegmentAmount += lastSegment.amount;
-      */
     }
-
-    uint nxmPriceInCoverAsset = pool().getTokenPriceInAsset(params.coverAsset);
     allocationRequest.coverId = coverId;
 
     (uint coverAmountInCoverAsset, uint amountDueInNXM) = requestAllocation(
       allocationRequest,
       poolAllocationRequests,
-      nxmPriceInCoverAsset,
-      segmentId
+      AllocationParams(
+        nxmPriceInCoverAsset,
+        previousSegmentAmount,
+        segmentId
+      )
     );
 
     if (coverAmountInCoverAsset < params.amount) {
@@ -330,25 +333,39 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
   function requestAllocation(
     AllocationRequest memory allocationRequest,
     PoolAllocationRequest[] memory poolAllocationRequests,
-    uint nxmPriceInCoverAsset,
-    uint segmentId
+    AllocationParams memory params
   ) internal returns (
     uint totalCoverAmountInCoverAsset,
     uint totalAmountDueInNXM
   ) {
 
-    RequestAllocationVariables memory vars = RequestAllocationVariables(0, 0, 0, 0);
+    uint oldSegmentAmountInNXMRepriced = covertAmountInCoverAssetToNXM(
+      params.previousSegmentAmount, params.nxmPriceInCoverAsset
+    );
+
+    console.log("oldSegmentAmountInNXMRepriced", oldSegmentAmountInNXMRepriced);
+
+    RequestAllocationVariables memory vars; // = RequestAllocationVariables(0, 0, 0, 0);
+
     uint totalCoverAmountInNXM;
 
-    vars.previousPoolAllocationsLength = segmentId > 0
-      ? coverSegmentAllocations[allocationRequest.coverId][segmentId - 1].length
-      : 0;
+    vars.previousPoolAllocationsLength = params.segmentId > 0
+     ? coverSegmentAllocations[allocationRequest.coverId][params.segmentId - 1].length
+     : 0;
+
+    vars.allocations = new PoolAllocation[](vars.previousPoolAllocationsLength);
+
+    for (uint i = 0; i < vars.previousPoolAllocationsLength; i++) {
+      vars.allocations[i] = coverSegmentAllocations[allocationRequest.coverId][params.segmentId - 1][i];
+
+      vars.previousCoverAmountTotalInNXM += vars.allocations[i].coverAmountInNXM;
+    }
 
     for (uint i = 0; i < poolAllocationRequests.length; i++) {
       // if there is a previous segment and this index is present on it
       if (vars.previousPoolAllocationsLength > i) {
         PoolAllocation memory previousPoolAllocation =
-          coverSegmentAllocations[allocationRequest.coverId][segmentId - 1][i];
+          coverSegmentAllocations[allocationRequest.coverId][params.segmentId - 1][i];
 
         // poolAllocationRequests must match the pools in the previous segment
         if (previousPoolAllocation.poolId != poolAllocationRequests[i].poolId) {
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

File: contracts/modules/cover/Cover.sol [hunk 1]
```diff
@@ -21,6 +21,8 @@ import "../../libraries/SafeUintCast.sol";
 import "../../libraries/StakingPoolLibrary.sol";
 import "../../interfaces/IStakingProducts.sol";
 import "../../interfaces/ICoverProducts.sol";
+import "../../libraries/Math.sol";
+import "hardhat/console.sol";
```
danoctavian, line 25: these imports and console.log statements will disappear from the final commit history by ammending the commit.

for convenience i benefit from having the logs here to debug certain parts like pricing

if it's an inconvenience when viewing the draft please let me know and i'll adapt

contracts/modules/cover/Cover.sol
```diff
@@ -105,11 +108,13 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
     PoolAllocationRequest[] memory poolAllocationRequests
   ) external payable onlyMember nonReentrant whenNotPaused returns (uint coverId) {
 
-    if (params.period < MIN_COVER_PERIOD) {
+    // check only for a new cover
```
roxdanila, line 111: the duration of the resulting segment should also be between min / max period. 

i.e. if your cover has 5 days left, you can extend it by at least 23 days, so the total period of the new segment is at least 28 days. same for max

danoctavian:
@roxdanila @shark0der 

The PR still has integration tests and unit tests in progress.

I left the console.log statements in for convenience of debugging, and they will be removed from the final commits through squashing (if it's a hindrance please let me know).

contracts/modules/cover/Cover.sol
```diff
@@ -105,11 +108,13 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+    if (params.coverId == 0 && params.period > 0 && params.period < MIN_COVER_PERIOD) {
```
roxdanila, line 112: `params.period > 0` - you don't need this part for new covers

contracts/interfaces/IStakingPool.sol
```diff
@@ -18,6 +18,8 @@ struct AllocationRequest {
   uint capacityReductionRatio;
   uint rewardRatio;
   uint globalMinPrice;
+  uint32 newPeriod;
```
roxdanila, line 21: rename to `extraPeriod`

contracts/interfaces/IStakingPool.sol
```diff
@@ -18,6 +18,8 @@ struct AllocationRequest {
...
+  uint32 remainingPeriod;
```
roxdanila, line 22: do you really need this here? you can technically deduct it based on `previousExpiration`

contracts/modules/cover/Cover.sol
```diff
@@ -151,20 +155,38 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
       allocationRequest.capacityReductionRatio = product.capacityReductionRatio;
       allocationRequest.rewardRatio = GLOBAL_REWARDS_RATIO;
       allocationRequest.globalMinPrice = GLOBAL_MIN_PRICE_RATIO;
+
+      // if it's the first segment the remaining period and new period are the requested period
+      allocationRequest.remainingPeriod =  uint32(allocationRequest.period);
+      allocationRequest.newPeriod =  uint32(allocationRequest.period);
```
roxdanila, line 161: extra spaces

contracts/interfaces/IStakingPool.sol
roxdanila, line 21: rename to `extraPeriod` to reflect what it actually is

contracts/interfaces/IStakingPool.sol
danoctavian, line 22: An alternative approach here is the following:

add only 1 new field called extraPeriod

let AllocationRequest.period = be the entire new period whether it's a new cover or an edit

In such a way that if it's an edit:

remainingPeriod = allocationRequest.period - allocationRequest.extraPeriod

contracts/mocks/Cover/CoverMockStakingPool.sol
```diff
@@ -75,7 +75,13 @@ contract CoverMockStakingPool is IStakingPool {
       ? calculateFixedPricePremium(amount, request.period, mockPrices[request.productId])
       : calculatePremium(mockPrices[request.productId], amount, request.period);
 
-    return (premium, allocationId);
+    if (amount == 0) {
+      return (0, 0);
+    }
+    uint extraPremium = premium * Math.max(
+      (amount - previousAllocationAmountInNXMRepriced), 0) / amount;
```
danoctavian, line 82: This computation won't work correctly if amount <  previousAllocationAmountInNXMRepriced since we're using uint, need to compare the 2 values

contracts/modules/cover/Cover.sol
```diff
@@ -316,28 +359,78 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+    uint oldSegmentAmountInNXMRepriced = getNXMForAssetAmount(
+      params.previousSegmentAmount, params.nxmPriceInCoverAsset
+    );
 
     for (uint i = 0; i < poolAllocationRequests.length; i++) {
       // if there is a previous segment and this index is present on it
       if (vars.previousPoolAllocationsLength > i) {
-        PoolAllocation memory previousPoolAllocation =
-          coverSegmentAllocations[allocationRequest.coverId][segmentId - 1][i];
+        PoolAllocation memory previousPoolAllocation = vars.previousPoolAllocations[i];
```
danoctavian, line 433: renamed array to previousPoolAllocations

I prefer to still use this local variable here `previousPoolAllocation` to not refer `vars.previousPoolAllocations[i]` in every place, not a strong preference

danoctavian:
Contract size notes:

 |  StakingPool                            ·                   23.881  ·       -14.977  │
 
 
  |  Cover                                  ·                   20.718  ·       -12.820  │
 
 
These 2 are the largest contracts - hence i strongly argue at this point to keep createStakingPool out of Cover.sol

Also StakingPool is also very close to max size after introducing the new pricing logic.

Still deployable.

contracts/modules/cover/Cover.sol
```diff
@@ -153,19 +165,32 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
 
-      // new cover
+      // Handle creating a new cover
       coverId = coverNFT.mint(params.owner);
       _coverData[coverId] = CoverData(params.productId, params.coverAsset, 0 /* amountPaidOut */);
 
+      allocationRequest.coverId = coverId;
+      // for a new cover, the first segment the remaining period and new period are the requested period
+      allocationRequest.remainingPeriod = uint32(allocationRequest.period);
+      allocationRequest.newPeriod = uint32(allocationRequest.period);
```
roxdanila, line 184: these should't be used at all for new covers, do we really need to set them?

danoctavian: Good question - The way i wrote requestAllocation is that it handles both create and edit with the same code paths by using

allocationRequest.remainingPeriod  as the period used to compute premium for the New Cover case.

Seeing that this may cause confusion - i'm thinking to basically have an if/else in request allocation and have 2 totally separate code paths for computing price for new cover and edit cover and not use these fields in both cases

contracts/modules/staking/StakingPool.sol
```diff
@@ -690,27 +711,64 @@ contract StakingPool is IStakingPool, Multicall {
...
+    uint totalCapacity,
+    AllocationRequest calldata request
+  ) internal returns (uint totalPremium) {
 
-      // store
-      rewardPerSecondCut[prevExpirationBucket] -= prevRewardsPerSecond;
-      rewardPerSecond -= prevRewardsPerSecond.toUint96();
+    if (amount > previousAllocationAmountInNXMRepriced) {
+
+      // this handles both new allocations (allocationId == 0) and edits that increase the amount
```
danoctavian, line 732: Screenshot 2023-09-15 at 9 46 17 AM
Merged the computation logic here again because of contract size

danoctavian:
Have some TODOs left in the integration test related to an off by 1 wei error that appears only on certain test runs because of timing

the issues comes from the fact that i compute the premium with a  slightly different logic than what the on-chain contract does, but haven't detected exactly where the difference appears

contracts/modules/staking/StakingPool.sol
```diff
@@ -682,6 +703,7 @@ contract StakingPool is IStakingPool, Multicall {
       }
 
       uint expirationBucket = Math.divCeil(block.timestamp + request.period, BUCKET_DURATION);
+
       uint rewardStreamPeriod = expirationBucket * BUCKET_DURATION - block.timestamp;
```
danoctavian, line 707: @shark0der do you see any concerns here with maintaining rewards streaming  the exact same, meaning, in the case of an edit the rewards are streamed based on the extra premium and just added to the rewards stream respecting the new end time of the edited segment as such:

https://github.com/NexusMutual/smart-contracts/blob/ad60a3a9815373b6b6c002ab7255ea7b2dd99751/contracts/modules/staking/StakingPool.sol#L705

```uint _rewardPerSecond = (premium * request.rewardRatio / REWARDS_DENOMINATOR) / rewardStreamPeriod;```

I heard there might be concerns regard APY in some way?

File: contracts/modules/cover/Cover.sol [hunk 12]
```diff
@@ -190,44 +212,63 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+      if (params.period + remainingPeriod < MIN_COVER_PERIOD) {
+        revert CoverPeriodTooShort();
+      }
+
+      if (params.period + remainingPeriod > MAX_COVER_PERIOD) {
+        revert CoverPeriodTooLong();
+      }
+
       allocationRequest.previousStart = lastSegment.start;
+
```
roxdanila, line 226: remove extra line

contracts/modules/cover/Cover.sol
```diff
@@ -190,44 +212,63 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
         revert ExpiredCoversCannotBeEdited();
       }
 
+      uint remainingPeriod = lastSegment.start + lastSegment.period - block.timestamp + 1;
```
roxdanila, line 215: shouldn't this be `-1` instead of `+1`?

contracts/modules/cover/Cover.sol
```diff
@@ -316,28 +357,79 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
-      : 0;
+    for (uint i = 0; i < poolAllocationRequests.length; i++) {
+
+      // converting asset amount to nxm and rounding up to the nearest NXM_PER_ALLOCATION_UNIT
+      uint coverAmountInNXM = getNXMForAssetAmount(
+        poolAllocationRequests[i].coverAmountInAsset,
+        nxmPriceInCoverAsset
+      );
+
+    (uint premiumInNXM, uint allocationId) = stakingProducts().stakingPool(poolAllocationRequests[i].poolId).requestAllocation(
```
roxdanila, line 379: careful with indentation

contracts/modules/cover/Cover.sol
```diff
@@ -316,28 +357,79 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+    AllocationRequest memory allocationRequest,
+    PoolAllocationRequest[] memory poolAllocationRequests,
+    AllocationParams memory params
+  ) internal returns (
+    uint totalCoverAmountInCoverAsset,
+    uint totalAmountDueInNXM
+  ) {
+
+    RequestAllocationVariables memory vars; // = RequestAllocationVariables(0, 0, 0, 0);
+
```
roxdanila, line 414: no need for all these blank lines imo, makes the code harder to read

roxdanila, line 414: remove these empty lines, makes it harder to read imo

contracts/modules/cover/Cover.sol
```diff
@@ -316,28 +357,79 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+    vars.previousPoolAllocationsLength = coverSegmentAllocations[allocationRequest.coverId][params.segmentId - 1].length;
+
+    vars.previousPoolAllocations = new PoolAllocation[](vars.previousPoolAllocationsLength);
+
+    for (uint i = 0; i < vars.previousPoolAllocationsLength; i++) {
+      vars.previousPoolAllocations[i] = coverSegmentAllocations[allocationRequest.coverId][params.segmentId - 1][i];
+
```
roxdanila, line 421: remove this empty line

contracts/modules/cover/Cover.sol
```diff
@@ -316,28 +357,79 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+      vars.previousTotalCoverAmountInNXM += vars.previousPoolAllocations[i].coverAmountInNXM;
+    }
+
+    uint oldSegmentAmountInNXMRepriced = getNXMForAssetAmount(
```
roxdanila, line 425: to be consistent, rename `oldSegmentAmountInNXMRepriced` -> `previousCoverAmountInNXMRepriced`

contracts/modules/cover/Cover.sol
```diff
@@ -316,28 +357,79 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
     }
   }
 
-  function requestAllocation(
+  function requestAllocationsForNewCover(
```
roxdanila, line 360: pls keep names consistent between the 2 functions - something along this lines:
`requestAllocationsNewCover`
`requestAllocationsEditCover`

contracts/modules/cover/Cover.sol
```diff
@@ -316,28 +357,79 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+
+      vars.previousTotalCoverAmountInNXM += vars.previousPoolAllocations[i].coverAmountInNXM;
+    }
+
+    uint oldSegmentAmountInNXMRepriced = getNXMForAssetAmount(
+      params.previousSegmentAmount, params.nxmPriceInCoverAsset
+    );
 
     for (uint i = 0; i < poolAllocationRequests.length; i++) {
       // if there is a previous segment and this index is present on it
```
roxdanila, line 430: there is always a previous segment in this case, the comment is a bit confusing. it should say 
`// if the allocation request id is present in the previous segment`

contracts/modules/cover/Cover.sol
```diff
@@ -346,43 +438,45 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
 
         // check if this request should be skipped, keeping the previous allocation
         if (poolAllocationRequests[i].skip) {
-          coverSegmentAllocations[allocationRequest.coverId][segmentId].push(previousPoolAllocation);
-          totalCoverAmountInNXM += previousPoolAllocation.coverAmountInNXM;
+          coverSegmentAllocations[allocationRequest.coverId][params.segmentId].push(previousPoolAllocation);
+          vars.totalCoverAmountInNXM += previousPoolAllocation.coverAmountInNXM;
           continue;
```
roxdanila, line 443: Two important observations regarding this bit of logic:
1. You always count towards `totalCoverAmountInNXM` the cover amount from this allocation, although the pool is skipped. This works fine when the user only performs an increase of the amount. However, if the user (also) increases the period, skipping a pool has a different semantic, and it's rather equivalent to fully abandoning the pool (give up the premium already paid on it and don't count that allocation towards the total cover amount), as that pool probably doesn't have the required capacity for the newly desired period. What should the behaviour be in this case?
2. As you pointed out @danoctavian, for skipped pools we never reprice the allocations, as we do for non-skipped pools. Is this a fair behaviour?

contracts/modules/cover/Cover.sol
```diff
@@ -346,43 +438,45 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
         allocationRequest.allocationId = previousPoolAllocation.allocationId;
+
+        uint poolAllocationRatio =
+          previousPoolAllocation.coverAmountInNXM
+          * COVER_ALLOCATION_RATIO_DENOMINATOR
+          / vars.previousTotalCoverAmountInNXM;
+
+        vars.previousAllocationAmountInNXMRepriced =
+          poolAllocationRatio * oldSegmentAmountInNXMRepriced
+          / COVER_ALLOCATION_RATIO_DENOMINATOR;
```
roxdanila, line 456: this is a bit hard to read - there might be a way to simplify it and some more consistent naming will also help

contracts/modules/cover/Cover.sol
```diff
@@ -346,43 +438,45 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+
       } else {
         // request new allocation id
         allocationRequest.allocationId = 0;
 
-        // zero out previous premium or refund
-        vars.previousPremiumInNXM = 0;
-        vars.refund = 0;
+        vars.previousAllocationAmountInNXMRepriced = 0;
```
roxdanila, line 462: more broadly on naming:
- all allocations are in NXM always, so probably worth dropping "InNXM"
- all allocations are either pool allocations, or total allocation - so worth dropping "pool" everywhere or keep it everywhere - it's quite inconsistent atm

contracts/modules/cover/Cover.sol
```diff
@@ -346,43 +438,45 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
       );
 
       (uint premiumInNXM, uint allocationId) = stakingProducts().stakingPool(poolAllocationRequests[i].poolId).requestAllocation(
         coverAmountInNXM,
-        vars.previousPremiumInNXM,
+        vars.previousAllocationAmountInNXMRepriced,
         allocationRequest
       );
 
       // omit deallocated pools from the segment
```
roxdanila, line 477: do we ever actually deallocate in the current design?

contracts/modules/cover/Cover.sol
```diff
@@ -392,15 +486,22 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
...
+      vars.totalCoverAmountInNXM += coverAmountInNXM;
     }
 
-    totalCoverAmountInCoverAsset = totalCoverAmountInNXM * nxmPriceInCoverAsset / ONE_NXM;
+    totalCoverAmountInCoverAsset = vars.totalCoverAmountInNXM * params.nxmPriceInCoverAsset / ONE_NXM;
 
     return (totalCoverAmountInCoverAsset, totalAmountDueInNXM);
   }
 
+  function getNXMForAssetAmount(uint amountInCoverAsset, uint nxmPriceInCoverAsset) pure internal returns (uint) {
```
roxdanila, line 498: I'd personally inline this, but it's subjective

contracts/modules/cover/Cover.sol
roxdanila: The spec says something else for skipped pools:
```
- For Pools that we “skip”:
    - deallocate the capacity, but rewards keep streaming to the stakers as per the original allocation
    - set `coverAmountInNXM` to 0```

danoctavian: At this point given that one can simply leave a pool in by submitting the same amount for that allocation and set request.period == 0 (no extra period), the skip logic doesn't seem necessary.

With that existing implementation of the skip boolean one had to check that only the amount is increased and not the period to apply it.

PR #950, opened 2023-10-13
Update deploy script + deployments build script

Gideonnn:
The deployments `build.js` script now also outputs abi json files to `dist/data/abis/`. The `addresses.json` file is also moved to the `dist/data` folder (from the `dist` folder). Now the `@nexusmutual/sdk` package can copy them over and re-expose them. This is convenient for loading the abis with libs who require json files instead of javascript objects.

But the actual real reason this was changed is because this enables the `frontend-next` to load freshly generated abis via some `npm link` magic. Check [Notion](https://www.notion.so/nxmcommunity/Local-development-95f84f09cbfb4b90bcdcba52e1d2fa90?pvs=4) out for more info.

Note: this includes the Ramm contract.

### DIFF
--- a/scripts/deploy/deploy.js
+++ b/scripts/deploy/deploy.js
@@ -9,7 +9,7 @@ const verifier = require('./verifier')();
 const { setEtherBalance } = require('../../test/utils').evm;
 
 const { AddressZero, MaxUint256 } = ethers.constants;
-const { parseEther } = ethers.utils;
+const { parseEther, parseUnits } = ethers.utils;
 const { ABI_DIR, ADDRESSES_FILE, INITIAL_MEMBERS = '' } = process.env;
 
 if (!ABI_DIR || !ADDRESSES_FILE) {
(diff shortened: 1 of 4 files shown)


deployments/build.js
```diff
@@ -29,6 +29,7 @@ const contractList = [
   'PriceFeedOracle',
   'ProductsV1',
   'ProposalCategory',
+  'Ramm',
```
Gideonnn, line 32: Not sure if I can include this already

scripts/deploy/deploy.js
```diff
@@ -586,13 +598,13 @@ async function main() {
...
-      config.CONTRACTS_ADDRESSES[alias] = address;
+    if (!addressesMap[alias] || isProxy) {
+      addressesMap[alias] = address;
     }
   }
 
-  console.log(`Updating config file ${configFile}`);
-  fs.writeFileSync(configFile, JSON.stringify(config, null, 2), 'utf8');
+  console.log(`Updating addresses.json ${addressesFile}`);
+  fs.writeFileSync(addressesFile, JSON.stringify(addressesMap, null, 2), 'utf8');
```
Gideonnn, line 607: After writing the `README.md` this seems very strange to me. The `build.js` script now generates ABIs based on Hardhat, but `addresses.json` is simply copied. Then `deploy.js` will actually output new files, which should actually be used by `build.js` to create a package output.

The output for abis and config (now addresses file location) were used by the `frontend-react` repo, but this now works differently.

@shark0der Let's chat about what will be the correct way forward, because now I think is needlessly complex. In my opinion running build and then deploy would make most sense. Now you need to run build -> deploy -> build.

PR #985, opened 2023-10-31
Check membership when minting

shark0der:
## Context

Previously the mint function from TokenController did not check if the destination address is a member or not.

This PR introduces an internal function `_mint` that performs this checks. All other TC functions should use it instead of directly calling the token contract.

## Test plan

Added a new test to check for revert when the destination address is not a member.

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/token/TokenController.sol
+++ b/contracts/modules/token/TokenController.sol
@@ -151,11 +151,21 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
   }
 
   /**
-  * @dev Mints new token for an address
-  * @param _member address to reward the minted tokens
+  * @dev Mints new tokens for an address and checks if the address is a member
+  * @param _member address to send the minted tokens to
   * @param _amount number of tokens to mint
   */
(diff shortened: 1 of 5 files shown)


contracts/modules/token/TokenController.sol
```diff
@@ -151,11 +151,21 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
+  * @dev Mints new tokens for an address and checks if the address is a member
+  * @param _member address to send the minted tokens to
   * @param _amount number of tokens to mint
   */
   function mint(address _member, uint _amount) public override onlyInternal {
+    _mint(_member, _amount);
+  }
+
+  /**
+  * @dev Internal function to mint new tokens for an address and checks if the address is a member
```
rackstar, line 163: is it worth adding a comment here that internal functions should only use `_mint` and never `token.mint` directly?

so that future devs are aware and don't make mistakes

roxdanila (approved):

PR #998, opened 2023-11-07
Get MCR out of Amsterdam

shark0der:
## Context

The current MCR value is driven by the MCR floor parameter which is quite high. Due to the MCR smoothing it won't get down to the gear value immediatelly. This prevents the users from using the `Ramm.swap()` right after the upgrade.

This PR includes a new function `MCR.teleportMCR()` which sets the mcr at 10k. It's public, only callable once and has a deadline set at 1st of December. Additionally there's a safeguard to make sure it's not called before the system is actually upgraded to this contract.

## Test plan

Unit tests added to make sure the above requirements hold.

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/capital/MCR.sol
+++ b/contracts/modules/capital/MCR.sol
@@ -39,8 +39,11 @@ contract MCR is IMCR, MasterAwareV2 {
   uint public constant MAX_MCR_ADJUSTMENT = 100;
   uint public constant BASIS_PRECISION = 10000;
 
-  constructor (address masterAddress) {
+  uint public immutable MCR_UPDATE_DEADLINE;
+
+  constructor (address masterAddress, uint mcrUpdateDeadline) {
     changeMasterAddress(masterAddress);
+    MCR_UPDATE_DEADLINE = mcrUpdateDeadline;
 
     if (masterAddress != address(0)) {
       previousMCR = IMCR(master.getLatestAddress("MC"));
@@ -91,14 +94,15 @@ contract MCR is IMCR, MasterAwareV2 {
   }
 
   /**
-   * @dev We need to move the mcr way below the current value otherwise swaps won't work
+   * @dev We need to move the mcr way below the current value otherwise swaps
+   *      won't work for a while until mcr moves down by itself
    * @dev Remove this code after the tokenomics upgrade.
    */
   function teleportMCR() external {
 
     require(address(previousMCR) == address(0), "MCR: not yet initialized");
     require(mcr > 10_000 ether, "MCR: already updated");
-    require(block.timestamp < 1701388800, "MCR: Deadline has passed"); // 1701388800 = December 1st, 2023
+    require(block.timestamp < MCR_UPDATE_DEADLINE, "MCR: Deadline has passed");
 
     mcr = 10_000 ether;
     desiredMCR = 10_000 ether;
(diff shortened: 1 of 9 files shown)


contracts/modules/capital/MCR.sol
```diff
@@ -90,6 +90,21 @@ contract MCR is IMCR, MasterAwareV2 {
     previousMCR = IMCR(address(0));
   }
 
+  /**
+   * @dev We need to move the mcr way below the current value otherwise swaps won't work
+   * @dev Remove this code after the tokenomics upgrade.
+   */
+  function teleportMCR() external {
```
rackstar, line 97: anyone can call this?

test/unit/MCR/teleportMCR.js
```diff
@@ -0,0 +1,99 @@
...
+    // deploy mcr with fake master
+    const mcr = await MCR.deploy(disposableMCR.address);
+
+    await expect(mcr.teleportMCR()).to.be.revertedWith('MCR: not yet initialized');
+
+    // trigger initialize and switch master address
+    await disposableMCR.initializeNextMcr(mcr.address, master.address);
+
+    // should not revert
+    await mcr.teleportMCR();
```
rackstar, line 97:
```solidity
await expect(mcr.teleportMCR()).to.not.be.reverted;
```

test/unit/MCR/teleportMCR.js
```diff
@@ -0,0 +1,99 @@
...
+
+describe('teleportMCR', function () {
+  it('teleportMCR updates values accordingly', async function () {
+    const fixture = await loadFixture(setup);
+    const { master } = fixture;
+
+    const mcr = await initMCR({ ...DEFAULT_MCR_PARAMS, master });
+
+    const { timestamp: currentTimestamp } = await ethers.provider.getBlock('latest');
+    const nextUpdateTime = currentTimestamp + 1;
```
rackstar, line 30: we need to set this deterministically to a timestamp < 1701388800 so it won't fail in the future

contracts/modules/capital/MCR.sol
```diff
@@ -90,6 +90,21 @@ contract MCR is IMCR, MasterAwareV2 {
...
+
+    require(address(previousMCR) == address(0), "MCR: not yet initialized");
```
roxdanila, line 99: do you need to call this function before we actually upgrade the contract?

shark0der: After the actual upgrade. Alternatively I could add this code to the initialize function but I wanted to make it easier to remove the code afterwards, so this was cleaner from that perspective. If we forget to - we have the deadline check.

shark0der: Will have to make the deadline immutable since it fails the tests because it introduces a time dependency.

test/unit/MCR/teleportMCR.js
shark0der: will make immutable and pass a constructor argument :+1:

test/unit/MCR/teleportMCR.js
shark0der: True, but the stack trace and error message when it actually fails is totally useless.

Issue #1053, opened 2024-01-10
Cover Re Investment


roxdanila:

roxdanila:
@MilGard91 please add:
- list of contracts to be deployed
- list of governance proposals to be executed
- list of commands to run afterwards to transfer funds

PR #1058, opened 2024-01-26
feat: Cover Re

MilGard91:
## Context

Issue: #1053 

- Add new contract `SafeTracker` to track  the balance of the safe and invested USDC. It will be added as asset to the `Pool` contract for the tracking of the balance
- Add new methods to `SwapOperator` to request transfer and to execute the transfer to `Safe`

## Test plan

- Unit tests added for the new contract and it's methods
- Unit tests for new methods on the `SwapOperator`
- Fork tests for the flow of the investment 

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/modules/capital/SafeTracker.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    // usdc debt (borrowed usdc)
+    uint debtUsdcAmount = debtUsdc.balanceOf(safe);
+    uint debtUsdcValueInEth = priceFeedOracle.getEthForAsset(address(usdc), debtUsdcAmount);
+
+    return ethAmount + usdcValueInEth + daiValueInEth - debtUsdcValueInEth;
+  }
+
+  function _transfer(address from, address to, uint256 amount) internal returns (bool) {
+    if (amount == 0 || msg.sender == internalContracts[uint(ID.P1)] || msg.sender == pool().swapOperator()) {
```
roxdanila, line 136: remove `swapOperator`

contracts/modules/capital/SafeTracker.sol
```diff
@@ -0,0 +1,152 @@
...
+      emit Transfer(from, to, amount);
+      return true;
+    }
+    revert();
```
roxdanila, line 140: Add message: "Exceeds balance"

contracts/modules/capital/SafeTracker.sol
```diff
@@ -0,0 +1,152 @@
...
+    // eth in the safe and aweth balance, aweth is 1:1 to eth
+    uint ethAmount = address(safe).balance + aweth.balanceOf(safe);
+
+    IPriceFeedOracle priceFeedOracle = pool().priceFeedOracle();
+
+    // dai in the safe
+    uint daiAmount = dai.balanceOf(safe);
+    uint daiValueInEth = priceFeedOracle.getEthForAsset(address(dai), daiAmount);
+
+    // usdc actually in the safe
```
roxdanila, line 124: update comment to include invested usdc too - it's currently inaccurate

contracts/external/aave/AaveDataTypes.sol
```diff
@@ -0,0 +1,265 @@
+// SPDX-License-Identifier: BUSL-1.1
```
shark0der, line 1: are `contracts/external/aave/*` only used for fork tests? if yes, i'd avoid creating interfaces and would only declare the abi directly in js. we have over a hundred contracts and if we keep adding the compilation time will grow

contracts/modules/capital/SwapOperator.sol
```diff
@@ -25,6 +26,12 @@ import "../../external/enzyme/IEnzymePolicyManager.sol";
 contract SwapOperator {
```
shark0der, line 26: SwapOperator should implement the `ISwapOperator` interface. Seems like the interface still has the `transferRequestedETH` function!

contracts/modules/capital/SwapOperator.sol
```diff
@@ -48,16 +55,28 @@ contract SwapOperator {
   uint public constant MIN_TIME_BETWEEN_ORDERS = 900; // 15 minutes
   uint public constant maxFee = 0.3 ether;
 
+  // Safe variables
+  address public safe;
+  Request public transferRequest;
+  mapping(address => bool) public allowedAssets;
```
shark0der, line 61: needs a comment explaining where are these assets allowed, allowed to what?

shark0der: Or maybe even rename to include "investment". `allowedInvestmentAssets`? Don't necessarily stick to this suggestion, discuss please.

contracts/modules/capital/SwapOperator.sol
```diff
@@ -48,16 +55,28 @@ contract SwapOperator {
...
+
+
   // Events
   event OrderPlaced(GPv2Order.Data order);
   event OrderClosed(GPv2Order.Data order, uint filledAmount);
   event Swapped(address indexed fromAsset, address indexed toAsset, uint amountIn, uint amountOut);
+  event TransferredToSafe(address asset, uint amount);
```
shark0der, line 68: Consider moving the events to the interface.

contracts/modules/capital/SwapOperator.sol
```diff
@@ -504,6 +536,61 @@ contract SwapOperator {
...
+    transferAssetTo(asset, safe, amount);
+    emit TransferredToSafe(asset, amount);
+  }
+
+  /**
+   * @dev Validate that a requested asset is not sent if amount surpasses the limit
+   * @param asset The address of the asset
+   * @param amount The amount of the asset
+   */
+  function checkTransferAvailability(address asset, uint amount) internal view {
```
shark0der, line 576: If this is the fix for the reported "Lack of Outflow Limit" audit issue see the comments on notion and revisit the design decision.

contracts/modules/capital/PriceFeedOracle.sol
```diff
@@ -10,16 +10,22 @@ contract PriceFeedOracle is IPriceFeedOracle {
...
+    address _safeTracker
   ) {
     require(
       _assetAddresses.length == _assetAggregators.length && _assetAggregators.length == _assetDecimals.length,
       "PriceFeedOracle: different args length"
     );
+    require(_safeTracker != address(0), "PriceFeedOracle: safeTracker cannot be zero address");
+
+    safeTracker = _safeTracker;
+    assets[_safeTracker] = OracleAsset(Aggregator(_safeTracker), 18);
```
shark0der, line 28: Given this, we might want to actually implement a dummy `latestAnswer()` function that would return `1e18`. It should be never called, but if it's accidentally called it would return a `1:1` rate.

contracts/modules/capital/SafeTracker.sol
```diff
@@ -0,0 +1,152 @@
...
+import "../../interfaces/IPriceFeedOracle.sol";
+import "../../interfaces/ISafeTracker.sol";
+
+contract SafeTracker is ISafeTracker, MasterAwareV2 {
+
+  uint public coverReInvestmentUSDC;
+
+  string public symbol = "NXMIS";
+  string public name = "NXMIS";
+  uint8 public decimals = 18;
```
shark0der, line 17: This contract is a proxy. Initialising like this they'll be stored in storage and the the proxy will return empty values. Make them constants.

contracts/modules/capital/SafeTracker.sol
```diff
@@ -0,0 +1,152 @@
...
+    address _aweth,
+    address _debtUsdc
+  ) {
+    require(_master != address(0), "SafeTracker: master cannot be zero address");
+    require(
+      _usdc != address(0) && _dai != address(0) && _aweth != address(0) && _debtUsdc != address(0),
+      "SafeTracker: tokens address cannot be zero address"
+    );
+
+    master = INXMMaster(_master);
```
shark0der, line 44: `master` is a storage variable and the contract is a proxy. This will initialize the storage of the implementation and not of the proxy. Remove this parameter from the constructor. See my comment on the issue on notion.

contracts/modules/capital/SafeTracker.sol
```diff
@@ -0,0 +1,152 @@
...
+  function transferFrom(address from, address to, uint256 amount) external returns(bool) {
+    return _transfer(from, to, amount);
+  }
+
+  function allowance(address, address) external pure returns (uint256) {
+    return 0;
+  }
+
+  function approve(address, uint256) external pure override returns (bool) {
+    revert();
```
shark0der, line 106: Probably would make sense to not revert similarly to the zero transfer and to make it behave more like a normal ERC20 we'd also need an event which would make it a normal (not pure and not view). No strong preference here tbh as it has no impact on functionality.

contracts/external/aave/AaveDataTypes.sol
MilGard91: Yes they are, I'll use their abis instead. There we other external interfaces that's why they were added this way...

contracts/modules/capital/SwapOperator.sol
MilGard91: Maybe something more safe related, since it is what we are sending to the safe (`allowedSafeTransferAssets`), if we in future have different investment flows we will need to change it again.

contracts/external/aave/AaveDataTypes.sol
roxdanila: I think the other interfaces are actually used in the code as well, not just in tests

contracts/modules/capital/SwapOperator.sol
roxdanila: I'd go for `allowedAssetsForSafeTransfer` as it's more explicit, but up to you

contracts/modules/capital/SafeTracker.sol
```diff
@@ -99,8 +99,12 @@ contract SafeTracker is ISafeTracker, MasterAwareV2 {
     return 0;
   }
 
-  function approve(address, uint256) external pure override returns (bool) {
-    revert();
+  function approve(address spender, uint256 value) external override returns (bool) {
+    if (value == 0) {
```
shark0der, line 103: Approve should work with any value, even if it exceeds your balance.

PR #1090, opened 2024-03-28
feat: Add CoverBroker

### LINKED ISSUE #1091: CoverBroker: implementation
## Overview

* the contract is kyc-ed and whitelisted as a member
* interacts with the Nexus Cover contract on behalf of the user
* mints the NFT in the end user's wallet

## Functions
* buyCover: calls Cover.buyCover and pass end user's address as BuyCoverParams.owner, so the NFT gets minted directly in their wallet
* switchMembership: in case we want in the future to redeploy the contract. This can be called by an owner address only
* transferFunds (?) - fish any tokens that accidentally get sent to the contract - can only be called by the owner address above

## Further notes:
* buying using ETH, DAI or any supported pool assets
* user -> external contract -> buy cover (will have extra buffer payment - so it does not revert)
* need to make sure any extra payment is sent back to user (any money before buyCover should stay in contract)

rackstar:
## Context

Closes #1091 

* [x] add ICoverBroker
* [x] add CoverBroker
* [x] add unit tests

## Test plan

* [x] add unit tests

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,97 @@
...
+    BuyCoverParams memory params,
+    PoolAllocationRequest[] calldata poolAllocationRequests
+  ) external payable nonReentrant returns (uint coverId) {
+    uint ethBalanceBefore = address(this).balance;
+    // set the owner to the caller
+    params.owner = msg.sender;
+    // call cover.buyCover with msg.value and params
+    coverId = cover.buyCover{value: msg.value}(params, poolAllocationRequests);
+    // send back any ETH refund to user
+    uint ethRefund = address(this).balance - ethBalanceBefore;
```
rackstar, line 60: i'm forwarding whatever the value the caller sends to this method to buyCover.

if the cover turned out cheaper than `msg.value`, I'm assuming the Cover contract will send the "change" back to CoverBroker.

is this assumption correct?

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,101 @@
...
+    coverId = cover.buyCover{value: msg.value}(params, poolAllocationRequests);
+
+    if (address(this).balance > ethBalanceBefore) {
+      // send back any ETH refund to user
+      unchecked {
+        uint ethRefund = address(this).balance - ethBalanceBefore;
+        (bool sent, ) = payable(msg.sender).call{value: ethRefund}("");
+        if (!sent) {
+          revert TransferFailed(msg.sender, ethRefund, ETH);
+        }
```
rackstar, line 69: i'm forwarding whatever the value the caller sends to this method to buyCover.

if the cover turned out cheaper than msg.value, I'm assuming the Cover contract will send the "change" back to CoverBroker.

is this assumption correct?

MilGard91: Yes, since the Cover contract returns the difference to the msg.sender the extra amount will be sent to the CoverBroker contract

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,90 @@
...
+
+import "@openzeppelin/contracts-v4/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts-v4/security/ReentrancyGuard.sol";
+
+import "../../interfaces/ICover.sol";
+import "../../interfaces/ICoverBroker.sol";
+import "../../interfaces/IMemberRoles.sol";
+
+/// @dev Allows cover distribution by buying cover in behalf of the caller
+contract CoverBroker is ICoverBroker, ReentrancyGuard {
```
roxdanila, line 13: alternative name suggestion: CoverWrapper

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,90 @@
...
+
+  /// @dev buys cover in behalf of the caller
+  function buyCover(
+    BuyCoverParams memory params,
+    PoolAllocationRequest[] calldata poolAllocationRequests
+  ) external payable nonReentrant returns (uint coverId) {
+    uint ethBalanceBefore = address(this).balance;
+
+    // set the cover owner to the caller
+    params.owner = msg.sender;
```
shark0der, line 44: Why override this? Maybe they indeed want to send it somewhere else. You can default to `msg.sender` if the value is the zero address but that's totally optional as it will revert if the zero address is sent.

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,90 @@
...
+    owner = _owner;
+    cover = ICover(_cover);
+    memberRoles = IMemberRoles(_memberRoles);
+  }
+
+  /// @dev buys cover in behalf of the caller
+  function buyCover(
+    BuyCoverParams memory params,
+    PoolAllocationRequest[] calldata poolAllocationRequests
+  ) external payable nonReentrant returns (uint coverId) {
```
shark0der, line 40: `nonReentrant` seems redundant

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,90 @@
...
+    uint ethBalanceBefore = address(this).balance;
```
shark0der, line 41: Please leave an empty line to delimitate the function header and the body

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,90 @@
...
+
+    // set the cover owner to the caller
+    params.owner = msg.sender;
+
+    // call cover.buyCover with msg.value and params
+    coverId = cover.buyCover{value: msg.value}(params, poolAllocationRequests);
+
+    if (address(this).balance > ethBalanceBefore) {
+      // send back any ETH refund to user
+      unchecked {
```
shark0der, line 51: Why did you wrap this in an unchecked block?

contracts/external/cover/CoverBroker.sol
shark0der: The assumption is correct but we need a `receive` function to allow ETH transfers back to this contract.

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,90 @@
...
+  ) external payable nonReentrant returns (uint coverId) {
+    uint ethBalanceBefore = address(this).balance;
+
+    // set the cover owner to the caller
+    params.owner = msg.sender;
+
+    // call cover.buyCover with msg.value and params
+    coverId = cover.buyCover{value: msg.value}(params, poolAllocationRequests);
+
+    if (address(this).balance > ethBalanceBefore) {
```
shark0der, line 49: Seems like only ETH cover buys are possible. AFAIR this is going to be the case for the first use-case but we should implement non-ETH cover buys as well.

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,90 @@
...
+
+  /// @dev transfers available funds to owner
+  function transferFunds(address assetAddress) external onlyOwner {
+    if (assetAddress == ETH) {
+      uint ethBalance = address(this).balance;
+      if (ethBalance == 0) {
+        revert ZeroBalance(ETH);
+      }
+
+      (bool sent, ) = payable(owner).call{value: ethBalance}("");
```
shark0der, line 74: nit: `msg.sender` will be cheaper than re-reading `owner` from storage, even if it's hot (2 gas vs 100 gas).

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,90 @@
...
+
+import "../../interfaces/ICover.sol";
+import "../../interfaces/ICoverBroker.sol";
+import "../../interfaces/IMemberRoles.sol";
+
+/// @dev Allows cover distribution by buying cover in behalf of the caller
+contract CoverBroker is ICoverBroker, ReentrancyGuard {
+
+  // Immutables
+  address owner;
```
shark0der, line 16: Use `Ownable` from OZ instead as it allows ownership change (for example if the wallet is compromised)

contracts/external/cover/CoverBroker.sol
rackstar: dropped

contracts/external/cover/CoverBroker.sol
rackstar: the if condition guarantees that it will not underflow

contracts/external/cover/CoverBroker.sol
rackstar: added Ownable.sol inheritance

contracts/external/cover/CoverBroker.sol
rackstar: add ERC20 payment support

contracts/external/cover/CoverBroker.sol
shark0der: The benefit is marginal, let's just remove it in favor of readability.

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,88 @@
...
+  ) external payable returns (uint coverId) {
+
+    uint ethBalanceBefore = address(this).balance;
+
+    if (params.paymentAsset == ETH_ASSET_ID) {
+      // ETH payment
+      coverId = cover.buyCover{value: msg.value}(params, poolAllocationRequests);
+    } else {
+      // ERC20 payment
+      coverId = cover.buyCover(params, poolAllocationRequests);
```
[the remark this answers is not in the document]
shark0der: We also need to pull the funds from the cover buyer.

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    // send any ETH refund back to msg.sender
+    if (address(this).balance > ethBalanceBefore) {
```
shark0der, line 45: We need to refund based on the cover asset

contracts/external/cover/CoverBroker.sol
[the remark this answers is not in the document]
shark0der: And I'd also revert when the payment asset is not ETH but `msg.value > 0`

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,88 @@
...
+        if (!sent) {
+          revert TransferFailed(msg.sender, ethRefund, ETH);
+        }
+      }
+    }
+  }
+
+  /// @dev switches the membership to the given address
+  function switchMembership(address newAddress) external onlyOwner {
+    memberRoles.switchMembership(newAddress);
```
shark0der, line 58: We likely need an NXM approval to the MR contract

shark0der:
Side note: will the payment with NXM work ok? In theory we have this contract so people that aren't KYC-ed can buy so it shouldn't be used to pay with NXM, but technically it is possible so we should make sure it will work.

contracts/external/cover/CoverBroker.sol
rackstar: removed

rackstar:
> Side note: will the payment with NXM work ok? In theory we have this contract so people that aren't KYC-ed can buy so it shouldn't be used to pay with NXM, but technically it is possible so we should make sure it will work.

the pool needs to support NXM as an asset. we retrieve the address of the paymentAsset id via the pool:

```solidity
    address paymentAsset = pool.getAsset(params.paymentAsset).assetAddress;
    IERC20 erc20 = IERC20(paymentAsset);
```

currently pool only supports these assets: [ETH, DAI, stETH NXMTY, rETH](https://etherscan.io/address/0xcafeaBED7e0653aFe9674A3ad862b78DB3F36e60#readContract#F7)

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,190 @@
...
+    }
+  }
+  
+  /// @notice Handles NXM payments for buying cover.
+  /// @dev Transfers NXM from the caller to the contract, then buys cover on behalf of the caller.
+  /// Calculates NXM refunds if any and sends back to msg.sender.
+  /// @param params The parameters required to buy cover.
+  /// @param poolAllocationRequests The allocation requests for the pool's liquidity.
+  /// @return coverId The ID of the purchased cover.
+  function _handleNxmPayment(
```
rackstar, line 142: _handleNxmPayment and _handleErc20Payment logic is very similar. if not for SafeERC20 we could have probably had the same code to handle both flows

roxdanila:
After reviewing this, I'm more strongly of the opinion we should remove the option to pay with NXM.
If someone owns NXM, they can purchase through our core contracts - it'd be cheaper. This would simplify the code as well.

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,194 @@
...
+/// @notice Enables non-members of the mutual to purchase cover policies.
+/// Supports ETH, NXM, and ERC20 asset payments which are supported by the pool.
+/// @dev See supported ERC20 payment methods via pool.getAssets.
+contract CoverBroker is ICoverBroker, Ownable {
+  using SafeERC20 for IERC20;
+
+  // Immutables
+  ICover public immutable cover;
+  IMemberRoles public immutable memberRoles;
+  IPool public immutable pool;
```
shark0der, line 26: The pool contract should not be immutable as it can change letting this contract call into the old pool.

shark0der: We'll have to add an immutable master address instead and pull the pool contract address from it.

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,194 @@
...
+  /// @dev MemberRoles contract needs to be approved to transfer NXM tokens to new membership address.
+  /// @param newAddress The address to which the membership will be switched.
+  function switchMembership(address newAddress) external onlyOwner {
+    nxmToken.approve(address(memberRoles), type(uint256).max);
+    memberRoles.switchMembership(newAddress);
+  }
+
+  /// @notice Transfers all available funds of a specified asset (ETH or ERC20) to the contract owner.
+  /// @param assetAddress The address of the asset to be transferred.
+  function transferFunds(address assetAddress) external onlyOwner {
```
shark0der, line 91: I think we could rename to maybe `rescueAsset` to make the purpose of the function more clear. Up to you to decide if it's better or not.

contracts/external/cover/CoverBroker.sol
shark0der: Yes, the differences I see:
1. you have to fetch the asset address in case of non-nxm
2. the approval target is the cover contract for non-nxm, and token controller for nxm

Perhaps we could fetch/figure the token address in the calling function and:
- either pass the target approval address and token address via params
- or set the approval directly and only pass the token address

shark0der: PS: Leaving them separate is also fine, I don't necessarily see an issue, especially because the nxm approval is done in the constructor.

shark0der: PPS: Rox was considering removing NXM support altogether, no strong opinions.

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  /// @notice Buys cover on behalf of the caller. Supports ETH, NXM and ERC20 asset payments which are supported by the pool.
+  /// @dev For NXM and ERC20 payments, ensure the Cover contract is approved to spend the tokens first (maxApproveCoverContract).
+  /// See supported ERC20 payment methods via pool.getAssets.
+  /// @param params The parameters required to buy cover.
+  /// @param poolAllocationRequests The allocation requests for the pool's liquidity.
+  /// @return coverId The ID of the purchased cover.
+  function buyCover(
+    BuyCoverParams calldata params,
```
roxdanila, line 52: we should validate that params.owner is no the null address or the address of this contract (do the nft doesn't get stuck here)

contracts/external/cover/CoverBroker.sol
rackstar: i like `rescueAsset`

contracts/external/cover/CoverBroker.sol
```diff
@@ -0,0 +1,194 @@
...
+    IERC20 asset = IERC20(assetAddress);
+    uint erc20Balance = asset.balanceOf(address(this));
+    if (erc20Balance == 0) {
+      revert ZeroBalance(assetAddress);
+    }
+
+    asset.transfer(msg.sender, erc20Balance);
+  }
+
+  /// @notice Handles ETH payments for buying cover.
```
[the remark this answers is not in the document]
rackstar: we can replace `_handle` with `_buyCover`.  so `_buyCoverEthPayment` etc..

contracts/external/cover/CoverBroker.sol
rackstar: i'll drop the NXM support as it'll make the contract simpler. I'll just add comments indicating that NXM payment for members should be done directly through Cover.buyCover

contracts/external/cover/CoverBroker.sol
rackstar: good catch. `params.owner` is not validated in the Cover contract

contracts/external/cover/CoverBroker.sol
rackstar: good catch

PR #1130, opened 2024-06-13
Fix/allowed pools migration

MilGard91:
## Context

Since allowed pool is a mapping of productId and array of poolIds, it needs to be fetched one by one with allowedPools(productId, index). Currently all the products if they have allowed pool they are with fixedPrice and have only one allowed pool, that's why we fetch only the first one.

Also there are 3 products that have fix price, but they are depricated and this one more reason why we don't migrate allowed pools for products that are depricated.

Modify the `migrateProductsAndProductTypes` to also migrate allowed pools for products that have fixed price and are not depricated

## Test plan

Tests will be included in the fork tests

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/modules/cover/CoverProducts.sol
+++ b/contracts/modules/cover/CoverProducts.sol
@@ -5,6 +5,7 @@ pragma solidity ^0.8.18;
 import "@openzeppelin/contracts-v4/security/ReentrancyGuard.sol";
 import "@openzeppelin/contracts-v4/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts-v4/token/ERC20/utils/SafeERC20.sol";
+import "hardhat/console.sol";
 
 import "../../abstract/MasterAwareV2.sol";
 import "../../abstract/Multicall.sol";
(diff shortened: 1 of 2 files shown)


contracts/modules/cover/CoverProducts.sol
```diff
@@ -5,6 +5,7 @@ pragma solidity ^0.8.18;
 import "@openzeppelin/contracts-v4/security/ReentrancyGuard.sol";
 import "@openzeppelin/contracts-v4/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts-v4/token/ERC20/utils/SafeERC20.sol";
+import "hardhat/console.sol";
```
roxdanila, line 8: please remove

contracts/modules/cover/CoverProducts.sol
```diff
@@ -320,7 +321,16 @@ contract CoverProducts is ICoverProducts, MasterAwareV2, Multicall {
...
       productNames[i] = _cover.productNames(i);
-      allowedPools[i] = _cover.allowedPools(i);
+
+      /*
+      * Currently only products with fixed price have allowed pools
+      * And all of them have only one allowed pool that's why we fetch only the first one
+      */
+      if (_productsToMigrate[i].useFixedPrice && !_productsToMigrate[i].isDeprecated) {
+        uint _allowedPool = _cover.allowedPools(i, 0);
+        allowedPools[i] = [_allowedPool];
```
roxdanila, line 331: we have just added a product that has 3 allowed pools, instead of 1 like the others... id 186

roxdanila (changes requested):

contracts/modules/cover/CoverProducts.sol
MilGard91: Yeah, I saw it. I will close this PR and create another one with a different approach.

MilGard91:
Closing this PR since the logic depends on all products to have only one allowed pool, and that's not the case anymore

Issue #1121, opened 2024-06-04
Structs and Enums not namespaced within interfaces


roxdanila:
Not all structs and enums with conflicting names are defined inside an interface and result in errors when importing both interface files. As an example, the struct named Configuration is defined in` IIndividualClaims.sol `and in `IYieldTokenIncidents.sol`. It is best practice to define all enums and structs within interfaces and not alongside interfaces in the same file, especially if there may be a naming conflict.

PR #1145, opened 2024-06-20
Chore: data types namespacing

shark0der:
## Context

#1121

Move enums and structs into their interfaces instead of having them at file level.
The `ClaimMethod` enum and the following structs that were left as is are because they're being referenced in other contracts.

```
ICoverProducts: ProductInitializationParams
IStakingPool: AllocationRequest, BurnStakeParams
IPool: SwapDetails, Asset
ICover: PoolAllocationRequest, BuyCoverParams, Product, ProductType, PoolAllocation, CoverData, CoverSegment
```

Note: There's also a workaround that can be used sometimes to avoid clashes in the future by importing only the desired symbols from the target file using the following syntax: `import { IIndividualClaims } from "./IIndividualClaims.sol";`. This way the `Configuration` struct wouldn't be imported. This won't always work, for example if you import two files while both are importing of everything from `IIndividualClaims.sol`.

## Test plan

Used the following script to generate a file that imports all interfaces:
```bash
cd contracts/interfaces
for i in *; do echo "import \"../../interfaces/$i\";"; done | grep -v IERC20.sol > ../ImportEverything.sol
```

IERC20 is excluded because of the ^0.5.0 pragma which breaks the import as there's more than one IERC20: one for 0.5 pragma, one for 0.8.

The generated contract compiled fine but I don't feel like it's a good idea to add it to git because:
1. It needs to be constantly updated as we add new contracts
2. It's not bulletproof - there can be still structs in other files (besides interfaces)

So instead of slowing down the compilation with newer files we should rather watch out for these issues when reviewing PRs.

## Checklist

- [ ] Rebased the base branch

## Review

* :rocket: = Feedback

### DIFF
--- a/contracts/interfaces/ICover.sol
+++ b/contracts/interfaces/ICover.sol
@@ -7,38 +7,17 @@ import "./IStakingNFT.sol";
 import "./IStakingPool.sol";
 import "./ICompleteStakingPoolFactory.sol";
 
-/* ========== DATA STRUCTURES ========== */
+/* io structs */
 
 enum ClaimMethod {
   IndividualClaims,
   YieldTokenIncidents
 }
 
-// Basically CoverStatus from QuotationData.sol but with the extra Migrated status to avoid
-// polluting Cover.sol state layout with new status variables.
-enum LegacyCoverStatus {
-  Active,
-  ClaimAccepted,
-  ClaimDenied,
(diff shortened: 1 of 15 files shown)


MilGard91 (approved):
LGTM

Issue #1173, opened 2024-07-08
Allow batch withdrawal of claimable NXM


roxdanila:
https://nexusmutual.slack.com/archives/C050DJKTPL3/p1719849714049159

rackstar:
![Image](https://github.com/user-attachments/assets/538d7b41-d58c-4b45-8ca6-245f3495f1c3)

rackstar:
# Withdraw Claimable NXM

* ensure to add validations to avoid any calls reverting
  * validate amount > 0
  * validate amount is indeed withdrawable (i.e. not locked)
  * TBD: use getClaimbleNXM for validation or call individual methods?
* ensure that withdrawClaimableNXM and getClaimableNXM should match correctly
* add option to skip claim of assessment stake (v2)
  * only v2 has optionality, v1 claim assessment stake will always be withdrawn

rackstar:
## Withdraw Governance Rewards
1. validate and skip is 0 rewards
```javascript
  GV.getPendingReward(account) => (uint pendingDAppReward)
```
2. withdraw gov rewards
```javascript
  // fromGovernance = true
  // fromAssessment = false
  // batchSize = 100
  TC.withdrawPendingRewards([account, fromGovernance, fromAssessment, batchSize])
```

rackstar:
## Withdraw Claim Assessment Rewards
1. validation (see FE code screenshot below)
```javascript
  AS.getRewards(account) => (uint totalPendingAmountInNXM, uint withdrawableAmountInNXM, uint withdrawableUntilIndex)
  AS.stakeOf(account) => (uint96 amount, uint104 rewardsWithdrawableFromIndex, uint16 fraudCount)
  AS.getVoteCountOfAssessor(account) => (uint voteCount)
```
2. withdraw claim assessment rewards
```javascript
  // fromGovernance = false
  // fromAssessment = true
  // batchSize = 0 // bug in assessment contract

  TC.withdrawPendingRewards([account, fromGovernance, fromAssessment, batchSize])
```


FE Code block:
![Image](https://github.com/user-attachments/assets/019124c4-2356-402e-acc8-9f318282d4ac)

rackstar:
## Withdraw Claim Assessment Stake (optional)
1. Read param if claim assessment stake should be withdrawn or not 
2. validate if 0 stake / isStakeLocked
```javascript
[...]

PR #1201, opened 2024-08-01
feat: Batch NXM withdrawals

roxdanila:
## Context

Closes https://github.com/NexusMutual/smart-contracts/issues/1173

Add `TokenController.withdrawNXM` to withdraw all NXM from the platform. The caller must do the checks if the NXM is withdrawable and modify params accordingly to skip / include specific NXM withdrawals.

* Assessment
  * add `unstakeForAll`
  * convert requires to custom error
  * remove redundant isMember check in `withdrawRewardsTo`
* TokenController
  * add `withdrawNXM`
  * add `_withdrawClaimAssessmentTokensForUser` internal function
  * remove `withdrawPendingRewards` in favour of `withdrawNXM`

## Test plan

* integration tests
* fork tests

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/modules/assessment/Assessment.sol
```diff
@@ -132,25 +132,49 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+  /// @dev At least stakeLockupPeriodInDays must have passed since the last vote.
+  ///
+  /// @param staker  The address of the staker whose stake will be unstaked
+  /// @param amount  The amount of nxm to unstake
+  /// @param to      The member address where the NXM is transfered to. Useful for switching
+  ///                membership during stake lockup period and thus allowing the user to withdraw
+  ///                their staked amount to the new address when possible.
+  function unstakeFor(address staker, uint96 amount, address to) external override whenNotPaused {
+    _unstake(staker, amount, to);
+  }
```
shark0der, line 150: `unstakeFor(rocky, rocky_staked_amount, anatol);`
thx for all the fish

contracts/modules/token/TokenController.sol
```diff
@@ -310,13 +311,68 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
+    // assessment stake
+    if (withdrawOptions.assessmentStake) {
+      IAssessment _assessment = assessment();
+      (uint96 amount, , ) = _assessment.stakeOf(msg.sender);
+      _assessment.unstakeFor(msg.sender, amount, msg.sender);
+    }
+
+    // assessment and governance rewards
+    if (withdrawOptions.governanceRewards || withdrawOptions.assessmentRewards) {
+      withdrawPendingRewards(
```
shark0der, line 339: Nuke the `withdrawPendingRewards` function and inline its code here. There's no point in leaving it.

contracts/modules/token/TokenController.sol
```diff
@@ -310,13 +311,68 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
+    V1CoverNotes calldata coverNotes,
+    uint batchSize,
+    WithdrawNxmOptions calldata withdrawOptions
+  ) external whenNotPaused {
+
+    // assessment stake
+    if (withdrawOptions.assessmentStake) {
+      IAssessment _assessment = assessment();
+      (uint96 amount, , ) = _assessment.stakeOf(msg.sender);
+      _assessment.unstakeFor(msg.sender, amount, msg.sender);
```
shark0der, line 334: I'd create one more function that doesn't need the amount as an input and withdraws all to avoid extra calls and extraneous validations.

contracts/modules/token/TokenController.sol
```diff
@@ -310,13 +311,68 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
+      withdrawPendingRewards(
+        msg.sender,
+        withdrawOptions.governanceRewards,
+        withdrawOptions.assessmentRewards,
+        batchSize
+      );
+    }
+
+    // staking pools rewards and stake (including manager rewards)
+    if (withdrawOptions.stakingPoolStake || withdrawOptions.stakingPoolRewards) {
```
shark0der, line 348: I'm considering to remove these flags for simplicity. The cases are as follows:
1. No stake/rewards - just pass an empty array.
2. Expired stake:
   a. withdrawing both is achieved by default
   b. withdrawing just the stake doesn't necessarily make sense - no reason to leave the rewards or withdraw them separately
   c. withdrawing just the rewards would (kinda) make sense if the user wants to withdraw the rewards but extend their stake, but extend calls withdraw+deposit so technically it would have the same effect as withdrawing and then depositing (just not atomic), so no point to do this either
3. Non-expired stake: only rewards are withdrawable, to avoid withdrawing them just pass an empty array (or skip the id)

In conclusion all flows seem to be covered and we can drop the flags

shark0der: ah and should pass `true, true` to the staking pool. if the stake is not expired there's a check there and it will be skipped anyway - no reverts there.

shark0der: ah, actually there is a slight edge case - withdrawing expired stake if the manager has voted in governance in which case the staker cannot withdraw the stake (but they can withdraw the rewards). I'd say it makes sense to skip both and withdraw both later, i.e. consider rewards unwithdrawable just like the stake.

contracts/modules/token/TokenController.sol
```diff
@@ -310,13 +311,68 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
+
+
+    // v1 pooled staking stake
+    if (withdrawOptions.v1PooledStakingStake) {
+      pooledStaking().withdrawForUser(msg.sender);
+    }
+
+    // v1 - always withdraw (no revert)
+    _withdrawClaimAssessmentTokensForUser(msg.sender);
+    pooledStaking().withdrawReward(msg.sender);
```
shark0der, line 370: Let's evict v1 stuff from this function so we don't have to change it at a later date (contracts+frontend).

contracts/modules/token/TokenController.sol
rackstar: new function:
* remove governance member vote check (as the NXM is going to the staker)
* but still keep assessment vote check

contracts/modules/assessment/Assessment.sol
```diff
@@ -189,20 +217,22 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
-        uint(IMemberRoles.Role.Member)
-      ),
-      "Destination address is not a member"
+    bool isMember = IMemberRoles(internalContracts[uint(ID.MR)]).checkRole(
+      destination,
+      uint(IMemberRoles.Role.Member)
     );
+    if (!isMember) {
+      revert NotMember(destination);
+    }
```
rackstar, line 226: this checks has the advantage of failing early but also adds gas as it duplicates the check that TokenController does on TokenController.mint:

```solidity
  function _mint(address _member, uint _amount) internal {
    require(
      _member == address(this) || token.whiteListed(_member),
      "TokenController: Address is not a member"
    );
    token.mint(_member, _amount);
  }
```

do we still want to keep it?

contracts/modules/token/TokenController.sol
rackstar: to withdraw stake / normal rewards we need:
* tokenIds
* corresponding trancheIds

```solidity
uint poolId = stakingNFT.stakingPoolOf(tokenId);
stakingPool(poolId).withdraw(tokenId, true, true, trancheIds);
```

to withdraw manager rewards we need:
* managed poolIds (since `stakingNFT.stakingPoolOf(0)` will not work)
* corresponding trancheIds for token 0

```solidity
stakingPool(poolId).withdraw(0, true, true, trancheIds);
```

rackstar: so I ended up with this struct to allow user to withdraw both normal / manager rewards at the same time:
```solidity
  struct StakingPoolDeposit {
    uint[] tokenIds;
    uint[][] tokenTrancheIds;
    uint[] managedPoolIds;
    uint[][] managerRewardsTrancheIds;
  }
```

rackstar: so I ended up with this struct:
```solidity
struct StakingPoolDeposit {
    uint[] tokenIds;
    uint[][] tokenTrancheIds;
    uint[] managedPoolIds;
    unit[][] managarRewardsTrancheIds
}
```
its possible to drop `managarRewardsTrancheIds` and re-use the `tokenTrancheIds` but it gets a bit complicated / un-intuitive for the caller:
```solidity
{
  tokenIds: [1,2,3,0,0,5] // if tokenId != 0 fetch poolId, else grab from managedPoolIds
  managedPoolIds: [0,0,0,1,2,0] // <-- only tokenId 0s have corresponding valid poolIds (index 3, 4)
  trancheIds: [[221],[222],[223],[224],[225]]
}
```

contracts/modules/assessment/Assessment.sol
shark0der: Back when it was added [aff26b8d](https://github.com/NexusMutual/smart-contracts/commit/6ac83abad6bca096537684c73232b966aff26b8d#diff-fe843a8caf1022f4bc4e945156cb3f5042b4b00eed433a0509ffb4feb334fc31R199), the check in `TokenController` [didn't exist yet](https://github.com/NexusMutual/smart-contracts/blob/6ac83abad6bca096537684c73232b966aff26b8d/contracts/modules/token/TokenController.sol#L101). It seems redundant now. I'm not a fan of early reverts anymore because normally reverts are caught at simulation phase (eth_call or eth_estimateGas) hence they don't consume gas.

contracts/modules/token/TokenController.sol
rackstar: as discussed it will be implemented as 2 separate structs:

```solidity
  struct StakingPoolDeposit {
    uint tokenId;
    uint[] trancheIds;
  }

  struct StakingPoolManagerReward {
    uint poolId;
    uint[] trancheIds;
  }
```

contracts/modules/assessment/Assessment.sol
```diff
@@ -132,25 +132,49 @@ contract Assessment is IAssessment, MasterAwareV2 {
...
+  function _unstake(address staker, uint96 amount, address to) internal {
+
+    uint stakeAmount = stakeOf[staker].amount;
+    if (amount > stakeAmount) {
+      revert InvalidAmount(stakeAmount);
+    }
+
     // Restrict unstaking to a different account if still locked for member vote
-    if (block.timestamp < nxm.isLockedForMV(msg.sender)) {
-      require(to == msg.sender, "Assessment: NXM is locked for voting in governance");
```
rackstar, line 138: This also seems redundant as `nxm.transfer` requires the voting lock to be expired:
```solidity
  function transfer(address to, uint256 value) public canTransfer(to) returns (bool) {

    require(isLockedForMV[msg.sender] < now); // if not voted under governance
    ...
  }
```

contracts/modules/assessment/Assessment.sol
```diff
@@ -189,20 +213,14 @@ contract Assessment is IAssessment, MasterAwareV2 {
     address destination,
     uint104 batchSize
   ) internal returns (uint withdrawn, uint withdrawnUntilIndex) {
-    require(
```
roxdanila, line 192: why was this removed?

rackstar: its a redundant check as it duplicates the check that TokenController.mint does.

relevant discussion: https://github.com/NexusMutual/smart-contracts/pull/1201#discussion_r1701728982

contracts/modules/token/TokenController.sol
```diff
@@ -277,46 +283,65 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
+    }
+
+    // assessment rewards
+    if (withdrawAssessment.rewards) {
+      // pass in 0 batchSize to withdraw ALL Assessment rewards
+      assessment().withdrawRewards(msg.sender, 0);
+    }
+
+    // governance rewards
+    uint governanceRewards = governance().claimReward(msg.sender, govRewardsBatchSize);
```
rackstar, line 323: This is how we fixed the batchSize issue. We pass in 0 batchSize for `assessment.withdrawRewards`, which effectively withdraws ALL rewards.

The batchSize from the parameters (`govRewardsBatchSize `) will now only be use in governance rewards.

roxdanila: is it possible that all assessment rewards withdrawal won't fit in a block? I guess that's why the batchSize argument was implemented in the first place

rackstar: There has been only 21 assessment so far so it seemed relatively safe to hardcode 0. but now in hindsight, this is a bad design.

it better to expose the assessment batchSize and let the caller (FE) have the option to withdraw all or in batches.

i will add a new `assessmentRewardsBatchSize` param. the withdrawNxm will now look like:

```solidity
  function withdrawNXM(
    WithdrawAssessment calldata withdrawAssessment,
    StakingPoolDeposit[] calldata stakingPoolDeposits,
    StakingPoolManagerReward[] calldata stakingPoolManagerRewards,
    uint assessmentRewardsBatchSize,
    uint govRewardsBatchSize
  ) external {
```

rackstar:
regarding @shark0der comment: https://github.com/NexusMutual/smart-contracts/pull/1201#discussion_r1700117536
> Non-expired stake: only rewards are withdrawable, to avoid withdrawing them just pass an empty array (or skip the id)

I can see a scenario which the user might want to withdraw available rewards (they might want cash out etc..) while the stake is still locked.

contracts/modules/token/TokenController.sol
```diff
@@ -246,87 +251,117 @@ contract TokenController is ITokenController, LockHandler, MasterAwareV2 {
...
-    if (fromAssessment) {
-      assessment().withdrawRewards(forUser, batchSize.toUint104());
+    // assessment stake
+    if (withdrawAssessment.stake) {
+      assessment().unstakeAllFor(msg.sender);
+    }
+
+    // assessment rewards
+    if (withdrawAssessment.rewards) {
+      // pass in 0 batchSize to withdraw ALL Assessment rewards
```
rackstar, line 327: should have removed this comment. will remove once audit review comes back as the burns PR is now rebased behind this branch

rackstar:
marking as draft so it won't be accidentally merged while on audit

PR #1203, opened 2024-08-06
Fix: skip deallocations when burning in grace period

shark0der:
## Context

When a burn is triggered while the cover is in grace period the deallocation is redundant since it will be done when the bucket expiration is processed.

The newly added code skips deallocations if the cover is expired.

## Test plan

Added a test to ensure the allocations stay the same.

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


rackstar:
marking as draft so it won't be accidentally merged while on audit

PR #1230, opened 2024-09-04
feat: long term limit order for SwapOperator

MilGard91:
## Context

Issue: #1218 

- Add a way to track assets that are transfer from the `Pool` to `swapOperator` for the orders
- When calculating eth value of the pool use priceFeedOracle to calculate the value from swapOperator

## Test plan

## Checklist

- [x] Rebased the base branch

## Review

* :rocket: = Feedback


contracts/modules/capital/Pool.sol
```diff
@@ -37,6 +36,9 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
   uint public constant MCR_RATIO_DECIMALS = 4;
   uint internal constant MAX_SLIPPAGE_DENOMINATOR = 10000;
 
+  mapping  (address => uint256) public assetsInSwapOperator;
```
rackstar, line 39: nit: should be no space between mapping and opening parens

contracts/modules/capital/Pool.sol
```diff
@@ -137,7 +140,14 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
         continue;
```
rackstar, line 140: it might be a good idea to assign `assets[i]` to a variable so we only SLOAD once

contracts/modules/capital/Pool.sol
```diff
@@ -137,7 +140,14 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
       }
 
-      total += getAssetValueInEth(assets[i].assetAddress);
+      address assetAddress = assets[i].assetAddress;
+      uint amountInSwapOperator;
+
+      if (isAssetInSwapOperator(i, _assetsInSwapOperatorBitmap)) {
+        amountInSwapOperator = assetsInSwapOperator[assetAddress];
+      }
```
rackstar, line 148: style: might be more concise and readable to make this ternary as we'll explicitly assign 0.

```solidity
      uint amountInSwapOperator = isAssetInSwapOperator(i, _assetsInSwapOperatorBitmap)
        ? assetsInSwapOperator[assetAddress]
        : 0;
```

contracts/modules/capital/Pool.sol
```diff
@@ -156,6 +166,30 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
   }
 
+  function getAssetId(address assetAddress) public view returns (uint) {
+    uint assetCount = assets.length;
+    for (uint i = 0; i < assetCount; i++) {
+      if (assets[i].assetAddress == assetAddress) {
+        return i;
+      }
+    }
+    revert("Pool: Asset not found");
```
rackstar, line 176: Is it worth taking this chance to convert the reverts to custom errors? There are 8 reverts in this contract

contracts/modules/capital/Pool.sol
```diff
@@ -275,8 +309,16 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
+  function setSwapAssetAmount(address assetAddress, uint value) external onlySwapOperator whenNotPaused {
+    assetsInSwapOperator[assetAddress] = value;
+
+    if (value > 0) {
+      uint assetId = getAssetId(assetAddress);
+      assetsInSwapOperatorBitmap = uint32(1 << assetId);
+    } else {
+      assetsInSwapOperatorBitmap = 0;
+    }
+
```
rackstar, line 321: nit: remove new space at end of function

contracts/modules/capital/Pool.sol
```diff
@@ -275,8 +309,16 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
     swapDetails[assetAddress].lastSwapTime = lastSwapTime;
   }
 
-  function setSwapValue(uint newValue) external onlySwapOperator whenNotPaused {
-    swapValue = newValue.toUint96();
+  function setSwapAssetAmount(address assetAddress, uint value) external onlySwapOperator whenNotPaused {
+    assetsInSwapOperator[assetAddress] = value;
```
rackstar, line 313: nit: add new space as delimeter at the start of the function

contracts/modules/capital/Pool.sol
```diff
@@ -137,7 +140,14 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
         continue;
       }
 
-      total += getAssetValueInEth(assets[i].assetAddress);
+      address assetAddress = assets[i].assetAddress;
+      uint amountInSwapOperator;
```
rackstar, line 144: this amount in swap operator is asset specific. the name `amountInSwapOperator` makes me think its not, is it worth prefixing with `asset`?

contracts/modules/capital/Pool.sol
```diff
@@ -127,8 +129,9 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
   ///
   function getPoolValueInEth() public override view returns (uint) {
 
-    uint total = address(this).balance + swapValue;
+    uint total = address(this).balance;
     uint assetCount = assets.length;
+    uint _assetsInSwapOperatorBitmap = assetsInSwapOperatorBitmap;
```
rackstar, line 134: assetsInSwapOperatorBitmap is uint32 should _assetsInSwapOperatorBitmap be uint32 as well?

contracts/modules/capital/Pool.sol
```diff
@@ -102,13 +104,13 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
 
     uint assetBalance;
 
     if (assetAddress.code.length != 0) {
       try IERC20(assetAddress).balanceOf(address(this)) returns (uint balance) {
-        assetBalance = balance;
+        assetBalance = balance + amountInSwapOperator;
       } catch {
         // If balanceOf reverts consider it 0
       }
```
rackstar, line 116: should `amountInSwapOperator` be also added in the `if (assetBalance == 0)` condition below?

MilGard91: Here we sum it with balance if it's zero and the balance is zero, assetBalance will be zero

contracts/modules/capital/Pool.sol
```diff
@@ -102,13 +104,13 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
 
-  function getAssetValueInEth(address assetAddress) internal view returns (uint) {
+  function getAssetValueInEth(address assetAddress, uint assetAmountInSwapOperator) internal view returns (uint) {
 
     uint assetBalance;
 
     if (assetAddress.code.length != 0) {
       try IERC20(assetAddress).balanceOf(address(this)) returns (uint balance) {
-        assetBalance = balance;
+        assetBalance = balance + assetAmountInSwapOperator;
```
shark0der, line 113: We could move this a few lines above to where we initialize the `assetBalance` uint. This way if the balance call fails we don't ignore the balance in the swap operator (although arguably we could as well ignore it already).

contracts/modules/capital/Pool.sol
rackstar: is it possible that all of the Pool's balance is in the SwapOperator? i.e. Pool's balance = 0, but amountInSwapOperator > 0

contracts/modules/capital/Pool.sol
```diff
@@ -37,6 +36,9 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
   uint public constant MCR_RATIO_DECIMALS = 4;
   uint internal constant MAX_SLIPPAGE_DENOMINATOR = 10000;
 
+  mapping(address => uint256) public assetsInSwapOperator;
+  uint32 internal assetsInSwapOperatorBitmap;
```
shark0der, line 40: these are not constants, move them up. the `swapOperator` uses 160 bits, could fit the 32 bits in the same slot. also `s/uint256/uint` please :innocent:

contracts/modules/capital/Pool.sol
```diff
@@ -127,17 +129,30 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
 
-      if (assets[i].isAbandoned) {
+      if (asset.isAbandoned) {
+        continue;
+      }
+
+      address assetAddress = asset.assetAddress;
+
+      uint assetAmountInSwapOperator = isAssetInSwapOperator(i, _assetsInSwapOperatorBitmap)
+        ? assetsInSwapOperator[assetAddress]
```
shark0der, line 147: we could use the asset id as the index here instead of the address. doesn't make too big of a difference gas-wise but might be a better ux for us when trying to get values from it.

contracts/modules/capital/Pool.sol
```diff
@@ -275,8 +315,16 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
 
-  function setSwapValue(uint newValue) external onlySwapOperator whenNotPaused {
-    swapValue = newValue.toUint96();
+  function setSwapAssetAmount(address assetAddress, uint value) external onlySwapOperator whenNotPaused {
+
+    assetsInSwapOperator[assetAddress] = value;
+
+    if (value > 0) {
+      uint assetId = getAssetId(assetAddress);
+      assetsInSwapOperatorBitmap = uint32(1 << assetId);
```
shark0der, line 324: Currently this isn't an issue because swap operator doesn't allow more than one order at a time, but I would insert a require statement checking that `assetsInSwapOperatorBitmap` is zero before overwriting it.

contracts/modules/capital/Pool.sol
```diff
@@ -156,6 +171,31 @@ contract Pool is IPool, MasterAwareV2, ReentrancyGuard {
...
+    revert("Pool: Asset not found");
+  }
+
+  function isAssetInSwapOperator(uint _assetId, uint _assetsInSwapOperatorBitmap) internal pure returns (bool) {
+
+    if (
+      // product does not use default cover assets
+      _assetsInSwapOperatorBitmap != 0 &&
+      // asset id is not in the product's cover assets bitmap
+      ((1 << _assetId) & _assetsInSwapOperatorBitmap == 0)
```
shark0der, line 191: the comments need a cleanup

contracts/modules/capital/Pool.sol
MilGard91: It would still be gt then 0

PR #1287, opened 2024-11-21
feat: limit orders contract

MilGard91:
## Description

Closes https://github.com/NexusMutual/smart-contracts/issues/1308

Add new contract that will handle verification and execution of cover limit orders

## Testing

In progress

## Checklist


contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+  /// @return coverId The ID of the purchased cover.
+  function _buyCoverEthPayment(
+    BuyCoverParams calldata params,
+    PoolAllocationRequest[] calldata poolAllocationRequests
+  ) internal returns (uint coverId) {
+
+    uint ethBalanceBefore = address(this).balance;
+
+    weth.transferFrom(params.owner, address(this), params.maxPremiumInAsset);
+    weth.withdraw(params.maxPremiumInAsset);
```
rackstar, line 127: why do the payment via WETH? why not just do ETH payment?

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+
+    // Compute hash
+    bytes32 orderHash = keccak256(
+      abi.encode(
+        params.productId,
+        params.coverId,
+        params.amount,
+        params.period,
+        params.paymentAsset,
+        params.coverAsset,
```
rackstar, line 81: in `cover.buyCover` function it only allows users to pay with the coverAsset or NXM only. is it worth adding this info in the function annotations for the users?

```
    // can pay with cover asset or nxm only
    if (params.paymentAsset != params.coverAsset && params.paymentAsset != NXM_ASSET_ID) {
      revert InvalidPaymentAsset();
    }
```

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+    if (erc20BalanceAfter > erc20BalanceBefore) {
+      uint erc20Refund = erc20BalanceAfter - erc20BalanceBefore;
+      erc20.safeTransfer(params.owner, erc20Refund);
+    }
+
+    // Emit event
+    emit OrderExecuted(params.owner, coverId);
+  }
+
+  /// @notice Allows the Cover contract to spend the maximum possible amount of a specified ERC20 token on behalf of the CoverBroker.
```
rackstar, line 177: CoverBroker -> CoverOrder

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+      return;
+    }
+
+    IERC20 asset = IERC20(assetAddress);
+    uint erc20Balance = asset.balanceOf(address(this));
+    if (erc20Balance == 0) {
+      revert ZeroBalance(assetAddress);
+    }
+
+    asset.transfer(msg.sender, erc20Balance);
```
rackstar, line 215: asset.safeTransfer?

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+    );
+
+    // Ensure the order has not already been executed
+    if (executedOrders[orderHash]) {
+      revert OrderAlreadyExecuted(limitOrderId);
+    }
+
+    // Recover the signer address from the signature
+    bytes32 ethSignedOrderHash = orderHash.toEthSignedMessageHash();
+    address signer = ethSignedOrderHash.recover(signature);
```
rackstar, line 94: CowSwap and SushiSwap uses EIP-712 for their signatures, presumably to ensure the signatures are domain-specific and cannot be replayed.

UniSwap uses EIP-2612 (i.e. EIP-712 + token approvals via signatures)

should we consider adopting these EIPs for our signature implementation?

rackstar (commented):
few comments

contracts/modules/cover/CoverOrder.sol
MilGard91: There is no option to allow contract to transfer ETH from your wallet. It is only possible to do it with ERC20 tokens.

MilGard91: Other way would be to transfer ETH beforehand which I wouln't do

rackstar: i'm with you now, we only want to pull in the moneys when we hit the our target price not before.

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+    weth.withdraw(params.maxPremiumInAsset);
+
+    coverId = cover.buyCover{value: params.maxPremiumInAsset}(params, poolAllocationRequests);
+
+    uint ethBalanceAfter = address(this).balance;
+
+    // transfer any ETH refund back to params.owner
+    if (ethBalanceAfter > ethBalanceBefore) {
+      uint ethRefund = ethBalanceAfter - ethBalanceBefore;
+      (bool sent,) = payable(params.owner).call{value: ethRefund}("");
```
shark0der, line 136: we should wrap and send `WETH` back

contracts/modules/cover/CoverOrder.sol
shark0der: We absolutely need a domain for the signatures if we ever want to go multi-chain.

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+      if (!sent) {
+        revert TransferFailed(params.owner, ethRefund, ETH);
+      }
+    }
+
+    // Emit event
+    emit OrderExecuted(params.owner, coverId);
```
shark0der, line 143: Sync with Tomasz regarding what info we'd want to include in the events because I bet he'd want more info here.

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+    bytes calldata signature
+  ) external payable returns (uint coverId) {
+
+    if (params.owner == address(0) || params.owner == address(this)) {
+      revert InvalidOwnerAddress();
+    }
+
+    if (params.paymentAsset == NXM_ASSET_ID) {
+      revert InvalidPaymentAsset();
+    }
```
shark0der, line 71: can you remind me please why we don't want to use nxm for payment here please?

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+    bytes32 ethSignedOrderHash = orderHash.toEthSignedMessageHash();
+    address signer = ethSignedOrderHash.recover(signature);
+
+    // Ensure the signer is the user who owns the order
+    if (signer != params.owner) {
+      revert InvalidSignature();
+    }
+
+    // Mark the order as executed
+    executedOrders[orderHash] = true;
```
shark0der, line 102: It is actually recommended to mark the signature as used instead of something else to avoid signature malleability issues (can't see a specific issue here but it's better to be extra cautious).

contracts/modules/cover/CoverOrder.sol
shark0der: The domain should be calculated inside the constructor and should be stored inside a immutable variable.

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
+// SPDX-License-Identifier: MIT
```
shark0der, line 1: 1. This is not an internal contract, this should be moved to `contracts/external/`, maybe together with cover broker.
2. Why `MIT` and not `GPL-3.0-only`?

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+    address _memberRoles,
+    address _nxmToken,
+    address _master,
+    address _weth,
+    address _owner
+  ) {
+    master = INXMMaster(_master);
+    cover = ICover(_cover);
+    memberRoles = IMemberRoles(_memberRoles);
+    nxmToken = INXMToken(_nxmToken);
```
shark0der, line 48: `Cover`, `MemberRoles` and `nxmToken` can be pulled from `Master` to simplify the params and have less room for error.

contracts/modules/cover/CoverOrder.sol
```diff
@@ -0,0 +1,225 @@
...
+
+  /// @notice Verifies and executes the order to buy cover on behalf of the creator of limit order
+  /// @param params Cover buy parameters
+  /// @param poolAllocationRequests Pool allocations for the cover
+  /// @param limitOrderId The ID of the limit order
+  /// @param signature The signature of the order
+  function executeOrder(
+    BuyCoverParams calldata params,
+    PoolAllocationRequest[] calldata poolAllocationRequests,
+    uint limitOrderId,
```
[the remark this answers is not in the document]
MilGard91: This was supposed to be nonce in the signature so we don't have colliding signatures and double execution of the order, so we can remove it if we add a deadline.

contracts/modules/cover/CoverOrder.sol
MilGard91: I had a note for this, but double checked it's not a requirement so I removed it

contracts/modules/cover/CoverOrder.sol
MilGard91: I'll add it here

contracts/external/cover/CoverOrder.sol
```diff
@@ -53,17 +63,30 @@ contract CoverOrder is ICoverOrder, Ownable, EIP712 {
   }
 
   /// @notice Verifies and executes the order to buy cover on behalf of the creator of limit order
+  /// @notice Function only allows users to pay with coverAsset or NXM, this is being checked the Cover contract
```
rackstar, line 66: missing "in"

this is being checked the Cover contract
Comment

contracts/external/cover/CoverOrder.sol
```diff
@@ -0,0 +1,307 @@
...
+        params.period,
+        params.paymentAsset,
+        params.coverAsset,
+        params.owner,
+        executionDetailsHash
+      )
+    );
+
+    // Generate the digest (domain separator + struct hash)
+    bytes32 digest = _hashTypedDataV4(structHash);
```
shark0der, line 181: really tempted to use this digest as the order id and use it as the key in the `orderStatus` mapping. prevents potential signature malleability issues.

contracts/external/cover/CoverOrder.sol
```diff
@@ -0,0 +1,310 @@
...
+        EXECUTE_ORDER_TYPEHASH,
+        params.productId,
+        params.amount,
+        params.period,
+        params.paymentAsset,
+        params.coverAsset,
+        params.owner,
+        executionDetailsHash
+      )
+    );
```
rackstar, line 179: should we add a params.nonce or salt?

currently, we'll block users from creating multiple orders of exactly the same parameter w/o it.

contracts/external/cover/CoverOrder.sol
```diff
@@ -0,0 +1,310 @@
...
+      revert OrderPriceNotMet();
+    }
+
+    if(!memberRoles.checkRole(msg.sender, uint(IMemberRoles.Role.Member))) {
+      revert NotAMember();
+    }
+
+    if (params.owner == address(0) || params.owner == address(this)) {
+      revert InvalidOwnerAddress();
+    }
```
rackstar, line 96: is it intentional that we are letting any member execute any other member's order? 

or do we want to restrict this to just be the owner?

rackstar: there might be front-run scenario here if we let any member execute any other member's order.

a `cancelOrder` from a member might be front-run by an attacker with `executeOrder`.

contracts/external/cover/CoverOrder.sol
```diff
@@ -0,0 +1,310 @@
...
+      revert OrderCannotBeExecutedYet();
+    }
+
+    if (params.maxPremiumInAsset > executionDetails.maxPremiumInAsset) {
+      revert OrderPriceNotMet();
+    }
+
+    if(!memberRoles.checkRole(msg.sender, uint(IMemberRoles.Role.Member))) {
+      revert NotAMember();
+    }
```
rackstar, line 92: should we also check that `params.owner` must be a member?

otherwise it might be possible for a non-member address to be on param.owner and a member executes the order in behalf of them which might break some of our assumptions

MilGard91: Like in the CoverBroker, anyone can buy cover, but if they want to claim then need to become a member.

contracts/external/cover/CoverOrder.sol
MilGard91: Then the limit order looses it's point, if the signer has to come and execute it when the time comes for it. The idea is someone creates an order anyone can execute it when the terms are met. You can always cancel before that.

contracts/external/cover/CoverOrder.sol
MilGard91: Adding a missing param ipfsData will fix this issue

test/integration/setup.js
```diff
@@ -290,6 +292,7 @@ async function setup() {
     { address: ramm.address, code: 'RA' },
     { address: st.address, code: 'ST' },
     { address: coverProducts.address, code: 'CP' },
+    { address: coverOrder.address, code: 'LO' },
```
shark0der, line 295: This will only lead to confusion. Let's rename `CoverOrder` to `LimitOrder` (or even `LimitOrders` - plural)

contracts/mocks/modules/LimitOrders/LimitOrdersCoverMock.sol
```diff
@@ -2,10 +2,12 @@
 
 pragma solidity ^0.8.18;
 import "../../generic/CoverGeneric.sol";
+import "hardhat/console.sol";
```
shark0der, line 5: :broom: :broom: :broom:

contracts/modules/cover/Cover.sol
```diff
@@ -467,7 +471,7 @@ contract Cover is ICover, MasterAwareV2, IStakingPoolBeacon, ReentrancyGuard, Mu
 
       if (remainder > 0) {
         // solhint-disable-next-line avoid-low-level-calls
-        (bool ok, /* data */) = address(msg.sender).call{value: remainder}("");
+        (bool ok, /* data */) = address(buyer).call{value: remainder}("");
```
shark0der, line 474: this might be undesirable since the buyer pays with WETH, but let's leave it as is and I'll get to it when I rebase on top of this. It's going to create conflicts anyway.

contracts/modules/cover/LimitOrders.sol
```diff
@@ -127,16 +130,14 @@ contract LimitOrders is ILimitOrders, MasterAwareV2, EIP712 {
     emit OrderCancelled(id);
   }
 
-  /// @notice Handles verification of the order signature
+  /// @notice Returns the hash of the structured data of the order
```
shark0der, line 133: `struct` is ambiguous, `build` is a bit misleading too, why not call the function `_getOrderHash`?

shark0der: I would also include the `_hashTypedDataV4` function call here as well so you'd return the actual digest directly and can call the function `getOrderId`. The execution details hashing may be moved here as well. The `_extractIdAndSigner` function can then be removed as the only extra thing it does is `ECDSA.recover` which can be inlined. I would also make the `getOrderId` function public as it may be useful for testing but could be used by other contracts.

contracts/modules/cover/Cover.sol
shark0der: would help to leave a todo though!

Issue #1292, opened 2024-11-22
Create functionality to set a Min Price at the listing level


rackstar:
Child of https://github.com/NexusMutual/smart-contracts/issues/1286

https://forum.nexusmutual.io/t/nmpip-protocol-pricing-parameter-adjustments-features/1558

PR #1298, opened 2024-12-03
refactor: change Product struct, remove YieldTokenIncidents contract

duxxud:
## Description

Prerequisite for #1292 
Change the `Product` struct, putting `minPrice` instead of `yieldTokenAddress`.
Removing the `YieldTokenIncidents` contract.

## Testing

Adjusted unit and integration tests.

## Checklist


contracts/interfaces/ICoverProducts.sol
```diff
@@ -17,7 +17,9 @@ struct ProductInitializationParams {
 
 struct Product {
   uint16 productType;
-  address yieldTokenAddress;
+  uint16 minPrice;
+  // leftover memory gap from the previously used address field yieldTokenAddress
+  bytes18 __gap;
```
shark0der, line 22: can we use a `uint144` instead? it's easier when everything's in bits instead of bytes - no conversions needed to calculate

lib/constants.js
```diff
@@ -125,7 +125,7 @@ const InternalContractsIDs = {
   // CL: 9, // CoverMigrator.sol - removed
   AS: 10, // Assessment.sol
   CI: 11, // IndividualClaims.sol - Claims for Individuals
-  CG: 12, // YieldTokenIncidents.sol - Claims for Groups
+  // CG: 12, // YieldTokenIncidents.sol - Claims for Groups
```
shark0der, line 128: just rename

lib/constants.js
```diff
@@ -152,7 +152,7 @@ const ContractCode = {
   StakingProducts: 'SP',
   TokenController: 'TC',
   // TokenData: 'TD',
-  YieldTokenIncidents: 'CG',
+  // YieldTokenIncidents: 'CG',
```
shark0der, line 155: just rename

test/fork/add-asset-cbbtc.js
```diff
@@ -93,7 +93,6 @@ describe('add cbBTC asset to Pool', function () {
     this.proposalCategory = await ethers.getContractAt(abis.ProposalCategory, addresses.ProposalCategory);
     this.stakingPoolFactory = await ethers.getContractAt(abis.StakingPoolFactory, addresses.StakingPoolFactory);
     this.pooledStaking = await ethers.getContractAt(abis.LegacyPooledStaking, addresses.LegacyPooledStaking);
-    this.yieldTokenIncidents = await ethers.getContractAt(abis.YieldTokenIncidents, addresses.YieldTokenIncidents);
```
shark0der, line 96: there was no need to touch any fork tests except the `basic-functionality-tests.js`. they're one-offs, once upgraded - the test's only purpose is to serve as a template/example, nothing more. leave as is already :)

test/fork/basic-functionality-tests.js
```diff
@@ -391,7 +305,8 @@ describe('basic functionality tests', function () {
         ipfsMetadata: '',
         product: {
           productType: 0,
-          yieldTokenAddress: AddressZero,
+          minPrice: 0,
+          __gap: emptyBytes(18),
```
shark0der, line 309: 1. if you switch to uint144 you'll be able to just pass 0
2. does it allow passing undefined instead? as omitting it entirely?

contracts/interfaces/IMasterAwareV2.sol
```diff
@@ -18,7 +18,7 @@ interface IMasterAwareV2 {
     CL, // CoverMigrator.sol - removed
     AS, // Assessment.sol
     CI, // IndividualClaims.sol - Claims for Individuals
-    CG, // YieldTokenIncidents.sol - Claims for Groups
+    // CG, // YieldTokenIncidents.sol - Claims for Groups
```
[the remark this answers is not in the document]
duxxud: Good catch, fixed. I will add a unit test for this.

lib/constants.js
duxxud: Here it will not be consistent with previous removals which are commented out?

lib/constants.js
duxxud: Same here, should we just comment to keep consistency?

contracts/interfaces/ICoverProducts.sol
duxxud: Agree, it will be easier to make changes in tests in the future (in smart contracts doesn't matter as it should be never used). Fixed.

test/fork/basic-functionality-tests.js
duxxud: 1. yes, that would be easier to maintain in the future.
2. no, must be defined.

lib/constants.js
rackstar: yep, let's keep it consistent

lib/helpers.js
```diff
@@ -11,6 +11,8 @@ const toBytes2 = s => toBytes(s, 2);
 const toBytes4 = s => toBytes(s, 4);
 const toBytes8 = s => toBytes(s, 8);
 
+const emptyBytes = size => toBytes('0', size);
+
```
rackstar, line 15: (nit) should we drop this if its not used anymore?

duxxud: removed

lib/constants.js
rackstar: lets keep it as commented out to be consistent

rackstar (approved):

PR #1303, opened 2025-01-13
chore: bump version to 2.11.0

duxxud:
## Description

Updating the version after #1302 changes


package.json
```diff
@@ -20,7 +20,7 @@
   },
   "homepage": "https://github.com/NexusMutual/smart-contracts",
   "dependencies": {
-    "@nexusmutual/deployments": "^2.10.0",
+    "@nexusmutual/deployments": "^2.11.0",
```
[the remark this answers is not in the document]
shark0der: After we release deployments (probably from the RC branch) we can push another PR to update the dep as well. My 2 cents: I don't really like this circular dep, it was introduced somewhat recently, ideally we should fetch all addresses from the chain.

deployments/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nexusmutual/deployments",
-  "version": "2.10.1",
+  "version": "2.11.0",
```
shark0der, line 3: deployments version should be patched at the same time with the address changes `/deployments/src/addresses.json`. there's also abis to update.

duxxud: In this case there are no address changes, and abis are git-ignored, so we just bump the version and publish package?

rackstar:
This PR should also contain the address changes for `deployments/src/addresses.json`

deployments/package.json
duxxud: Actually we added deployment of AssessmentViewer and NexusViewer so their addresses will change. Will add it.

rackstar (approved):

PR #1368, opened 2025-04-25
fix: renewal of a cover that was bought with limitOrders

MilGard91:
Two issues we currently have:
- When there is an order with initial cover buy and renewals, we don't pass an original cover ID when renewing, so the cover doesn't get edited, and we would have double covers for some period. The second renewal would fail because we have a check for the expiration by the initial cover ID that doesn't have a new cover linked to it, and limit order contract would see it as expired, and the renewal wouldn't happen.

- When renewing a cover (edit), since we don't pass the original cover ID, we never get to check the approval and ownership of the cover. And it fails since the msg.sender is the LimitOrders contract and not the owner of the cover. 

## Description

- Created `_params` that copy arguments from `params` passed to the execute function and add coverId if the cover was initially bought by the LimitOrders contract

- Move the check for approval and ownership to the `buyCover` and `executeCoverBuy` methods so we can compare different values. In the case of `buyCover`, we still check the information of msg.sender, but in executeCoverBuy we use the buyer param for the check.

## Testing

## Checklist

### DIFF
--- a/contracts/modules/cover/LimitOrders.sol
+++ b/contracts/modules/cover/LimitOrders.sol
@@ -82,7 +82,6 @@ contract LimitOrders is ILimitOrders, MasterAwareV2, EIP712 {
     bytes calldata signature,
     SettlementDetails memory settlementDetails
   ) external onlyInternalSolver returns (uint coverId) {
-
     require(params.owner != address(0) && params.owner != address(this), InvalidOwnerAddress());
     require(executionDetails.maxPremiumInAsset >= settlementDetails.fee + params.maxPremiumInAsset, OrderPriceNotMet());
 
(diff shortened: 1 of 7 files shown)


contracts/modules/cover/LimitOrders.sol
```diff
@@ -101,16 +100,19 @@ contract LimitOrders is ILimitOrders, MasterAwareV2, EIP712 {
     // Ensure the order is not cancelled
     require(!_orderStatus.isCancelled, OrderAlreadyCancelled());
 
+    uint originalCoverId = _orderStatus.coverId;
     bool isNewCover = _orderStatus.coverId == 0;
```
shark0der, line 104:
```suggestion
    bool isNewCover = _orderStatus.coverId == 0 && params.coverId == 0;
```

PR #1390, opened 2025-05-19
WIP: Permission Claims - Voting Mechanics

rackstar:
## Description

## Testing

## Checklist


contracts/modules/assessment/PermissionedAssessment.sol
```diff
@@ -0,0 +1,329 @@
...
+
+    // Update ballot
+    assessment.ballot[assessorMemberId] = Ballot({
+      vote: vote,
+      ipfsHash: ipfsHash,
+      timestamp: uint32(block.timestamp)
+    });
+
+    // Get the tally again after the vote
+    (acceptCount, denyCount) = _getVoteTally(claimId, assessorGroup);
```
rackstar, line 168: we can avoid calling `_getVoteTally` a second time by manually adjusting the vote counts:
* if its an update
  1. remove the previous vote from the tally
  2. increment the tally with the current vote
* if its a new vote
  1. increment the tally with the current vote
  
but it looked less readable, so opted for better readability for a small cost in gas

contracts/modules/assessment/PermissionedAssessment.sol
```diff
@@ -0,0 +1,396 @@
...
+  }
+
+  /* === MUTATIVE FUNCTIONS ==== */
+
+  /// @notice Initiates a new assessment for a claim
+  /// @param claimId Unique identifier for the claim
+  /// @param productTypeId Type of product the claim is for
+  /// @dev Only callable by internal contracts
+  /// @dev Reverts if an assessment already exists for the given claimId
+  function startAssessment(bytes32 claimId, uint16 productTypeId) external onlyInternal {
```
duxxud, line 149: This interface changes how it works from Claims, as we had before assessmentId which was returned and saved in Claims. Now it's more coupled, maybe change is for better, but requires a lot of changes in Claims also then.

contracts/modules/assessment/PermissionedAssessment.sol
duxxud: But `_getVotedTally` seems heavy in gas as it goes through the whole enumerable set, then check maps for votes. I thing we should try to not call it at all when casting regular votes (left comment below about that, one option is to split `castVote` in few functions as we need total vote count only in special cases).

contracts/modules/assessment/PermissionedAssessment.sol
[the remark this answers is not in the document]
rackstar: good catch. it should be uint

contracts/modules/assessment/PermissionedAssessment.sol
```diff
@@ -0,0 +1,396 @@
...
+    Assessment storage assessment,
+    EnumerableSet.UintSet storage assessorGroup,
+    uint256 assessorGroupLength
+  ) internal view returns (uint256 acceptCount, uint256 denyCount) {
+
+    acceptCount = 0;
+    denyCount = 0;
+
+    for (uint i = 0; i < assessorGroupLength;) {
+        uint256 assessorMemberId = assessorGroup.at(i);
```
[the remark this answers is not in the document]
rackstar: yes, I did consider the memory approach as well but since we are accessing each item only once it should be the same gas cost. so went with storage method as it seemed more readable.

and yes, might be worth doing our own custom EnumberableSet that supports uint32. we should go with the memory approach if that is the case

contracts/interfaces/IPermissionedAssessment.sol
```diff
@@ -0,0 +1,138 @@
...
+  struct AssessmentGroupView {
+    uint id;
+    string ipfsMetadata;
+    address[] assessors;
+  }
+
+  struct Ballot {
+    bytes32 ipfsHash;
+    Vote vote;
+    uint32 timestamp;
```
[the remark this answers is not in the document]
rackstar: yes, we could probably get the timestamp from the event but since it also packs into slot 1, I've added it in for convenience.

contracts/interfaces/IPermissionedAssessment.sol
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  struct Assessment {
+    uint32 start;
+    uint32 end;
+    uint32 assessorGroupId;
+    uint32 finalizedAt; // 0, not closed yet else timestamp of closure
+    AssessmentResult result; // decision snapshot
```
duxxud, line 33: if we have `finalizedAt`, this can be just bool `accepted`?

contracts/modules/assessment/PermissionedAssessment.sol
```diff
@@ -0,0 +1,350 @@
...
+    Ballot storage ballot = assessment.ballot[assessorMemberId];
+
+    ballot.vote = vote;
+    ballot.ipfsHash = ipfsHash;
+    ballot.timestamp = uint32(block.timestamp);
+
+    emit VoteCast(claimId, msg.sender, assessorMemberId, vote, ipfsHash);
+  }
+
+  function closeAssessment(uint256 claimId) external {
```
duxxud, line 164: should it revert in all cases when not closeable?

contracts/interfaces/IPermissionedAssessment.sol
```diff
@@ -0,0 +1,146 @@
...
+
+  struct Ballot {
+    bytes32 ipfsHash;
+    Vote vote;
+    uint32 timestamp;
+  }
+
+  struct Assessment {
+    uint32 start;
+    uint32 end;
```
duxxud, line 30: we probably also don't need `end` anymore, because it shouldn't be changeable and can be derived as `start + VOTING_PERIOD`

PR #1394, opened 2025-06-11
feat: permissioned assessments

### LINKED ISSUE #1386: Claim Assessment - assessment groups management

### LINKED ISSUE #1387: Claim Assessment - voting mechanics

duxxud:
## Description

Closes #1386, Closes #1387 

## Testing

## Checklist

### DIFF
--- a/test-v6/unit/Assessment/castVote.js
+++ b/test-v6/unit/Assessment/castVote.js
@@ -15,7 +15,7 @@ describe('castVote', function () {
     await expect(castVote).to.be.revertedWithCustomError(assessment, 'InvalidClaimId');
   });
 
-  it('should revert when called by non-assessor', async function () {
+  it('should revert when called by non-member', async function () {
     const { contracts, accounts, constants } = await loadFixture(setup);
     const { assessment } = contracts;
     const { CLAIM_ID } = constants;
@@ -23,6 +23,17 @@ describe('castVote', function () {
     const [nonMember] = accounts.nonMembers;
 
     const castVote = assessment.connect(nonMember).castVote(CLAIM_ID, true, constants.IPFS_HASH);
+    await expect(castVote).to.be.revertedWithCustomError(assessment, 'OnlyMember');
+  });
+
+  it('should revert when called by member but not an assessor', async function () {
+    const { contracts, accounts, constants } = await loadFixture(setup);
+    const { assessment } = contracts;
+    const { CLAIM_ID } = constants;
+
+    const [member] = accounts.members;
+
+    const castVote = assessment.connect(member).castVote(CLAIM_ID, true, constants.IPFS_HASH);
     await expect(castVote).to.be.revertedWithCustomError(assessment, 'InvalidAssessor');
   });
 
--- a/test-v6/unit/Assessment/undoVotes.js
+++ b/test-v6/unit/Assessment/undoVotes.js
@@ -3,7 +3,6 @@ const { expect } = require('chai');
 const { setTime } = require('./helpers');
 const { loadFixture } = require('@nomicfoundation/hardhat-network-helpers');
 const { setup } = require('./setup');
-const { setEtherBalance } = require('../../utils/evm');
 
 describe('undoVotes', function () {
   it('should revert when called by non-governor contract', async function () {


contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,523 @@
...
-import "../../interfaces/IMemberRoles.sol";
-import "../../interfaces/INXMToken.sol";
-import "../../interfaces/ITokenController.sol";
-import "../../interfaces/IRamm.sol";
-import "../../libraries/Math.sol";
-import "../../libraries/SafeUintCast.sol";
+import {Multicall} from "../../abstract/Multicall.sol";
+import {IAssessment} from "../../interfaces/IAssessment.sol";
+import {RegistryAware, C_GOVERNOR, C_COVER_PRODUCTS, C_CLAIMS, C_ASSESSMENT} from "../../abstract/RegistryAware.sol";
+import {SafeUintCast} from "../../libraries/SafeUintCast.sol";
```
shark0der, line 10: Please import the files directly as we do through the repo in other contracts. If there are name conflicts try to rename or move the conflicting items inside the interface (as opposed to keeping them at top level).

contracts/interfaces/IAssessment.sol
```diff
@@ -3,180 +3,122 @@
...
-    // An amount of ETH which is sent back to the claimant when the poll result is positive,
-    // otherwise it is kep it the pool to back the assessment rewards. This allows claimants to
-    // open an unlimited amount of claims and prevents unbacked NXM to be minted through the
-    // assessment process.
-    uint128 assessmentDepositInETH;
+  enum AssessmentResult {
+    NONE,
+    ACCEPTED,
+    DENIED
   }
```
shark0der, line 36: `AssessmentResult` doesn't seem to be used anywhere?

contracts/interfaces/IAssessment.sol
```diff
@@ -3,180 +3,122 @@
...
+  struct Assessment {
+    uint32 assessorGroupId; // TODO: set AssessmentData instead
     uint32 start;
-
-    // Timestamp of when the poll ends.
-    uint32 end;
+    uint32 finalizedAt; // 0, not closed yet else timestamp of closure
+    uint8 acceptVotes; // 0, if not finalized yet, should only be set onced finalized
+    uint8 denyVotes; // 0, if not finalized yet, should only be set onced finalized
+    mapping(uint assessor => Ballot) ballot; // only stores latest choice
```
shark0der, line 29: Do not nest mappings inside structs, there will be issues with reading and writing to it, for example cannot write a `Assessment memory` to storage due to the mapping, likewise reading from storage to memory could be misleading. Define a `mapping(uint => mapping(uint => Ballot))` instead.

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,523 @@
...
+  function setGroupMetadata(uint groupId, bytes32 ipfsMetadata) external onlyContracts(C_GOVERNOR) {
+    require(groupId > 0 && groupId <= _groupCount, InvalidGroupId());
 
-    return (totalPendingAmountInNXM, withdrawableAmountInNXM, withdrawableUntilIndex);
+    _groupsMetadata[groupId] = ipfsMetadata;
+    emit SetGroupMetadata(groupId, ipfsMetadata);
   }
 
-  /* === MUTATIVE FUNCTIONS ==== */
+  // todo: remove by address or by memberId?
```
shark0der, line 121: If we remove by address, they can switch their membership to a different address and cosntantly avoid being removed (as we have the timelock in Governor). Removing by ID would prevent this. Would make sense in this case to also update the add function to use the member ids.

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,523 @@
...
+      _groupsForAssessor[assessorMemberId].remove(groupId);
+      emit RemoveAssessorFromGroup(groupId, assessorMemberId);
+    }
+  }
 
-  /// Increases the sender's stake by the specified amount and transfers NXM to this contract
-  ///
-  /// @param amount  The amount of nxm to stake
-  function stake(uint96 amount) public whenNotPaused {
+  function removeAssessorsFromAllGroups(address[] calldata assessors) external onlyContracts(C_GOVERNOR) {
```
shark0der, line 135: likewise, by member id

contracts/interfaces/IAssessment.sol
```diff
@@ -3,180 +3,122 @@
...
-    uint256 lastFraudulentVoteIndex,
-    uint96 burnAmount,
-    uint16 fraudCount,
-    uint256 voteBatchSize
-  ) external;
+  event AssessmentStarted(
+    uint indexed claimId,
+    uint32 assessorGroupId,
+    uint32 start,
+    uint32 end
```
shark0der, line 95: no point in using uint32 for events, just use full uints, the cast will be done automatically

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,523 @@
...
+    require(groupId > 0 && groupId <= _groupCount, InvalidGroupId());
 
-    return (totalPendingAmountInNXM, withdrawableAmountInNXM, withdrawableUntilIndex);
+    _groupsMetadata[groupId] = ipfsMetadata;
+    emit SetGroupMetadata(groupId, ipfsMetadata);
   }
 
-  /* === MUTATIVE FUNCTIONS ==== */
+  // todo: remove by address or by memberId?
+  function removeAssessorsFromGroup(address[] calldata assessors, uint groupId) external onlyContracts(C_GOVERNOR) {
```
shark0der, line 122: It's tempting to pass a single assessor instead of an array. The new Governor can have multiple actions at once, we can call this function multiple times.

contracts/modules/assessment/Claims.sol
```diff
@@ -0,0 +1,301 @@
...
+    return claim;
+  }
+
+  /// Redeems payouts for accepted claims
+  ///
+  /// @dev Anyone can call this function, the payout always being transfered to the NFT owner.
+  /// When the tokens are transfered the assessment deposit is also sent back.
+  ///
+  /// @param claimId  Claim identifier
+  function redeemClaimPayout(uint104 claimId) external override whenNotPaused(C_CLAIMS) {
```
shark0der, line 274: uint104 -> uint ?

contracts/modules/assessment/Claims.sol
```diff
@@ -0,0 +1,301 @@
...
+  /* =========== CONSTANTS =========== */
+
+  uint constant public PAYOUT_REDEMPTION_PERIOD = 30 days;
+
+  uint constant public CLAIM_DEPOSIT_IN_ETH = 0.1 ether; // TODO: set deposit amount
+
+  /* ========== CONSTRUCTOR ========== */
+
+  constructor(address _registry) RegistryAware(_registry) {
+    _nextClaimId = 1; // TODO: start from the last claim Id
```
shark0der, line 41: you can't initialize in a constructor - this is going to be a proxy.

contracts/modules/assessment/Assessment.sol
```diff
@@ -262,61 +242,82 @@ contract Assessment is IAssessment, RegistryAware, Multicall {
...
+    require(totalVotesFromGroup == groupSize, NotEverybodyVoted());
+
+    assessment.votingEnd = block.timestamp.toUint32();
+    _assessments[claimId] = assessment;
+
+    emit AssessmentVotingEndChanged(claimId, assessment.votingEnd);
+  }
+
+  function resetVotingPeriod(uint claimId) override external onlyContracts(C_GOVERNOR) {
+    Assessment memory assessment = _assessments[claimId];
```
rackstar, line 313: since we are only reading the start, cooldownPeriod and votingEnd fields once. is it better to use storage instead?

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
   }
 
-  /// Withdraws all of the the given staker's stake
-  ///
-  /// @dev At least stakeLockupPeriodInDays must have passed since the last vote.
-  ///
-  /// @param staker  The address of the staker whose stake will be unstaked
-  function unstakeAllFor(address staker) external override whenNotPaused onlyTokenController {
-    _unstake(staker, stakeOf[staker].amount, staker);
+  function removeAssessorFromGroup(uint assessorMemberId, uint groupId) override external onlyContracts(C_GOVERNOR) {
```
shark0der, line 120: withdrawing membership should either automatically remove from groups, or we should prevent withdrawing membership while being in an assessment group.

shark0der: Tempting to proxy this through `TokenController` since `Registry` will already depend on TC and I don't really want to make Registry depend on `Assessment` as well

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
+      uint groupId = assessorsGroups[groupIndex];
+      _groups[groupId].remove(assessorMemberId);
+      emit AssessorRemovedFromGroup(groupId, assessorMemberId);
     }
 
-    stakeOf[staker].amount -= amount;
-    nxm.transfer(to, amount);
-
-    emit StakeWithdrawn(staker, to, amount);
+    _clearSet(_groupsForAssessor[assessorMemberId]._inner);
```
shark0der, line 140: This should really be part of the library itself. Given we're generally considering to copy over the libs instead of having external deps - we could copy over this one and add this function to it. Would also be very good to have tests for this lib since we're dealing with internals here.

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
-        (uint(assessment.poll.accepted) + uint(assessment.poll.denied));
+  function setAssessmentDataForProductTypes(
+    uint[] calldata productTypeIds,
+    uint cooldownPeriod,
+    uint groupId
+  ) override external onlyContracts(C_GOVERNOR) {
+    uint length = productTypeIds.length;
+    for (uint i = 0; i < length; i++) {
+      _assessmentData[productTypeIds[i]] = AssessmentData({
+        assessingGroupId: groupId.toUint16(),
```
shark0der, line 87: 1. the group id is not validated.
2. what happens if we change the group while there are assessments that are in progress?

shark0der: on (2), I guess I've just figured it out - you copy the id when starting an assessment so there should be no issue.

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
+    Assessment memory assessment = _assessments[claimId];
+    require(assessment.start != 0, InvalidClaimId());
 
-    uint96 stakeAmount = stakeOf[msg.sender].amount;
-    require(stakeAmount > 0, StakeRequired());
+    return assessment.assessingGroupId;
+  }
+
+  function getAssessment(uint claimId) override external view returns(Assessment memory assessment) {
+    return _assessments[claimId];
```
shark0der, line 202: should it revert if the assessment doesn't exist?

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
+  /// @notice Returns the assessor group ID for a given claim
+  /// @param claimId The claim identifier
+  /// @return The group ID of the assessors for the claim
+  function assessorGroupOf(uint claimId) override external view returns (uint32) {
+    Assessment memory assessment = _assessments[claimId];
+    require(assessment.start != 0, InvalidClaimId());
 
-    uint96 stakeAmount = stakeOf[msg.sender].amount;
-    require(stakeAmount > 0, StakeRequired());
+    return assessment.assessingGroupId;
```
shark0der, line 198: this doesn't seem to be used, should we remove it since we have `getAssessment`?

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
 
-    assessments[assessmentId].poll = poll;
+  /// @notice Returns the ballot for a given claim and assessor
+  /// @param claimId The claim identifier
+  /// @param assessor The address of the assessor
+  /// @return The Ballot struct for the assessor on the claim
+  function ballotOf(uint claimId, address assessor) override external view returns (Ballot memory) {
+    (uint assessorMemberId, ) = _validateAssessor(claimId, assessor);
+    return _ballots[assessorMemberId][claimId];
+  }
```
shark0der, line 236: 1. I really don't like that we query by assessor address here. If they switch to a different one we can no longer fetch this?
2. It's not a good idea to use `_validateAssessor` here. If the assessor was removed from the group you won't be able to fetch the ballot anymore

I would generally advise to remove the `_validateAssessor` helper function as it does too many things:
- fetches the member id
- verifies it's indeed inside the assessment group
- fetches the assessment itself
- validates the assessment exists
- returns the assessment and the member id
Did I expect it to do this based on its name? Hell no.

Separately I would also inline `_hasCooldownPassed` just because it can be a one-liner so there aren't any "savings"

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
+  ///      Must be called by an active assessor for the claim to prevent potential front-running.
+  ///      Sets the assessment's `votingEnd` to the current block timestamp.
+  /// @param claimId The unique identifier for the claim.
+  function closeVotingEarly(uint claimId) override external {
+    // an assessor on the claim must call this to prevent front-running
+    (, Assessment memory assessment) = _validateAssessor(claimId, msg.sender);
+    require(block.timestamp < assessment.votingEnd, VotingAlreadyClosed());
+
+    uint[] memory assessors = getGroupAssessors(assessment.assessingGroupId);
+    uint groupSize = assessors.length;
```
shark0der, line 307: you should use the set's length instead of fetching all members from storage

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
-      // burn from a different merkle tree.
-      burnAmount = burnAmount > _stake.amount ? _stake.amount : burnAmount;
-      _stake.amount -= burnAmount;
-      _stake.fraudCount++;
-
-      // TODO: consider burning the tokens in the token controller contract
-      _ramm().updateTwap();
-      nxm.burn(burnAmount);
-    }
+    require(totalVotesFromGroup == groupSize, NotEverybodyVoted());
```
shark0der, line 319: why not sum up deny + accept instead?

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
 
-    stakeOf[assessor] = _stake;
+    emit AssessmentVotingEndChanged(claimId, assessment.votingEnd);
   }
 
-  /* ========== DEPENDENCIES ========== */
+  /// @notice Resets the voting period for a claim, starting a new full voting window.
+  /// @dev Can only be called by the Governor contract. Reverts if the assessment's cooldown period has already passed.
+  /// @param claimId The unique identifier for the claim.
+  function resetVotingPeriod(uint claimId) override external onlyContracts(C_GOVERNOR) {
```
shark0der, line 330: naming: consider `extend` instead of `reset`. Resseting sounds like setting it to a default, while extending seems more appropriate here.

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
+    // make new group id
+    if (groupId == 0) {
+      groupId = ++_groupCount;
+    }
 
-    emit StakeDeposited(msg.sender, amount);
+    uint length = assessorMemberIds.length;
+    for (uint i = 0; i < length; i++) {
+      uint assessorMemberId = assessorMemberIds[i];
+      require(assessorMemberId != 0, InvalidMemberId());
```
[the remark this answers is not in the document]
rackstar: the caller can pass in 0 as memberId and we should guard against it right?

contracts/modules/assessment/Assessment.sol
```diff
@@ -2,515 +2,365 @@
...
-
-  // An array of merkle tree roots used to indicate fraudulent assessors. Each root represents a
-  // fraud attempt by one or multiple addresses. Once the root is submitted by adivsory board
-  // members through governance, burnFraud uses this root to burn the fraudulent assessors' stakes
-  // and correct the outcome of the poll.
-  bytes32[] public fraudResolution;
-
-  Assessment[] public override assessments;
+  mapping(uint assessorMemberId => mapping(uint claimId => Ballot)) private _ballots;
+  // todo: do we want just event instead of storing?
```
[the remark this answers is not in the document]
rackstar: ok keeping as storage

contracts/modules/assessment/Assessment.sol
rackstar: actually I'm tempted to make the viewer functions to not revert and just return default values if it does not exist.

this might be friendlier for the FE so that they can do multicall on the view functions safely and just interpret the results

so in `assessorGroupOf` i'll drop the require:
Screenshot 2025-07-10 at 16 28 20

contracts/modules/assessment/Assessment.sol
rackstar: so we want to use the storage approach instead of memory on the loop when checking if a current assessor has voted or not.

with memory, at best it will be pretty much similar to storage if all have voted, but we might save a bit of gas if we use storage and not all have voted.

contracts/modules/assessment/Assessment.sol
rackstar: I replaced `TC.removeFromWhiteList` with `TC.removeFromAssessorGroupsAndWhiteList`

PR #1402, opened 2025-07-28
fix: fix submitClaims + add Claims.depositRetrieved field

rackstar:
## Description

* fix submitClaim
* add ClaimAlreadyPaidOut error on submitClaim
* add Claims.depositRetrieved field
* add isAssessor viewer
* add retrieveDeposit unit tests

## Testing

* unit tests

## Checklist


contracts/modules/assessment/Claims.sol
```diff
@@ -172,19 +172,27 @@ contract Claims is IClaims, RegistryAware {
...
+          status != IAssessment.AssessmentStatus.COOLDOWN,
           ClaimIsBeingAssessed()
         );
 
-        require(
-          status == IAssessment.AssessmentStatus.DENIED ||
-          block.timestamp >= cooldownEnd + PAYOUT_REDEMPTION_PERIOD ||
-          _claims[claimId].payoutRedeemed,
-          PayoutCanStillBeRedeemed()
-        );
```
rackstar, line 187: There were 2 bugs here:
1. the require is assuming that if its not DENIED, its ACCEPTED. it can also be a DRAW and with draw payoutRedeemed is always false.
2. on payoutRedeemed the key should be `previousSubmission` not `claimId`.

contracts/modules/assessment/Claims.sol
```diff
@@ -172,19 +172,27 @@ contract Claims is IClaims, RegistryAware {
...
+            !_claims[previousSubmission].payoutRedeemed,
+            ClaimAlreadyPaidOut()
+          );
+
+          // if payout not yet redeemed and still within redemption period
+          require(
+            block.timestamp >= cooldownEnd + PAYOUT_REDEMPTION_PERIOD,
+            PayoutCanStillBeRedeemed()
+          );
+        }
```
rackstar, line 195: I refactored it to make it explicit what exactly are we guarding against.

I also added a new `ClaimAlreadyPaidOut` error if user is trying to double-dip

contracts/modules/assessment/Claims.sol
duxxud: Regarding 1: we have `_claims[claimId].payoutRedeemed = true;` in retriveDeposit as well, so in that case if it is a DRAW it is required to retrieve deposit first before being able to submit new claim. I agree though that we can allow to submit new claim in that case, even before getting deposit back.

contracts/modules/assessment/Claims.sol
shark0der, line 195: sounds like if the claim was accepted but the user didn't redeem the payout you allow it to resubmit, is that the intention?

rackstar:
the "non-assessor" require in `TC.removeFromWhitelist` will be added by @MilGard91 when he adds the registry dependency to TC:
```solidity
  function removeFromWhitelist(address _member) public override onlyInternal {
  
    require(token.balanceOf(_member) == 0, MemberBalanceNotZero());
    
    uint memberId = registry.getMemberId(_member);
    require(!assessment().isAssessor(memberId), MustNotBeAssessorToLeave());
    
    token.removeFromWhiteList(_member);
  }
```

contracts/modules/assessment/Claims.sol
```diff
@@ -28,47 +28,39 @@ contract Claims is IClaims, RegistryAware {
...
   uint private _nextClaimId;
 
+  /* =========== IMMUTABLES =========== */
+
+  ICover public immutable cover;
+  ICoverNFT public immutable coverNFT;
+  ICoverProducts public immutable coverProducts;
+  IAssessment public immutable assessment;
+  IPool public immutable pool;
+  IRamm public immutable ramm;
```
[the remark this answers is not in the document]
shark0der: technically there's a hard dependency on them anyway. those contracts are proxies, they won't change, there's no point to fetch them on every call. we're already using this pattern for other contracts.

contracts/modules/assessment/Claims.sol
```diff
@@ -222,65 +230,74 @@ contract Claims is IClaims, RegistryAware {
...
+    require(!claim.depositRetrieved, DepositAlreadyRetrieved());
 
-    address payable coverOwner = payable(_coverNFT().ownerOf(claim.coverId));
+    _claims[claimId].depositRetrieved = true;
 
-    _pool().returnDeposit(coverOwner, CLAIM_DEPOSIT_IN_ETH);
+    address payable coverOwner = payable(coverNFT.ownerOf(claim.coverId));
 
-    emit ClaimDepositRetrived(claimId, coverOwner);
+    pool.returnDeposit(coverOwner, CLAIM_DEPOSIT_IN_ETH);
```
shark0der, line 285: just a note here, more of a "to keep in mind": if the deposit changes at any point, we'll have to make sure there are no open claims during the contract upgrade. the alternative would be to store the deposit amount which I'm not sure if is worth implementing.

contracts/modules/token/TokenController.sol
```diff
@@ -121,13 +121,18 @@ contract TokenController is ITokenController, ITokenControllerErrors, LockHandle
...
-  /// @dev Removes an address from the whitelist in the token.
+  /// @notice Removes a member from all assessor groups and the whitelist.
+  /// @dev Requires the member's token balance to be zero before removal.
   /// @param _member The address to remove.
   function removeFromWhitelist(address _member) public override onlyInternal {
     require(token.balanceOf(_member) == 0, MemberBalanceNotZero());
     token.removeFromWhiteList(_member);
   }
 
+  /// @notice Switches membership from one address to another, transferring all tokens and whitelist status.
```
shark0der, line 132: switching from one address to another is essentially the "whitelist status" transfer. remove the latter.

test/unit/Claims/setup.js
```diff
@@ -35,30 +29,27 @@ async function setup() {
...
 
+  const [governanceAccount] = accounts.governanceContracts;
   await Promise.all([
     registry.addContract(ContractIndexes.C_COVER, await cover.getAddress(), false),
     registry.addContract(ContractIndexes.C_COVER_NFT, await coverNFT.getAddress(), false),
     registry.addContract(ContractIndexes.C_COVER_PRODUCTS, await coverProducts.getAddress(), false),
     registry.addContract(ContractIndexes.C_ASSESSMENT, await assessment.getAddress(), false),
     registry.addContract(ContractIndexes.C_POOL, await pool.getAddress(), false),
     registry.addContract(ContractIndexes.C_RAMM, await ramm.getAddress(), false),
+    registry.addContract(ContractIndexes.C_GOVERNOR, governanceAccount.address, false),
```
shark0der, line 47: for contracts you can use `contractinstance.target` so it's sync

PR #1406, opened 2025-08-04
DRAFT: feat: add TokenController.unstakeAssessmentFor

rackstar:
## Description

Before we can deprecate the current version of Assessment we first need to empty the contract of stake and rewards. For rewards we can call `withdrawRewards` but for stake we need to create this new helper function `unstakeAssessmentFor` to help us return users stake.

## Testing

* unit test

## Checklist

### DIFF
--- a/contracts/interfaces/ITokenController.sol
+++ b/contracts/interfaces/ITokenController.sol
@@ -85,6 +85,8 @@ interface ITokenController {
 
   function withdrawClaimAssessmentTokens(address[] calldata users) external;
 
+  function unstakeAssessmentFor(address member) external;
+
   function transferStakingPoolsOwnership(address from, address to) external;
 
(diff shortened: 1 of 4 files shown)


shark0der (changes requested):
Absolutely no reason to touch TokenController for this. 
1. The function needs to be in the `Assessment` contract and we won't have to clean anything afterwards.
2. We need to be able to pass an array of addresses to batch-withdraw.
3. There's probably no reason to check the origin of the call anymore, letting anyone call it should be fine.
4. No reason to send this directly to `release-candidate`. We have a bunch of operations to execute anyway during the upgrade, this will be one of them.

rackstar:
closing in favour of: https://github.com/NexusMutual/smart-contracts/pull/1412

PR #1436, opened 2025-08-27
feat: get member claims

shark0der:
## Description

Store all claims of a member in an array.

## Testing

Test added.

## Checklist

### DIFF
--- a/contracts/modules/assessment/Claims.sol
+++ b/contracts/modules/assessment/Claims.sol
@@ -26,6 +26,8 @@ contract Claims is IClaims, RegistryAware {
   // cover as long as the last submitted claim reached a final state.
   mapping(uint coverId => uint claimId) public lastClaimSubmissionOnCover;
 
+  mapping(uint memberId => uint[] claimIds) private _memberClaims;
+
   uint private _nextClaimId;
 
   /* =========== IMMUTABLES =========== */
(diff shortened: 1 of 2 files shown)


contracts/modules/assessment/Claims.sol
```diff
@@ -152,10 +158,14 @@ contract Claims is IClaims, RegistryAware {
     uint32 coverId,
     uint96 requestedAmount,
     bytes32 ipfsMetadata
-  ) external payable override onlyMember whenNotPaused(PAUSE_CLAIMS) returns (Claim memory claim) {
+  ) external payable override whenNotPaused(PAUSE_CLAIMS) returns (Claim memory claim) {
+
+    uint memberId = registry.getMemberId(msg.sender);
+    require(memberId > 0, OnlyMember());
```
duxxud, line 164: nit discussion: I see we will be reusing this more and more instead of modifier `onlyMember`. If we want to be fancy this can be a good place for transient stroage in the modifier, I imagine it can be something like

```
  transient uint private msgSenderMemberId;

  modifier onlyMember() {
    msgSenderMemberId = registry.getMemberId(msg.sender);
    require(msgSenderMemberId > 0, OnlyMember());
    _;
    msgSenderMemberId = 0;
  }
```

and then just use the modifier and `msgSenderMemberId` in function when needed.

duxxud (approved):

contracts/modules/assessment/Claims.sol
shark0der: I wouldn't implement it right now but it's definitely something to keep in mind.
Zeroing it out after the call is probably not necessary as it will be cleared when the transaction ends.
The only thing I don't like is that we decouple the place where we set the value from the place we use it - re-entrancy may be a potential attack vector if we're not careful (unlikely but theoretically possible), in both cases - with or without zeroing out.

duxxud: good points

PR #1443, opened 2025-09-09
test: add layout tests

MilGard91:
## Description

## Testing

## Checklist


shark0der (changes requested):
1. `mainnetLayout.json` is supposed to be a symlink to the actual `<version>.json` layout
2. `Claims` should be entirely ignored, it seems to be failing the tests currently
3. I guess we need a dump for v2.12, but I'm willing to use an older version given in the previous release we didn't check the layouts properly.
4. For the future: to consider versioning the exceptions similar to the layout snapshots, in that case we'll have an entire history of slot changes in these files
5. We could move the entire `layout` folder from `test-v5` to `test` in a separated commit so git history would be shown on these tests correctly.
6. Please sort the contract names in the array :)

PR #1504, opened 2025-12-08
feat: add data and deadline to buyCoverWithRi signature

MilGard91:
## Description

## Testing

## Checklist


contracts/modules/cover/Cover.sol
```diff
@@ -199,6 +204,9 @@ contract Cover is ICover, EIP712, RegistryAware, ReentrancyGuard, Multicall {
       riRequest.premium,
       params.period,
       params.coverAsset,
+      keccak256(riRequest.data),
+      riRequest.dataFormat,
+      riRequest.deadline,
```
rackstar, line 211: noticed that we're not including params.owner on the signed message. is this a deliberate design decision to allow anyone to execute the quote?

contracts/modules/cover/Cover.sol
```diff
@@ -189,6 +191,9 @@ contract Cover is ICover, EIP712, RegistryAware, ReentrancyGuard, Multicall {
           "uint256 premium,",
           "uint32 period,",
           "uint8 coverAsset,",
+          "bytes data,",
+          "uint8 dataFormat,",
+          "uint32 deadline,",
           "uint256 nonce)"
         )
       ),
```
rackstar: this seems better as a immutable constant calculated once at compile instead of calculated on each call

rackstar: @MilGard91 if we want to make very minimal changes we can hardcode the hash value instead:
```solidity
// keccak256("RiQuote(uint256 coverId,uint24 productId,...)")
bytes32 internal constant RI_QUOTE_TYPEHASH =  0x...; 
```

contracts/modules/cover/Cover.sol
rackstar, line 216: we should drop this if we want NXM as paymentAsset. `_buyCover` function already handles the payment validation:
```
require(params.paymentAsset == params.coverAsset || params.paymentAsset == NXM_ASSET_ID, InvalidPaymentAsset());
```

contracts/modules/cover/Cover.sol
rackstar: just realized hardcodong '0x...' is the same as below. so we should just go with below as its more readable: 

```solidity
bytes32 internal constant RI_QUOTE_TYPEHASH = keccak256("RiQuote(uint256 coverId,uint24 productId,...)")
```

shark0der (approved):

